PR #27, opened 2015-07-22
Make produceOperation public

MattKiazyk:
Currently, any subclass of `Operation` can't call produceOperation as it's private to `Operation`


danthorpe:
@MattKiazyk - Didn't spot this - thanks!

PR #31, opened 2015-07-24
added convenience init for ReachabilityCondition to allow default url

MattKiazyk:
I found it repetitive when using the `ReachabilityCondition` to always include the url.

**Updates**
- This PR creates a new `convenience init()` which sets a default url - which is set in the class. 
- `url` is made public which allows a user to set the `url` after the class is created

### DIFF
--- a/framework/Operations/Conditions/ReachabilityCondition.swift
+++ b/framework/Operations/Conditions/ReachabilityCondition.swift
@@ -188,6 +188,8 @@ attached to is asked about its readiness.
 */
 public class ReachabilityCondition: OperationCondition {
 
+    static let defaultURL = NSURL(string: "http://apple.com")
+
     public enum Connectivity: Int {
         case AnyConnectionKind = 1, ConnectedViaWWAN, ConnectedViaWiFi
     }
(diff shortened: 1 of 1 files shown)


danthorpe:
Hmm, I appreciate this might be tedious - I've not actually used `ReachabilityCondition` myself much yet.

At the moment, this project is really quite new and developing quickly, and I'd rather not put a static constant default URL into the class.

In the meantime, if you did, it would solve your problem?

```swift
import Operations

class DefaultReachabilityCondition: Reachability {
    convenience init(defaultURL: NSURL = NSURL(string: "http://www.apple.com")!) {
        super.init(url: defaultURL)
    }
}
``` 

Because... I'm thinking it would be better to have a `ReachabilityObserver` for doing things like executing operations when the network becomes available, as opposed to failing when the reachability condition fails.

MattKiazyk:
Ah thanks for the direction. Ended up going with the extension instead

```
extension ReachabilityCondition {
    convenience init(defaultURL: NSURL = NSURL(string: "http://www.apple.com")!) {
        self.init(url:defaultURL)
    }
}
```
You are correct that the static url shouldn't be in the class.

danthorpe:
@MattKiazyk - oh, yes, of course, and extension is even better.

And, thanks for the feedback - really appreciate it, and please let me know of any other rough edges!

Also, I'm assuming you know that you can use:

```swift
pod 'Operations', :git => 'https://github.com/danthorpe/Operations.git', :branch => 'development'
```

if you want to stay on the bleeding edge. Since `0.5.0` there is now a `LoggingObserver` and `GatedOperation`.

I'll get round to Carthage support in due course.

MattKiazyk:
I'm currently just on master since not using any of the new stuff, but will move it, if I'm needing any of the new stuff for sure

Still trying to wrap my head around everything!

danthorpe:
Okay - makes sense. I ended up watching the WWDC video a few times before starting this properly. 😃

When I get a bit of time (hopefully) over the weekend, or next week, I'll document more of the types, and create an example project.

danthorpe:
@MattKiazyk Hi, you might want to try using `ReachableOperation` now, it's in master as of version 0.6.0. Let me know if you have any problems.I tested it extensively, but it's very hard to QA this sort of thing.

PR #7, opened 2015-07-27
Support all dependencies succeeding condition

danthorpe:


danthorpe:
I don't think this condition is actually possible. I've emailed the following to Dave DeLong:


> Hi Dave, sorry to bother you again...

> I’m still going through the Advanced NSOperations stuff, as I’ve said, and I’ve been writing unit tests to aid my understanding. 

> I’m looking at the `NoCancelledCondition` which isn’t used in your example app, so, trying to infer a little bit how you imagine it to work. My issue is that if a dependency is cancelled, the operation itself never evaluates its conditions - as the parent NSOperation never enters the ready state - so, the expected error cannot be produced. In other words, `NSOperation` by itself already implements the concept of a no cancelled condition. Right?

> My implementation, with unit tests is up here: https://github.com/danthorpe/Operations/pull/7

> I’d really appreciate your thoughts - because I’m thinking that the `NoCancelledCondition` while it sounds nice in practice isn’t actually possible with the way `NSOperation`’s dependencies work.

> Cheers,
> Dan

danthorpe:
Had some contact with Dave, and he's confirmed that this is a current bug with the `Operation` implementation in the sample code.

PR #48, opened 2015-08-14
Clean up AddressBookOperations, ensure silent boolean is available.

danthorpe:


danthorpe:
1. Now all AB operations, and the group exists conditon accepts a suppress permissions boolean, which defaults to false. This will SilentCondition all the conditions.

2. Moves all the addition subclasses into a Cocoapod subspec called "+Extras" e.g. to use these AddressBook operation, you will need to use pod 'Operations/+Extras' to get them.

PR #47, opened 2015-08-16
Possible to create a deadlock using mutual exclusive conditions and group operations.

danthorpe:
It is possible to achieve a deadlock through the use of mutually exclusive conditions (such as alert) and group operations.

When a group operation has a child operation which has a mutually exclusive condition. Adding a condition with the same mutual exclusivity condition type to the group will create a deadlock. This is because the child's condition is added first to the exclusivity manager, and so the group condition will never evaluate. This stops the group from executing, and the child relinquishing it's lock on that mutually exclusive condition.

A simple workaround is to either

1. If you're adding the group's condition inside the group itself, then you can add the child operations during it's `execute` method.

2. If you're adding the group's condition outside of the group, you will need to setup another operation (say a `BlockOperation` and then add what would be the group's condition to it. From inside this operation you can then `produceOperation` the group operation.

I'm currently thinking of a proper solution to this problem as it's that obvious, especially considering an operation's conditions might not be known if using ones from this framework (e.g. `AddressBookCondition` acquires a lock on `AlertPresentation`)

### DIFF
--- a/Operations/Operations/GroupOperation.swift
+++ b/Operations/Operations/GroupOperation.swift
@@ -22,19 +22,16 @@ import Foundation
 */
 public class GroupOperation: Operation {
 
-    private let _queue = OperationQueue()
-    private let _finishingOperation = NSBlockOperation(block: {})
-    private var _aggregateErrors = Array<ErrorType>()
+    private let queue = OperationQueue()
+    private let operations: [NSOperation]
+    private let finishingOperation = NSBlockOperation(block: {})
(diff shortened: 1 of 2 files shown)


danthorpe:
So, yeah, the solution to this is actually incredibly simple - the group operation only adds it's children during it's own `execute` phase. This means that the groups conditions are evaluated first, and only if they pass do the child operations get added, which triggers their conditions to evaluate.

PR #54, opened 2015-08-25
Refactor AddressBook Operations.

danthorpe:
The AddressBookOperations need a bit of an overhaul, as they cannot really be composed currently, as each operation generates it's own `ABAddressBookRef`.

Hence, need to improve the initial access operation, and avoid returning the address book in a completion block. Will also require a more complete abstraction of the system address book api. 

Also, create presentation operations for UI:
- [x] ABNewPersonViewController
- [x] ABPersonViewController


danthorpe:
This is a pretty huge set of changes, I've been playing with the Operations in another app, and will be creating an Address Book focused example app in the next few weeks - just in time for `Contacts.framework`, lols.

PR #55, opened 2015-08-27
Fix compilation and warnings on Xcode7b6.

mblsha:
I've modified 'unused map result' warnings to use foreach syntax in order to improve efficiency a bit. Other option was to silence the warning by adding '_ =' prefix, but I thought that was worse.

### DIFF
--- a/Operations/Operations/GroupOperation.swift
+++ b/Operations/Operations/GroupOperation.swift
@@ -75,7 +75,7 @@ extension GroupOperation: OperationQueueDelegate {
     }
 
     public func operationQueue(queue: OperationQueue, operationDidFinish operation: NSOperation, withErrors errors: [ErrorType]) {
-        _aggregateErrors.extend(errors)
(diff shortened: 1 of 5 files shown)


danthorpe:
This is great @mblsha - I'm a little behind the curve on b7! I'm just updating my build machine, so that everything compiles, will merge asap.

jshier:
Just FYI, but all of the `.map` calls could have been replaced with `.forEach` rather than actual for loops. Keeps the lines exactly the same otherwise. Any interest in such a change?

danthorpe:
Thanks for the info @jshier! That's okay, I will look into it this weekend. I'm a little bit behind on Swift 2.0 but I'm planning on bringing this branch up to date. Including finishing off the Safari operation, and a Contacts.framework version of what I've been doing in #54. Stay tuned!

mblsha:
@jshier Thanks! Fixed in #56 :-)

PR #56, opened 2015-08-28
Use forEach method instead of 'for .. in ..'.

mblsha:
I liked @jshier's suggestion, and improved the code straight away.


danthorpe:
Merged! Thanks @mblsha and @jshier!

PR #59, opened 2015-08-30
Cancel Operation when it finishes with an error.

mblsha:
I think the intention for NoCancelledCondition was to cancel the dependencies when an operation either cancels or finishes with an error. But current implementation works only for explicit cancellations.

If my reading of the NoCancelledCondition intent was incorrect, then I need to add NoFailedCondition, and somehow check that dependent operations finished without having any errors, but there's no way to check for that: the easiest way to solve this would be to add `var errors: [ErrorType] { get { _internalErrors }}` property.

### DIFF
--- a/Operations/Operations/GroupOperation.swift
+++ b/Operations/Operations/GroupOperation.swift
@@ -42,7 +42,9 @@ public class GroupOperation: Operation {
     }
 
     public override func cancel() {
-        _queue.cancelAllOperations()
+        if !self.cancelled {
+            _queue.cancelAllOperations()
+        }
         super.cancel()
     }
 
--- a/Operations/Operations/Operation.swift
+++ b/Operations/Operations/Operation.swift
@@ -212,6 +212,10 @@ public class Operation: NSOperation {
             state = .Finishing
             
             let combinedErrors = _internalErrors + errors
+            if !combinedErrors.isEmpty {
+                // Allow dependent tasks to be cancelled using NoCancelledCondition
+                cancel()
+            }
             finished(combinedErrors)
             
             observers.forEach { $0.operationDidFinish(self, errors: combinedErrors) }
--- a/OperationsTests/NoCancelledConditionTests.swift
+++ b/OperationsTests/NoCancelledConditionTests.swift
@@ -88,5 +88,40 @@ class NoCancelledConditionTests: OperationTests {
         XCTAssertEqual(receivedErrors.count, 1)
     }
 
+    func test__errored_operation_cancels_dependency() {
+        let expectation = expectationWithDescription("Test: \(__FUNCTION__)")
+        let operation = TestOperation()
+        let dependency = TestOperation(delay: 0, error: TestOperation.Error.SimulatedError)
+        operation.addDependency(dependency)
+        operation.addCondition(NoCancelledCondition())
 
+        operation.addObserver(BlockObserver { (_ , errors) in
+            XCTAssertEqual(errors.count, 1)
+            expectation.fulfill()
+        })
+
+        runOperations(operation, dependency)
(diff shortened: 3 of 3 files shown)


danthorpe:
@mblsha well, the intention of `NoCancelledCondition` is to cancel dependencies when an operation is cancelled. I had a discussion about this with Dave while I was writing it as his original sample confused the hell out of me.

But I can see why you would want to do this. I just need to have a think about the impact of calling `cancel()` in the `Operation` base class will have for operations/groups which don't have `NoCancelledCondition`.

danthorpe:
@mblsha okay, I think this is okay, but I think I've also discovered a bug in some of the ready state machine logic. I'm gonna merge this, and will be working on the repo today to bring both Swift 1.2 and 2.0 branches up to feature parity...

mblsha:
@danthorpe Currently I'm battling with Operation not starting when it has two custom conditions. `super.ready` is always false, and conditions are never evaluated. Could it be the same bug?

danthorpe:
If `super.ready` is always false, then it will never evaluate conditions. What is the parent class? If you can create a new ticket and share any details I'll try to help.

The possible bug, which I've spotted is that if the operation is already cancelled, then it would be possible for it to report it self as being ready in states where it shouldn't be. But... this is a relatively low likelihood of having an impact.

Regard this branch I've just merged, I'm back porting it into `development` in #60 and discovering that the unit tests fail, and indeed the dependencies are not considered cancelled. Do you know if there are any changes to `NSOperation` in iOS 9?

mblsha:
Parent class is just `Operation`, so it looks pretty weird. Actually, I did a clunky OS X port and use it inside my small Mac app on OS X 10.10.

Here's a commit where a test with two conditions fail, but everything else passes: https://github.com/mblsha/Operations/commit/9b15777aa6c7a24e115800a1e25a1153ca7bc899#diff-2528f89492f51d699de19560f18a5282R50

BTW, are you interested in OS X compatibility? Ideally files will be moved around to iOS/OSX subfolders, so that podspec could be easily created: https://github.com/SwiftBond/Bond/blob/bond-4/Bond.podspec. But doing it this way will make merging with other branches harder. Other option would be to leave files at their current places and just do an exclusion like this (which looks nasty, but should be easier to merge):
```
  s.osx.exclude_files = "Operations/**/*{RemoteNotification,UserConfirmation,AddressBook,LocationCondition,BackgroundObserver,NetworkObserver,AlertOperation,LocationOperation}*"
```

danthorpe:
Hmm, that's *is* weird! Am I reading this correctly? Both conditions are essentially the same yes?

danthorpe:
@mblsha yes, I am absolutely interested in getting `Operations` working on OS X. I think more than likely I'm missing some `#if iOS #endif` braces around the iOS only stuff.

I think an NSFileManager set of operations/conditions is a great idea - is this what you're working on in your os x branch?

danthorpe:
@mblsha could something to do with the access of `NSFileManager.defaultManager` perhaps if this was a static property on the condition, then both condition values are accessing the same manager?

mblsha:
@danthorpe I'm making a small app for making my work life a little easier (a pretty unique usecase, so I"m going to be open-sourcing only the framework bits — I also have a ShellCommand operation that I use to invoke commands).

Yeah, both conditions are the same. But I was able to reproduce the issue by adding three `NoCancelledCondition()` (two is not enough, apparently).

The NSFileManager code is never running, so there's some common underlying problem, maybe a timing issue.

danthorpe:
Okay. Good to know, well, I'm just updating the Swift 2.0 branch, so I'll look into this next.

danthorpe:
Created #62 for this bug.

PR #60, opened 2015-08-30
Update Swift 1.2 with recent changes in Swift 2.0

danthorpe:
Also there is a slight bug in how `Operation` evaluates readiness, which should be sorted. Also, I'm thinking that the internal `State` should be made thread safe...


danthorpe:
@mblsha will this work for your use case too? Don't see why we can't just expose the errors and create a `failed` boolean state...

mblsha:
@danthorpe Looks nice. Should I port it to the `swift_2.0` branch?

danthorpe:
@mblsha that's okay - I'll port it across shortly. It's a real hassle trying to keep two branches up-to-date, so it'll be easier if I do everything together.

PR #63, opened 2015-08-31
Tests take too long to finish

danthorpe:
Most test operations have execution times of a second. Taking that there are many tests, the whole suite as a whole takes too long to run.

I think it would be possible to make most of them instant (0-second delay), and for those that rely on timers make intervals very small (`0.01` second should be fine, I think).


danthorpe:
Development:

    Executed 129 tests, with 0 failures (0 unexpected) in 54.521 (54.608) seconds

This branch: 

    Executed 129 tests, with 0 failures (0 unexpected) in 18.308 (18.380) seconds

PR #74, opened 2015-09-06
Improvements to AddressBook external change observers

danthorpe:


danthorpe:
Still not really happy with this, but have made changes.

PR #62, opened 2015-09-14
Two identical conditions fail to evaluate correctly

danthorpe:


danthorpe:
@mblsha err, are you still having trouble with two conditions which are the same causing the operation to not start? Would it be possible for you to help me make a test case for this?

mblsha:
It still reproduces on up-to-date swift_2.0 branch with the following patch:

```
diff --git a/OperationsTests/NoFailedDependenciesConditionTests.swift b/OperationsTests/NoFailedDependenciesConditionTests.swift
index 47f525c..c00ebb1 100644
--- a/OperationsTests/NoFailedDependenciesConditionTests.swift
+++ b/OperationsTests/NoFailedDependenciesConditionTests.swift
@@ -115,6 +115,8 @@ class NoFailedDependenciesConditionTests: OperationTests {
         let dependency = GroupOperation(operations: [childOperation])
         operation.addDependency(dependency)
         operation.addCondition(NoFailedDependenciesCondition())
+        operation.addCondition(NoFailedDependenciesCondition())
+        operation.addCondition(NoFailedDependenciesCondition())
 
         var receivedErrors = [ErrorType]()
         operation.addObserver(BlockObserver { (_ , errors) in
```

If three conditions are not enough, just add a couple more :-)

mblsha:
Now tried it on master branch with Xcode 6.4, and instead of timing out it fails with a `receivedErrors.count` mismatch (it's greater than 1).

So probably one can debug this by (Xtracing)[https://github.com/johnno1962/Xtrace] the NSOperations subclass on both `master` and `swift_2.0` branches to note the differences in calls. Or maybe there's an other way to debug this.

danthorpe:
Okay, cool. Well.... looking at the Swift 1.2 (development/master) branches, then this isn't a bug, and is correct behavior. Each condition is treated separately, so the test fails because the assert is incorrect - there should be N errors for N failing conditions.

I will investigate what is happening on Swift 2.0 - this branch will need to be dumped though, as it's branched from development.

danthorpe:
This whole bug is just so weird though. I've got it so that even adding

```swift
op.addCondition(BlockCondition { true })
op.addCondition(BlockCondition { true })
```

means that the parent `NSOperation` readiness is never reached... I'm thinking that perhaps the conditions needs to be a `Set` but I *really* don't want to have to enforce `Hashable` on the `OperationCondition` protocol, or that conditions must be `class`.

I'm going to have to think some more about this. At the moment, I guess all I can think is, just don't attach identical conditions to the same operation... not very helpful I know.

mblsha:
You seem to be right: I've added 6 different types of conditions, and Operation starts just fine:
```
        operation.addCondition(NoFailedDependenciesCondition())
        operation.addCondition(BlockCondition { true })
        operation.addCondition(MutuallyExclusive<UIViewController>())
        operation.addCondition(NegatedCondition(BlockCondition { false }))
        let condition = ReachabilityCondition(url: NSURL(string: "http://apple.com")!, reachability: TestableReachability(networkStatus: .Reachable(.ViaWiFi)))
        operation.addCondition(condition)
        operation.addCondition(SilentCondition(BlockCondition{ true }))
```

I've tried running the timeouting test with ARC, but it detected no memory errors. Maybe it still overwrites some portion of NSOperation's memory without triggering ARC's alarm bells. Weird.

danthorpe:
That's an interesting idea about ARC / memory layout. And, it has to be something odd like that related to `NSOperation` or Xcode 7 GM as I added some logging of `super.ready` and it is just stuck on `false` - plus it's not re-produceable in Xcode 6.4, and the diff is actually fairly minor & unrelated.

I think maybe the next step is to try in the Xcode 7.1 beta.

mblsha:
Same timeout on current Xcode 7.1 beta.

danthorpe:
Okay - thanks @mblsha, I've not had a change to install it yet. I'm going to get in touch with Dave DeLong about this, he might be able to shed some light as he can talk to the  engineer who works on `NSOperation`.

mblsha:
Strangely enough, I'm unable to reproduce the issue by modifying `Advanced NSOperations` source code.

mblsha:
Fixed in #86

PR #81, opened 2015-09-15
Add or make visible some convenience functions

jshier:
Creates `addOperations()` on `GroupOperation` so multiple operations can be added simultaneously.

Makes `addDependencies()` public, so it can be used outside the framework.

What kind of tests do I need to write?

### DIFF
--- a/Operations/Operations/GroupOperation.swift
+++ b/Operations/Operations/GroupOperation.swift
@@ -54,6 +54,10 @@ public class GroupOperation: Operation {
     public func addOperation(operation: NSOperation) {
         queue.addOperation(operation)
     }
+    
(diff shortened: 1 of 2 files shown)


danthorpe:
Err, the unit tests are failing because for Swift 2.0 they need to on a branch named... "swift_2/<< feature name >>"

I can take care of this though - all good ease of use additions!

danthorpe:
I think also... I'm going to remove the `waitUntilFinished:` API. Because it's really not a good idea to block the queues.

danthorpe:
@jshier I've implemented your same changes with a test in #82. Will close this PR, but thanks for the input, very happy to receive such contributions. This week I'll be moving Swift 2.0 branch to the master/development, and versioning 1.0 and 2.0, so development will be a lot easier soon.

PR #85, opened 2015-09-18
Fix errors in Permissions Example

danthorpe:
https://github.com/danthorpe/Operations/blob/development/examples/Permissions/Podfile has a line:

    pod 'YapDatabaseExtensions', :git => 'https://github.com/danthorpe/YapDatabaseExtensions.git', :branch => 'swift_2.0'

But there is no swift_2.0 branch in that repository.


danthorpe:
@Abizern okay, the example works. You shouldn't even need to run `pod install`. It's pretty basic at the moment. Just change the team to your own so that you can run it on a device.

Also, the code which is doing the permission checking, I'm working on something to make this a lot easier for any permission type, so, most of that boiler plate will disappear.

PR #86, opened 2015-09-19
Fix mutual exclusivity

### LINKED ISSUE #62: Two identical conditions fail to evaluate correctly

mblsha:
Fixes #62.

### DIFF
--- a/Operations/Queue/OperationQueue.swift
+++ b/Operations/Queue/OperationQueue.swift
@@ -45,7 +45,7 @@ public class OperationQueue: NSOperationQueue {
 
             // Check for exclusive mutability constraints
             let concurrencyCategories: [String] = operation.conditions.flatMap { condition in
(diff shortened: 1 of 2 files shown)


mblsha:
I added Xtrace.h / Xtrace.mm to the project, and then in the test I did `op.xtrace()`, and compared the trace from the good run with the one from the bad timeout run. The culprit was additional unwanted dependency. `Advanced NSOperations`'s code was correct in this regard, so it was probably a typo.

danthorpe:
Yep, regression introduced in c8296267, okay merging now. Will do a 2.0.1 release. And will also cherry pick into a 1.0.1 release. Really not sure if people are still stuck on Swift 1.2 or not.

Great find @mblsha thanks so much!

danthorpe:
Except it's not a bug in 1.0, so just 2.0.1 then.

PR #88, opened 2015-09-20
Add OS X Support

### LINKED ISSUE #61: Setup OS X test targets.
For full OS X support need to be able to run tests...

mblsha:
Fixes #61.


Operations.podspec
```diff
@@ -15,22 +15,24 @@ session Advanced NSOperations: https://developer.apple.com/videos/wwdc/2015/?id=
   s.module_name       = 'Operations'
   s.social_media_url  = 'https://twitter.com/danthorpe'
   s.requires_arc      = true
-  s.platform          = :ios, '8.0'
   s.default_subspec   = 'Base'
+  s.ios.deployment_target = "8.0"
+  s.osx.deployment_target = "10.10"
 
   s.subspec 'Base' do |ss|
     ss.source_files      = 'Operations/**/*.{swift,m,h}'
```
mblsha: @danthorpe BTW, this rule also greedily includes AddressBook stuff. Should I also fix that?

mblsha:
And I like how files in https://github.com/SwiftBond/Bond are structured. If you would allow me to reorganize the project this way, we could get rid of `#if os(iOS)` and ugly .podspec excludes.
bond 2015-09-20 12-18-22

danthorpe:
Woah. I'm on my way to the football so I can't look at this in detail just yet! But looks good so far. 

Yes, that default subspec is too greedy - mistake - thanks.

But.... I was going to re-org the project to support a core set of features suitable for iOS Extensions, and dovetail OS X (watchOS & tvOS?) support at the same time. There is another issue which you might have seen?

danthorpe:
Oh yes - how SwiftBond does it was what I was thinking! If you have time - this would be amazing!

danthorpe:
I've not looked at SwiftBond's actual project structure, but does it also have separate targets for different platforms? I don't want to exclude Carthage or just building directly from source. I was thinking frameworks for "Operations-iOS", "Operations-OSX" etc? What do you think?

mblsha:
SwiftBond has two targets: Bond and BondOSX, and employs a neat trick: both targets have same `Product Name` and `Product Bundle Identifier`. This allows for Test sharing without `#if`'s. Bond also shares the `Public Header` and `Info.plist`.

Bond arranges its files into a neat directory structure, and makes sure that only the relevant files are included for proper targets: this is done by checking `Target Membership` checkboxes in File Inspector (I've unchecked a couple of files already, as having `#if os(iOS)` still resulted in weird compilation errors on OS X).

In the same spirit I left the Operations / OperationsTests as is and added OperationsOSX and OperationsOSXTests, and removed extrensous headers and Info.plists, so there are no new files added in this pull request.

Haven't looked at watchOS / tvOS, but guess adding those targets could be done using the same means.

mblsha:
Will try re-arranging the files now :-)

Operations.podspec
```diff
@@ -15,22 +15,33 @@ session Advanced NSOperations: https://developer.apple.com/videos/wwdc/2015/?id=
...
   s.requires_arc      = true
-  s.platform          = :ios, '8.0'
-  s.default_subspec   = 'Base'
+  s.ios.deployment_target = "8.0"
+  s.osx.deployment_target = "10.10"
 
-  s.subspec 'Base' do |ss|
-    ss.source_files      = 'Operations/**/*.{swift,m,h}'
-    ss.exclude_files     = 'Operations/Extras/**/*.{swift,m,h}'
-  end
```
mblsha: I've checked this podspec on my Mac project, and it still works.

mblsha:
Hmm, it appears that I've broken the `examples/Permissions`. Looking into it.

Operations.podspec
mblsha: Ah, it seems that without a Base subspec, adding `Operations` results in adding all subspecs as well. Restoring the Base subspec.

mblsha:
There. Checked the podspec on both my OS X project and Permissions example.

danthorpe:
@mblsha hmm, unfortunately now the unit tests don't compile in CI. I think what I'll do, is pull your development branch into a feature branch, go through your changes and then merge that.

mblsha:
Okay, as I can't see the actual Buildkite error messages. But both targets work for me here, using the Xcode 7 GM.

danthorpe:
You can run `$ .scripts/test.sh` to run the unit tests. It has an issue with `AddressBook.swift:696:11` - use of undeclared type 'OPRAddressBookChangeHandlerContainer'

danthorpe:
OS X support, and iOS Extension API, watchOS, tvOS support is added in #90. I used the same strategy here, except with three more targets, and updated the unit test scripts.

Thanks @mblsha!

PR #93, opened 2015-09-24
Contacts #1

danthorpe:
Support use of Contacts framework (across all appropriate platforms)

- [x] `ContactsCondition` - a condition to require permissions to Contacts.framework
- [x] `ContactsOperation` - basic low level operation(s) which own the store, and provide "execute" method suitable for subclassing.
- [x] `GetContacts` - operation to get contacts based on predicate
- [x] `GetGroup` - operation to get a group with a name, create it if necessary.

### DIFF
--- a/Operations.podspec
+++ b/Operations.podspec
@@ -34,9 +34,16 @@ session Advanced NSOperations: https://developer.apple.com/videos/wwdc/2015/?id=
   end
 
   s.subspec '+AddressBook' do |ss|
-    ss.platform = :ios, "8.0"
     ss.dependency 'Operations/App'
-    ss.source_files   = 'Operations/Extras/AddressBook/iOS'
+    ss.source_files = [
+      'Operations/Extras/AddressBook/iOS',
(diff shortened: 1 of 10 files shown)


tkrajacic:
OS X 10.11 also has the Contacts framework now, and parity would be nice ;)

danthorpe:
Hi @tkrajacic - it's in progress, should merge into `development` branch tonight/tomorrow. And release 2.2 early next week.

At the moment, I'm not unifying any of the AddressBook / Contacts operations. Partly to avoid premature optimization, and because they have some subtle (and not so subtle) differences. So if you want to support multiple versions, then you will need to use `#available` to create different operations.

danthorpe:
This issue PR is going to become too big/unweildly, so this first issue creates the types and basics. There will need to be additional tickets to add proper test coverage to the operations, which should be possibly, as they're generic over the contact store type.

For application usage though, there are proper type aliases which hide this implementation detail - thanks Swift 2.0!

PR #97, opened 2015-10-01
Generic and unified authorization interface.

danthorpe:
Currently, testing the authorization of device or service capabilities is a fractured and verbose process. What I want to have is a unified and simple interface to figure out the authorization state of whatever capability.

Therefore, it requires a single `status` enum, a generic `AuthorizationStatus` operation, with a generic parameter which will accommodate everything from `ABAddressBook` to `HealthKit`.

- [x] Location
- [x] Calendar (EventKit)
- [x] Health
- [x] Photos
- [x] Passbook
- [x] Cloud
- ~~[ ] AddressBook~~ 
- ~~[ ] Contacts~~ 

`AddressBookCondition` & `ContactsCondition` will be unified and turned into a single capability in the next version. In other words, `Capability.Contacts` will be a capability, which on iOS 9 will utilize Contacts.framework, and on iOS AddressBook.framework, but will offer a unified API for the same operations. - This is quite a bit more work, and I don't want this branch to bloat out of scope.


danthorpe:
Okay, made a start on this, and currently added support for Location stuff. Here's an example of how it works / will work.

The problem at the moment is that figuring out the authorization status for Core Location is a bit of pain, like this:

### The current way...

```swift
let authorization = BlockOperation { 
    self.state = .Authorized
}
authorization.addCondition(SilentCondition(LocationCondition()))
authorization.addObserver(BlockObserver { _, errors in 
    if let error = errors.first as? LocationCondition.Error {
        switch error {
                case .AuthenticationStatusNotSufficient(CLAuthorizationStatus.NotDetermined, _):
                    self.state = .Unknown
                case .AuthenticationStatusNotSufficient(CLAuthorizationStatus.Denied, _):
                    self.state = .Denied
                default:
                    self.state = .Unknown
       }
    }
})

queue.addOperation(authorization)
```

Which is not exactly ideal, because

1. I probably just want to do something for `NotDetermined`, `Denied` etc - and I don't want to have to maintain the state of this authorization status - the framework should just tell me.
2. Each type of capability will have broadly the same sort of authorization statuses, but having separate conditions for each one, with individual errors leads to poor discoverability. And results in lots of repeated boilerplate code which is generally doing the same stuff.

Therefore, instead of defining the API around individual condition types & permission operations, we can condense this functionality into a `CapabilityType`.

### The new way...

```swift
queue.addOperation(GetAuthorizationStatus(Location(.WhenInUse)) { enabled, status in 
    switch (enabled, status) {
        case (false, _):
           // Location services are not enabled
        case (true, .NotDetermined):
           // Location services are enabled, but not currently determined for the app.
    }
})
``` 

Things to note here...
1. `GetAuthorizationStatus` just gets the current status, it will not trigger any permission alerts. So it can be used as much as needed to check the status of the capability - i.e. to setup a view / button / UI.
2. `GetAuthorizationStatus` is generic over the _capability_, which in turn has a nested type for the status. This means, for `Location` that the status received in the block is a `CLAuthorizationStatus` so that native types are maintained. This is in preference to Operations defining its own common status enum.
3. `GetAuthorizationStatus` is initialized (with no label) with its capability. The capability implements `CapabilityType` which in turn must be initialized with a _requirement_ (`AuthorizationStatusType.Requirement`) which allows the API to support mechanisms such as authorization levels (used in CoreLocation, HealthKit etc). Capabilities which don't have this, will be defined with a default `Void` requirement.

### Triggering permissions... 
In order to trigger permissions, we have a new condition, `AuthorizedFor<Capability>` which works like this...

```swift
let operation = MyOperation()
operation.addCondition(AuthorizedFor(Location(.WhenInUse)))
queue.addOperation(operation)
```

This is similar to the current bespoke condition style, with these significant changes.

1. Single condition for all "capability" related permissions.
2. Single `ErrorType` (`CapabilityError`) for all "capability" related errors, including (`.NotAvailable` and `AuthorizationNotGranted`).

### Other benefits

Now that discovering and requesting the authorization status is generic over capability, it means that we could write generic UI code like buttons and views etc which automatically support the correct behavior for requesting permissions - potentially saving a lot of copy/paste code.

### Implementation details

In order to make the various _capability_ types testable, they need to support a nested generic type called `Registrar` which conforms to `CapabilityRegistrarType`, although each capability will extend this. Therefore unit tests can implement testable registrars, classes like `CLLocationManager` conform via very simple extensions (the only bits which cannot be tested). And, framework consumers can be presented with a clean type via a typealias. For example...

```swift
public typealias Location = _LocationCapability<CLLocationManager>
```

The benefit of this, is that the operations and conditions can be tested against `_LocationCapability` which in turn can be initialized with a testable registrar to ensure that it is working correctly. By allowing each `CapabilityType` to define the interface for its own registrar allows it to mirror the real type so that there is minimal code which cannot be tested.

PR #99, opened 2015-10-06
Support CapabilityType in AddressBook & Contacts

danthorpe:
I want to unify AddressBook & Contacts functionality, so that framework consumers do not need to jump through `#available(iOS 9, *)` hoops. This is the functionality that I want. I'm going to keep it under an ~~"AddressBook"~~ "Contacts" name.

- [ ] Get Contacts
- [ ] Map Contacts
- [ ] Get Group
- [ ] Create Group
- [ ] Remove Group
- [ ] Add Contacts to Group
- [ ] Remove Contacts from Group

Merging the actual functionality into a `CapabilityType` will give us the authorization for free.


danthorpe:
Okay, so after trying to get this to work, it turns out that it's not really possible to provide a testable unified API coverage different OS frameworks across different platforms & platform versions.

The problem is this, lets say we want to define a capability: `Capability.Contacts` which, on iOS 9 would utilize the `Contacts.framework` but on iOS 8 would use `ABAddressBook.framework`, and we want to be able to unit test it as much as possible. This would require making the underlying type generic over the "registrar" or contact store. Think `CNContactStore` and `ABAddressBookRef` (or wrapper).

So, we might try this:

```swift
extension Capability {

    @available(iOS 9.0, OSX 10.11, watchOS 2.0, *)
    public typealias Contacts = _ContactsCapability<ContactsStore>

    @available(iOS, deprecated=9.0)
    public typealias Contacts = _ContactsCapability<AddressBookStore_iOS>
}
```

But this is not allowed by the Swift compiler - the compiler thinks there is a name clash here. Using `@available(iOS 9.0, unavailable)` is not valid. `unavailable` can only indicate that the declaration is not available on an entire platform, not restricted to the version of a platform.

Okay, so we've already failed at the first hurdle. If you wanted to add a constraint to an operation, you'd need something like this:

```swift
if #available(iOS 9.0, *) {
    op.addConstraint(Capability.Contacts())
}
else {
    op.addConstraint(Capability.AddressBook_iOS())
}
```

Now you might think it would be possible to define a function to return the correct capability.... but, no, because the generic type will be different.

This problem exists as you go further down the rabbit hole. It boils down to these restrictions...

### It's not possible to use `@available` for stored properties.

This means that it's not possible to define say one type `ContactStore` which internally will allocate a `CNContactStore` on iOS 9, and `ABAddressBookRef` on iOS 8. 

We have to define a common API in a protocol which is available across all platforms & platform versions, using `typealias` for everything. E.g.

```swift
public protocol ContactType {
    typealias Identifier
    typealias PropertyDescriptor
}

public protocol GroupType {
    typealias Identifier
}

public protocol ContactsStoreType: ContactsRegistrarType {
    typealias Contact: ContactType
    typealias Group: GroupType
}
```

This is okay, we can (probably) write wrappers around `CNContactStore` and its friends and `ABAddressBook` and its friends which conforms to this common API.

To actually use these types in operations, say, `CreateContactsGroupOperation` in a normal way, it would need to be defined like this....

```swift
class _CreateContactsGroupOperation<Store: ContactsStoreType>: _ContactsOperation<Store> { 
    // etc
}
```

But, this leads us back to the same issue as with the capability type. We would then need to define different typealias across platforms to set the store type. Which means that the call site would still need a bunch of `#available` mess.

So, at the moment, in this framework there is a set of (poorly tested) `AddressBook` operations for iOS 8. And there is a set of (tested as much as possible) `Contacts` operations for iOS 9. Framework consumers will need to likely need to define their own logic for working with both in the same codebase.

e.g.

```swift
enum SystemContact {
    case Contact(CNContact)
    case ABPerson(AddressBookPerson)
}
```

If anyone has any bright ideas on another approach - let me know!

PR #89, opened 2015-10-21
Support for tvOS

danthorpe:
With the new Apple TV, the support for tvOS would be great. :)


danthorpe:
Have updated the podspec for watchOS and tvOS this now works - although currently only on development. See the comment above - but pod linting passes with the latest Xcode and CocoaPods.

I have removed all contacts related stuff from the watchOS target and will restore it as part of #99.

PR #113, opened 2015-12-04
Suppress AddressBook related warnings when building against iOS 9.0

danthorpe:


danthorpe:
Output from Carthage now looks like...

```bash
*** Fetching Operations
*** Checking out Operations at "b3dc08476b137322839e668d9c43fbf051a7f901"
*** xcodebuild output can be found in /var/folders/6j/9k2drdbs1fv9fw5d6wng5yzh0000gn/T/carthage-xcodebuild.BZ5ckd.log
*** Building scheme "Operations-OSX" in Operations.xcodeproj
*** Building scheme "Operations-tvOS" in Operations.xcodeproj
*** Building scheme "Operations-watchOS" in Operations.xcodeproj
*** Building scheme "Operations-iOS" in Operations.xcodeproj
*** Building scheme "Operations-iOS-Extension" in Operations.xcodeproj
```

Partly the reason for this was also that the deployment target of the extension framework was set to 9.0 which is wrong, as we should still be supporting 8.0 - whoops!

Issue #112, opened 2015-12-04
iOS extensions warnings when building via carthage


JamieWhite:
How do I build the iOS extensions framework so I can silence the extension warning?

danthorpe:
Hi @JamieWhite, I there is a bug here I think, the module names are all just `Operations` and so the files made for the framework is just `Operations.framework` which means that while Carthage does correctly build all the frameworks, when it builds the `iOS` framework it overwrites the `iOS-Extension` framework.

If you have a fork, you could rename the Product Name in the iOS Extension target to `Operations-Extension` or something. This would be the first thing I would try, as I think it will change the files that carthage generates.

Also, I've removed the zip from the 2.4.0 release in Github. So, running `update` will now build.

```bash
$ carthage update
*** Fetching Operations
*** Checking out Operations at "2.4.0"
*** xcodebuild output can be found in /var/folders/6j/9k2drdbs1fv9fw5d6wng5yzh0000gn/T/carthage-xcodebuild.Y9xU8c.log
*** Building scheme "Operations-OSX" in Operations.xcodeproj
*** Building scheme "Operations-watchOS" in Operations.xcodeproj
*** Building scheme "Operations-tvOS" in Operations.xcodeproj
*** Building scheme "Operations-iOS-Extension" in Operations.xcodeproj
/Users/daniel/Desktop/Operations-examples/Carthage/Checkouts/Operations/Operations/Extras/AddressBook/iOS/AddressBook.swift:644:11: warning: 'AddressBook' was deprecated in iOS 9.0
/// Plus lots more AddressBook related warnings...

Ice:Operations-examples daniel$ 
```

JamieWhite:
Seems someone has spoke to Carthage about this: https://github.com/Carthage/Carthage/issues/933

JamieWhite:
Looks like we're still waiting on Carthage for this. I think Im gonna try and get my fork workable in extensions for the time being. I'm still using PSOperations and I'm keen to get off it.

danthorpe:
Yeah, looks like it, although I'm not had a chance to catch up on where Carthage is. If anyone has any suggestions to get round this - PRs are more than welcome!

JamieWhite:
Interesting update out by Carthage: https://github.com/Carthage/Carthage/releases/tag/0.12

[...]

PR #114, opened 2015-12-04
Create different framework files for iOS Extension

danthorpe:
There is basically a problem where building the frameworks means that the iOS Extension framework get's over-written by the iOS (app) framework.

This is to address #112, see also an issue raised in the Carthage project: https://github.com/Carthage/Carthage/issues/933


danthorpe:
I thought I could fix this by just changing the product name. But that doesn't work because then the module name also gets changed.

If I explicitly set the module name, then I get a build error (although it says warning)

> warning: PRODUCT_MODULE_NAME may not be overridden for framework target 'Operations-iOS-Extension'

I've also tried overriding the WRAPPER_EXTENSION to be `extension.framework` but this doesn't work either.

Next ideas: 

1. Rename the files via a script as the last build phase? This is the most pragmatic idea so I'll try it next, the possible problem I see is that the unit tests won't be able to find the bundle? Could also just delete the test target for the extension scheme - all of the code is tested in the main app framework target.

2. Probably the "correct" way to handle this is to use a module map with an umbrella module called `Operations` and then internal modules for the app level stuff. So that

    ```swift
    import Operations // imports the extension api compatible stuff
    import Operations.Features // ? imports the non-extension api compatible stuff, not available in extensions.
    ```` 

Anyone got any bright ideas?

danthorpe:
Okay, at the moment, on this branch there are two Xcode projects, one for normal frameworks, and the other for extension API only frameworks. However, waiting for this PR to merge before Carthage will handle it: https://github.com/Carthage/Carthage/pull/892

I'm gonna hold off merging this until that change makes it through.

danthorpe:
Okay, I'm actually going to merge this - as it's got significant changes which are also in CI (just the script names).I'll just make sure that it doesn't break normal Carthage usage...

Issue #115, opened 2015-12-04
GroupOperations aggregateErrors


JamieWhite:
Is there any reason this is an internal method?

PR #116, opened 2015-12-04
Making GroupOperation aggregateError: a public method

JamieWhite:
This is so we can aggregate errors from child operations

The open issue is here: https://github.com/danthorpe/Operations/issues/115


danthorpe:
Hmm - I need to make some changes to my CI scripts now (they got changed for #114).

JamieWhite:
Admittedly this failed locally but I was convinced it couldn't have failed from my change so pushed anyway to see what CI said.

danthorpe:
You'll also need to change the property to be visible too - just make it `public private(set) var aggregateErrors`.

I don't normally use this much as I also almost inspect errors when I override `operationDidFinish` - again thanks for the spot!

JamieWhite:
Well I've just learned something. I didn't know public private(set) was a thing.

JamieWhite:
So the unit tests pass for me locally. What was failing beforehand was a test on GatedOperation but I cant find the test now.

danthorpe:
Don't worry about these build failures - it's nothing to do with your changes - will have it merged in a jiffy.

danthorpe:
Yeah, the test suite, especially on iOS has some poor/unstable tests which fail occasionally, mostly in LocationOperationTests, but sometimes GatedOperation.

danthorpe:
Okay, I've merged #114 into `development` which is now building - when all is well, you should be able to pull dev, push up here, and the unit tests will run in CI. I suggest you just branch again from development and add your changes - quite a bit of moving around happened today.

JamieWhite:
test__when_gate_is_open_operation_is_performed fails for me locally

danthorpe:
Is it an async timeout?

JamieWhite:
aaaaaand re-running it passes. I think it's a faulty test.

The problem was XCTAssertTrue(gate.finished) randomly fails

JamieWhite:
Big code coverage drop!

danthorpe:
Yeah, I should probably remove/rewrite them.

danthorpe:
Don't worry about that - it's because of path changes, and i've not updated Codecov yet....

danthorpe:
Damnit - I think that's because it's only sending the coverage of the extension tests now, as development has somehow gone down to 26%... okay, one more ticket.

danthorpe:
Okay, this just needs development pulled into it again. Sorry. Should be okay now.

JamieWhite:
Sorted.

Issue #121, opened 2015-12-10
Crash when logging UserLocationOperation


jshier:
I created a simple `OperationQueue` manager to find the current user location:
```swift
import Operations

typealias LocationCompletionHandler = (location: CLLocation) -> Void

class LocationManager : OperationQueue {
    private static let sharedManager = LocationManager()
    
    static var lastUserLocation: CLLocation?
    
    static func currentUserLocation(accuracy: CLLocationAccuracy = kCLLocationAccuracyThreeKilometers, completionHandler: LocationCompletionHandler) {
        let memoizingCompletionHandler: CLLocation -> Void = { location in
            lastUserLocation = location
        }
        
        let locationOperation = UserLocationOperation(accuracy: accuracy, completion: memoizingCompletionHandler)
        sharedManager.addOperation(locationOperation)
    }
}
```
Attempting to run it as so:
```swift
LocationManager.currentUserLocation { location in
    log("*** currentUserLocation: \(location) ***")
}
```
(I have my own `log` function)
Results in a crash in `Logging.swift` line 271. `EXC_BAD_ACCESS (code=EXC_I386_GPFLT)` Earlier in the call stack is `LocationOperation.swift`, line 141. This is on the main thread.

danthorpe:
Hi @jshier - sorry about this. I must have missed something in the logging stuff I've added recently. It's almost 2am now, but will look into this tomorrow/later.

danthorpe:
Hi @jshier is there any chance you could take a look at the Permissions example project on the branch in #122

I have added to the Location example your Manager class (more or less, you don't call the completion handler in the code above).

And when I run this I get...

> User Location >>> MutuallyExclusive<CLLocationManager>
> User Location: Initialized -> Pending
> User Location: Pending -> EvaluatingConditions
> User Location: EvaluatingConditions -> Ready
> User Location: Ready -> Executing
> User Location: did start
> User Location updated last location: <+51.30969064,-0.12566690> +/- 65.00m (speed -1.00 mps /  course -1.00) @ 10/12/2015, 11:31:24 Greenwich Mean Time
> got the location: <+51.30969064,-0.12566690> +/- 65.00m (speed -1.00 mps / course -1.00) @ > 10/12/2015, 11:31:24 Greenwich Mean Time
> User Location: Executing -> Finishing
> User Location: did finish with no errors.
> User Location: Finishing -> Finished
> User Location <<< MutuallyExclusive<CLLocationManager>
 
Without any crashes.

[...]

PR #122, opened 2015-12-10
Crash in location log - Issue #121

danthorpe:


danthorpe:
This just adds an example of a `LocationManager` to the examples.

Issue #123, opened 2015-12-11
NSOperation's completionBlock is called twice.


difujia:
Try the following test. 
```swift
func testCompletionBlock() {
    let queue = OperationQueue()
    let blockOp = BlockOperation()
    let exp = expectationWithDescription("Block finished")
    blockOp.addCompletionBlock {
        exp.fulfill()
    }
    queue.addOperation(blockOp)
    waitForExpectationsWithTimeout(5, handler: nil)
}
```
Then the test runner will log:
>Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'API violation - multiple calls made to -[XCTestExpectation fulfill] for Block finished.'

The same error appear even if I set the `completionBlock` directly, which I think make no difference when only one completionBlock exists.

*Discussion* Xcode would not always stop the test as the error pops up, I think this is due to a race condition that the second call of `fulfill()` may happen either before or after `waitForExpectationsWithTimeout` has seen the effect of the first call (in case of after, the test has finished running and succeeded). Anyway, this is probably irrelevant of the issue. Isn't the `completionBlock` guaranteed to execute only once by Foundation framework?

danthorpe:
Hi @difujia, thanks for the feedback. This is a great spot, and may explain some of the flaky unit tests in #119. Will look into it today.

danthorpe:
Looking into this more closely. Have re-produces your exception. Hmm, I think you're right about a race condition with regard to fulfill - which explains why some of the tests are inconsistent (I need to refactor them).

But, also, there is this test: https://github.com/danthorpe/Operations/blob/development/Tests/Core/OperationTests.swift#L142 which admittedly doesn't test against multiple invocation of the completion block, but according to this bug, should be failing.

Also, I'll just add, that most of the tests fulfill their expectation using a BlockObserver (https://github.com/danthorpe/Operations/blob/development/Tests/Core/OperationTests.swift#L100) instead of a completion handler. This is because it's a little ill-defined exactly when that completion handler will be run by NSOperation.

Anyway, I will continue to investigate...

danthorpe:
Okay, so more info. This bug doesn't have anything to do with chaining completion blocks (as @difujia said). Changing the test case a little bit to see what's happening into this:

```swift
func test__block_operation_with_default_block_runs_completion_block_once() {
    let _queue = OperationQueue()
    let expectation = expectationWithDescription("Test: \(__FUNCTION__)")
    let operation = BlockOperation()
[...]

PR #125, opened 2015-12-11
Bug: completion block is called twice

danthorpe:
This is a work in progress on my investigation into this bug.

### DIFF
--- a/Sources/Core/Shared/Operation.swift
+++ b/Sources/Core/Shared/Operation.swift
@@ -92,20 +92,11 @@ public class Operation: NSOperation {
         }
         set (newState) {
             willChangeValueForKey("state")
-
-            log.verbose("\(operationName): \(_state) -> \(newState)")
-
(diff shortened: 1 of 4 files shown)


Sources/Core/Shared/Operation.swift
```diff
@@ -92,20 +92,11 @@ public class Operation: NSOperation {
...
         set (newState) {
             willChangeValueForKey("state")
-
-            log.verbose("\(operationName): \(_state) -> \(newState)")
-
-            stateLock.withCriticalScope { () -> Void in
-
-                switch (_state, newState) {
-                case (.Finished, _):
-                    break
```
danthorpe: This should actually be considered an invalid transition. As it turns out code never enters this path (which is good) but doesn't help us with the bug.

danthorpe:
Okay, so this fixes #123 - turns out that it's not `Operation` but `BlockOperation` and only when using the non-main-thread normal initializer, which includes the default argument.

PR #128, opened 2015-12-14
Fix issues with code coverage reports not being generated.

danthorpe:


danthorpe:
Restored coverage - turns out it was a temperamental issue with Codecov's uploader.

Issue #133, opened 2015-12-14
DelayOperation and BlockOperation lack cancellation handlers


jshier:
I don't know whether this was intentional, but it has lead to some unexpected behavior for me. A simple test like this:
```swift
let delayOperation = DelayOperation(interval: 5)
delayOperation.completionBlock = { print("delay completion block") }
delayOperation.addObserver(LoggingObserver())
let otherOperation = BlockOperation {
    print("other block completion")
}
otherOperation.addDependency(delayOperation)
otherOperation.addObserver(LoggingObserver())
let queue = OperationQueue()
queue.addOperations([delayOperation, otherOperation], waitUntilFinished: false)
queue.cancelAllOperations()
```
leads to both the `DelayOperation` and `BlockOperation` executing successfully. Looking at the code it seems like neither of these classes override `cancel` and so execute completely. Is this intentional?

danthorpe:
Hi @jshier - nope, not intentional at all. Consider this a bug, will fix it.

danthorpe:
Okay, looking more closely at `DelayOperation` - we do check the `cancelled` state inside the `dispatch_after` closure. Which is where the operation is finished. If it is cancelled, then `finish` has already been called. In which case, the completion block will have been executed. 

Note that from the `NSOperation` [docs](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/NSOperation_class/#//apple_ref/occ/instp/NSOperation/completionBlock) on `completionBlock`:

> A finished operation may finish either because it was cancelled or because it successfully completed its task. You should take that fact into account when writing your block code. Similarly, you should not make any assumptions about the successful completion of dependent operations, which may themselves have been cancelled.

So, despite the delay operation being cancelled, the above code should still print out `delay completion block` - again, I do see this as yet another unexpected (albeit documented) behavior of `NSOperation`. So - not a bug in `DelayOperation`, although in theory there is scope to improve the way the delay operation spends its time to support more frequent checking of `cancelled`.

In terms of `BlockOperation` - I think there is a bug here, because the `execute` implementation doesn't check the state of `cancelled`.

jshier:
I think some of this comes down to my (continued) misunderstanding of how the cancellation of operations works. I think, fundamentally, I was assuming that the cancellation of operations would invalidate any dependency chain established and dependent operations wouldn't even execute. (Which I think I've done before when reporting bugs here.) Now that I think about it, that's the reason for conditions, isn't it? 

[...]

PR #134, opened 2015-12-14
Cancellation issues with Delay and Block operations

danthorpe:
This is to address #133.

### DIFF
--- a/Tests/Core/OperationTests.swift
+++ b/Tests/Core/OperationTests.swift
@@ -428,6 +428,21 @@ class CancellationOperationTests: OperationTests {
 
         XCTAssertFalse(operation.didExecute)
     }
+
+    func test__operation_with_dependency_whole_queue_cancelled() {
+        let delay = DelayOperation(interval: 2)
+        delay.log.severity = .Verbose
+
+        let operation = TestOperation()
+        operation.log.severity = .Verbose
+
+        operation.addDependency(delay)
+
+        runOperations(delay, operation)
+        queue.cancelAllOperations()
+
+        XCTAssertFalse(operation.didExecute)
+    }
 }
 
 
--- a/Tests/Core/OperationTests.swift
+++ b/Tests/Core/OperationTests.swift
@@ -25,6 +25,7 @@ class TestOperation: Operation {
         simulatedError = error
         producedOperation = produced
         super.init()
+        name = "Test Operation"
     }
 
     override func execute() {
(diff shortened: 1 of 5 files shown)


Tests/Core/OperationTests.swift
```diff
@@ -379,4 +391,44 @@ class DelayOperationTests: OperationTests {
...
+        delay.cancel()
+
+        runOperations(delay, operation)
+        waitForExpectationsWithTimeout(5, handler: nil)
+
+        XCTAssertTrue(operation.didExecute)
+    }
+
+
+    func test__operation_with_dependency_cancelled_after_adding_does_not_execute() {
```
danthorpe: For this test, the output in the test runner looks like this:

```
Test Case '-[iOS_OperationsTests.CancellationOperationTests test__operation_with_dependency_cancelled_after_adding_does_not_execute]' started.
Delay for 2.0 seconds: Initialized -> Pending
Delay for 2.0 seconds: Pending -> Ready
Delay for 2.0 seconds: Ready -> Executing
Test Operation: Initialized -> Pending
Delay for 2.0 seconds: Did cancel.
Delay for Test Case '-[iOS_OperationsTests.CancellationOperationTests test__operation_with_dependency_cancelled_after_adding_does_not_execute]' passed (0.001 seconds).
2.0 seconds: Will Execute
```

Note that in this case, the delay still enters the executing state. But also that the dependent `TestOperation` never moves from of the `.Pending` state as far as we can tell.

Tests/Core/OperationTests.swift
```diff
@@ -379,4 +391,59 @@ class DelayOperationTests: OperationTests {
...
+
+        operation.addDependency(delay)
+
+        runOperations(delay, operation)
+        delay.cancel()
+
+        XCTAssertFalse(operation.didExecute)
+    }
+
+    func test__operation_with_dependency_whole_queue_cancelled() {
```
danthorpe: For this test, the output in the test runner looks like this:

```
Test Case '-[iOS_OperationsTests.CancellationOperationTests test__operation_with_dependency_whole_queue_cancelled]' started.
Delay for 2.0 seconds: Initialized -> Pending
Delay for 2.0 seconds: Pending -> Ready
Test Operation: Initialized -> Pending
Delay for 2.0 seconds: Ready -> Executing
Delay for 2.0 seconds: Did cancel.
Test Case '-[iOS_OperationsTests.CancellationOperationTests test__operation_with_dependency_whole_queue_cancelled]' passed (0.001 seconds).
Test Operation: Did cancel.
Delay for 2.0 seconds: Will Execute
Test Operation: Pending -> Finishing
Test Operation: Did finish with no errors.
Test Operation: Finishing -> Finished
```

PR #140, opened 2015-12-16
Make OperationObserver protocol more granular

danthorpe:


danthorpe:
This includes some fairly significant & in some cases breaking changes.

1. `OperationObserver` has been refactored to be a refinement of 4 other observer protocols. Each protocol is responsible for a specific event. These are *started*, *cancelled*, *produced* and *finished*.

2. Each protocol refines `OperationObserverType` which is an empty protocol. `Operation` will accept any observer conforming to this type.

3. There are now 4 new block based observers, each receives a non-optional block and conforms to the appropriate observer protocol. E.g. `FinishedObserver` takes a single block which is executed when the operation finishes.

4. `BlockObserver` now composes optional instances of these 4 other observers and is instantiated with optional block types.

However, in general, I don't think `BlockObserver` is a very Swift like type. Instead I recommend writing clear observers and using multiple specific instances with non-optional blocks.

The breaking changes are the constructor for `BlockObserver` which has changed slightly. But, framework consumers can take this opportunity to refactor their code and use a more appropriate observer. For example, if previously you had this...

```swift
operation.addObserver(BlockObserver { _, errors in
   // handle some errors here
})
```

This is better expressed as:

```swift
operation.addObserver(FinishedObserver { _, errors in
   // handle some errors here
})
```

PR #142, opened 2015-12-17
'ready' should be used to validate `addCondition` and `addObserver`

difujia:
Improvements: conditions and observers should be added before operations is 'ready' instead of 'executing'.

Despite the fact that the time window for state transition from 'ready' to 'executing' is rather short, 'ready' is the more accurate state for last chance of adding conditions and observers, because condition expresses 'readiness' and observer relies on 'readiness' to receive all life cycle events.


danthorpe:
Hi @difujia - can you pull development - these asserts have become precondition. Then should be good to go.

Agree with you on your main point.

difujia:
Done that.

danthorpe:
Merged, thanks @difujia!

PR #137, opened 2015-12-16
Result Injection

danthorpe:
This is a work in progress for the discussion in #135.

- [x] Basic implementation
- [x] Add `addOperations(operations: NSOperation...)` to NSOperationQueue.
- [x] Tests

PR #143, opened 2015-12-17
Proposed implementation to pipe objects among operations.

difujia:
Proposed implementation of a way to allow operations to retrieve resources provided by some objects which could be other operations.

This is a lightweight approach to allow objects be transferred from one operation to another. 

Those objects are considered to be resources and can be provided by `OperationsResourceProvider`s. The mechanism is very similar to `OperationsCondition`. Their commonality is that, first they both may produce a dependency operation which may in turn produce a resource. Second, **all** conditions and resource providers are consulted after their dependency operations have finished execution, so condition results and resources reflect the results of dependencies. Third, `OperationCondition` and `OperationResourceProvider` abstract away how the condition results and resources are generated. So we know conditions do not have to use an operation to determine the results, resource providers do not have to use an operation to produce resources either (i.e. resources may be produced just before a `NeedResourceOperation` become ready).

The only difference between `OperationCondition` and `OperationResourceProvider` is that they produce different values. Conditions produce boolean values and resource providers produce arbitrary objects. Another difference to note is that conditions must be met for an operation to proceed, while resource providers do not have to produce the desired objects at all. An `NeedResourceOperation` has total control whether it wants to proceed when a resource is absent.

In this implementation, the three parties (i.e. consuming operation, resource provider and producing operation) are self-contained. Consuming operation retrieves resources by type during execution, and controls how to proceed. Resource provider knows how to produce a resource (likely to delegate to a producing operation). Producing operation can be any normal operation and produce their results via the completion block as usual. 

Obviously, this is a dirty implementation. The main intention is to show my understanding of these concept and my personally preferred usage. Any thoughts are welcome.

### DIFF
--- a/Operations.xcodeproj/project.pbxproj
+++ b/Operations.xcodeproj/project.pbxproj
@@ -225,6 +225,8 @@
 		6526197B1C11FE6D00654091 /* AddressBookConditionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6526171D1C11F5EB00654091 /* AddressBookConditionTests.swift */; };
 		6526197C1C11FE6D00654091 /* AddressBookTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6526171E1C11F5EB00654091 /* AddressBookTests.swift */; };
 		6526197D1C11FE6D00654091 /* ContactsOperationsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 652617201C11F5EB00654091 /* ContactsOperationsTests.swift */; };
+		BF1E62D81C224CF800728E25 /* OperationResource.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF1E62D71C224CF800728E25 /* OperationResource.swift */; };
+		BF1E62DA1C226CB100728E25 /* OperationResourceTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF1E62D91C226CB100728E25 /* OperationResourceTests.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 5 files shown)


danthorpe:
Hi @difujia I'll look through your proposal, but can you review the recent changes in the project. Specifically #137.

difujia:
Hi @danthorpe, yeah I've noticed that. I will try it later.

danthorpe:
Hi @difujia. Well, I'll be honest, I'm not sure about this. My view is this:

If an operation has a resource requirement, that resource can either be made available synchronously or asynchronously.

For synchronously available resources - it is better to use established dependency injection patterns, i.e. an argument in the initializer. And there are already frameworks for more sophisticated DI patterns. For code reuse regarding synchronous resource requirements - this framework is already heavily reliant on inheritance patterns (thanks to `NSOperation`), so it is best to use that or perhaps protocols and default implementations.

For *asynchronously* available resources - these should be provided by an `Operation` subclass - which is what we agree on. `NSOperation` already has a mechanism for dependencies which we can use to configure scheduling. All that is required is moving the resource from the dependency (or result as I've called it) to the requirement of the operation.

I think adding `OperationResourceProvider` increases the complexity and overall "cognitive load" for developers - there is already the concept of observers and conditions. So, I don't think the additions provide enough advantages to merit the increased complexity. Additionally, although I'm less sure on this, I think these changes would result in application code which is less clear and direct - there is yet another element of asynchronous "magic" going on.

In #137 I used a common metaphor of `DataRetieval` and `DataProcessing` classes to explain how results were set as requirements. Could you describe how your proposal would achieve the same result? 

One aspect of your proposal is that `OperationResourceProvider` is a protocol - so it'll need a type which implements it. Does this mean that every pair of participating operations needs another type? If so - this seems like far too much boilerplate. Obviously we could use a generic type and introduce more protocols. But at this point it starts to look very much like what I've already done, but instead of:

```swift
let retrieval = DataRetrieval()
let processing = DataProcessing()
processing.injectResultFromDependency(retrieval)
```

you have:

```swift
let retrieval = DataRetrieval()
let processing = DataProcessing()
processing.addResourceProvider(ResourceProvider(from: retrieval, to: processing))
```

which to me is a lot less clear what will happen.

Anyway, happy to discuss this further, if there are particular usage cases which I'm not seeing, please provide some example or help me understand why you think this proposal is warranted.

Thanks for the efforts!

danthorpe:
Gonna close this for now.

Issue #144, opened 2015-12-17
Any way to make logger optional?


shsteven:
First of all, thank you @danthorpe  for creating this library!
It's even better than the original sample code thanks to all the improvements.

When using both Operations and a logging framework e.g. SwiftyBeaver, log.debug() stops working because both frameworks declare the same function.

Is there any way to make LoggerType optional?

danthorpe:
Hi @shsteven - thanks a lot, much appreciated

In terms of disabling logging (when using `development` branch) - you can put `LogManager.enabled = false` - this will be made available in the next version, hopefully soon.

But, I don't think this is the issue, there is a name spacing clash here? Which is what I don't quite follow, however I don't think you're the only person to have this issue - and I'd really like to get to the bottom of it. Can you provide a code example of how you're using logging and what is going on?

`log` is a property on `Operation` - if you have a function called `log` then you can still use it inside an `Operation` subclass by specifying the module that it is defined in. E.g. 

```swift
class MyOperation: Operation {
    func execute() {
        log.info("This is an info message")
        SwiftyBeaver.log.debug("This is a log from another framework")
    }
}
```

On the `development` it's also possible to make the built in logger to pipe through to a custom logging framework. Although, this might be lacking in some features (e.g. severity). But essentially, you just need to set the `logger` property.

```swift
LogManager.logger = { message in
    SwiftBeaver.print(message)
}
```

I don't wish to drop the built in logger - because I actually find it really useful for developing and debugging the framework. Equally I can't really support every different external logging tool. So, I opted to make the cleanest separation, where the severity and queue was defined internally, but the block which actually processes the received message could be overridden.

One improvement I can make to this is to also pass the severity through to that block, which would allow for more flexibility is adapting it to an external system.

danthorpe:
Also - I'm trying to avoid having any dependencies with Operations - hence why it doesn't use SwiftyBeaver itself.

shsteven:
Hi @danthorpe ,

Thank you for your quick response!

Here's a concrete example:

```swift
import Operations
import SwiftyBeaver

class SampleOperation: Operation {
    override func execute() {
        log.info("This is an info message that works fine")
        log.debug("This doesn't work")
        SwiftyBeaver.debug("This works")
    }
}
```

Prefixing with framework name works.
And I completely agree seeing operation logs is very useful. 

However, it feels a bit awkward to use `SwiftyBeaver.debug()` instead of `log.debug()`. 
[...]

PR #145, opened 2015-12-17
Customizing the logger

danthorpe:


examples/Permissions/Permissions/AppDelegate.swift
```diff
@@ -16,8 +19,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            case .Info:
+                log.info(message, file, function, line: line)
+            case .Warning:
+                log.warning(message, file, function, line: line)
+            case .Fatal:
+                log.error(message, file, function, line: line)
+            }
+        }
+        
+        LogManager.severity = .Info
```
danthorpe: Okay, so here is how to configure the internal logger to use an external logging framework such as SwiftyBeaver.

And so now logging output looks like this....

custom logging

The only thing that is perhaps a little bit annoying, is that the various severity levels might not always match up. Operations defines:

```swift
public enum LogSeverity: Int, Comparable {

    /// Chatty
    case Verbose = 0

    /// Public Service Announcements
    case Notice

    /// Info Bulletin
    case Info

    /// Careful, Errors Occurring
    case Warning

    /// Everything Is On Fire
    case Fatal
}
```

And, SwiftyBeaver defines:

```swift
    public enum Level: Int {
        case Verbose = 0
        case Debug = 1
        case Info = 2
        case Warning = 3
        case Error = 4
    }
```

So, you'll have to match things up...

danthorpe:
Hi @shsteven - what are your thoughts on this? See the changes in the AppDelegate...

shsteven:
Looks great!

Matching up the logging levels is fine, as long as all logs are captured.

Do you mind adding a short description to README? Something like

> Integrating with other logging libraries
> LogManager can be set up to pipe logs to other logging libraries, such as XCGLogger or SwiftyBeaver. Sample code is available in AppDelegate.swift in the Permissions example app.

shsteven:
#144

danthorpe:
@shsteven agreed that it needs to be documented - I'm working on overhauling the documentation separately and will have a place for all of this.

Issue #153, opened 2015-12-29
A Polling Operation


jshier:
My current project has several network endpoints that I'm required to poll against. This is due to backend design, not desire, and I have no control over this requirement. As the number of endpoints I need to poll against has increased, my need for a generic solution has as well. So I've started doing so and while I have a prototype working, additional guidance would be appreciated. Here's what it looks like so far:
```swift
import Operations

typealias OperationGenerator = () -> Operation

class PollingOperation : GroupOperation {
    var activeOperation: Operation
    var delayOperation: DelayOperation
    
    private let operationGenerator: OperationGenerator
    private let interval: NSTimeInterval
    
    init(interval: NSTimeInterval, operationGenerator: OperationGenerator) {
        self.interval = interval
        self.operationGenerator = operationGenerator
        activeOperation = operationGenerator()
        delayOperation = DelayOperation(interval: interval)
        super.init(operations: [])
    }
    
    override func execute() {
        continuePolling()
        
        super.execute()
    }
    
    override func operationDidFinish(operation: NSOperation, withErrors errors: [ErrorType]) {
        if let pollable = operation as? Pollable {
            if pollable.shouldContinuePolling() {
                continuePolling()
            }
            else {
                finish()
            }
        }
    }
    
    func continuePolling() {
        delayOperation = DelayOperation(interval: interval)
        activeOperation = operationGenerator()
        activeOperation.addDependency(delayOperation)
        
        addOperations([delayOperation, activeOperation])
    }
}

protocol Pollable {
    func shouldContinuePolling() -> Bool
}
```
My main issue at the moment is how to make the `Pollable` protocol fully generic so I can add conformance to any arbitrary operation. Here's the current `TestOperation`:
```swift
class TestOperation : Operation {
    
    override init() {
        super.init()
        name = "Test Operation"
    }
    
    override func execute() {
        print("*** Executing TestOperation. ***")
        finish()
    }
}

extension TestOperation : Pollable {
    func shouldContinuePolling() -> Bool {
        let timeInterval = NSDate().timeIntervalSince1970
        let remainder = Int64(timeInterval) % 10
        
        print("*** shouldContinuePolling() for interval: \(timeInterval), remainder: \(remainder). ***")
        return (remainder != 0)
    }
}
```
Called like this:
```swift
let pollingOperation = PollingOperation(interval: 1) { () -> Operation in
    return TestOperation()
}
let queue = OperationQueue()
queue.addOperation(pollingOperation)
```

[...]

PR #154, opened 2015-12-30
Retry Operation

danthorpe:
Here are my thoughts on a possible `RetryOperation`.

1. It would be a way of reliably executing an operation which can "fail" (finish with errors).
2. It would be a little bit like `ReachableOperation` / `ComposedOperation` where the target operation is composed inside a group operation.
3. The target operation gets a `NoFailedDependenciesCondition` added
4. If the target operation finishes with any errors, another instance of the same operation is added to the group after a delay
5. When creating the `RetryOperation`, the initializer should accept a (minimum) time delay, a flag for whether the delay should be exponentially backed-off (is that a term?).
6. Perhaps also define a maximum number of attempts - but default to `.None`.

It might also be useful to architecture this as a subclass of `RepeatedOperation` which is basically just like above, except it just executes instances of the same operation a number of times. Would that be useful? I don't know....

### DIFF
--- a/Sources/Core/Shared/RepeatedOperation.swift
+++ b/Sources/Core/Shared/RepeatedOperation.swift
@@ -178,6 +178,7 @@ struct FibonacciWaitGenerator: GeneratorType {
     }
 }
 
+
 /**
  Define a strategy for waiting a given time interval. The strategy
  can then create a NSTimeInterval generator. The strategies are:
@@ -289,7 +290,7 @@ public enum WaitStrategy {
  )
  ```
 
- - See: WaitStrategy
+ - See: Wait
  - See: Repeatable
 
 */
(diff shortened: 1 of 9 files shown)


Sources/Core/Shared/RepeatedOperation.swift
```diff
@@ -0,0 +1,122 @@
...
+    internal func nextDelayOperation() -> DelayOperation? {
+        return delay.next().map { DelayOperation(interval: $0) }
+    }
+}
+
+public protocol RepeatingOperationType {
+    var shouldRepeat: Bool { get }
+}
+
+public class RepeatingOperationGenerator<O where O: NSOperation, O: RepeatingOperationType>: GeneratorType {
```
danthorpe: Okay, so I need to actually write the tests for this to see if it will work, but the idea behind this is to have an `Operation` subclass, where the previous instance determines whether or not it should be run again. This is to implement something like a `PollingOperation` as described in #153.

```swift
class TestOperation: Operation, RepeatingOperationType {
    var shouldRepeat: Bool {
        // etc
    }
}

let operation = RepeatedOperation(generator: RepeatingOperationGenerator { return TestOperation() })
queue.addOperation(operation)
```

danthorpe: Clearly, we'll clean this up a bit in terms of convenience so that it would/could just be:

```swift
class TestOperation: Operation, RepeatingOperationType {
    var shouldRepeat: Bool {
        // etc
    }
}

let operation = RepeatedOperation { return TestOperation() }
queue.addOperation(operation)
```

danthorpe:
This branch adds `RepeatedOperation` and its subclass `RetryOperation`.

### `RepeatedOperation`

This operation must be initialized with a [generator](https://developer.apple.com/library/mac/documentation/Swift/Reference/Swift_GeneratorType_Protocol/index.html) which has an element which subclasses `NSOperation`. e.g.

```swift
let operation = RepeatedOperation(anyGenerator { return MyOperation() })
``` 

The operation is a `GroupOperation` subclass which works by adding new instances of the operation to its group. This happens initially when the group starts, and then again when the child operation finishes.

After the initial child operation completes, new operations are added with a delay on the queue. The time interval of the delay can be configured via the first argument to the `RepeatedOperation`.

There are two ways to stop the operations from repeating.

1. Return `nil` from the generator passed to the initializer
2. Set the 2nd argument, `maxNumberOfAttempts` which is an optional `Int` defaulting to `.None`.

The first argument is a `WaitStrategy` which is an enum of various different mechanisms for defining waiting. These are: `.Fixed`, `.Random`, `.Incrementing`, `.Exponential` and `.Fibonacci`. Each of the strategies accept arguments which govern how they work. For example, `.Fixed` takes a single `NSTimeInterval`, `.Random` accepts minimum and maximum intervals, etc etc.

For example, to use exponential back-off, with a maximum of 10 attempts:

```swift
let operation = RepeatedOperation(strategy: .Exponential((minimum: 1, maximum: 300)), maxNumberOfAttempts: 10, anyGenerator { return MyOperation() })
```

#### `Repeatable`

`Repeatable` is a very simple protocol, which your `NSOperation` subclass can conform to. This allows the previous operation to define whether a new one should be executed. For this special case, `RepeatedOperation` can be initialized like this:

```swift
let operation = RepeatedOperation { return MyRepeatableOperation() }
```

and

```swift
let operation = RepeatedOperation(strategy: .Random((minimum: 0.1, maximum: 0.5))) { return MyRepeatableOperation() }
```

The trailing closure syntax is not available for standard `NSOperation` subclasses - you'll have to use `anyGenerator`.

Note that `RepeatedOperation` ignores any errors that the child operation might generate - it simply keeps on repeating. This makes it ideal for "polling" style functionality, and operations which are idempotent.

#### `RetryOperation`

`RetryOperation` is a subclass of `RepeatedOperation` except for two significant differences.

1. The composed operation type must be an `Operation` subclass (not just `NSOperation`). This is because it will automatically add `NoFailedDependenciesCondition` to the child instances.
2. It only adds new child operations when the previous one finished with errors. Everything else is the same. This means, that if the generator returns `nil` - it will also end even if the previous child finished with an error.

This makes it ideal of cases where an `Operation` can fail, but must be executed only once - i.e. once it succeeds then stop.

#### `ReachableOperation`

Do not uses these operations for network / reachability related operations. For these cases, continue to use `ReachableOperation`. `ReachableOperation` will be refactored to use `RetryOperation` internally.

PR #163, opened 2016-01-08
Refactor ReachableOperation

danthorpe:
- [x] Subclass `ComposedOperation`
- [x] Refactor `Reachability` to use function callbacks, instead of polling.
- [x] Maximize test coverage, currently:

    File | % Before | % After
    ------|------------|------------
    ReachilityCondition.swift | 82% | 86%
    ReachableOperation.swift | 54% | 92%
    Reachability.swift | 39% | 62%

### DIFF
--- a/Sources/Features/Shared/Reachability.swift
+++ b/Sources/Features/Shared/Reachability.swift
@@ -11,6 +11,14 @@ import SystemConfiguration
 
 public struct Reachability {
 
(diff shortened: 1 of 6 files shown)


danthorpe:
Unfortunately it's not possible to make the Coverage Patch check pass (> 75% coverage) because the majority of the code which is refactored, is inherently untestable SystemConfiguation function calls.

PR #165, opened 2016-01-11
Refactor CloudKitOperation, phase 3

danthorpe:
### Phase Three

- [x] Refactor `ReachableOperation`

Refactor `ReachableOperation`, to subclass `ComposedOperation`, and then `CloudKitOperation` should subclass `ReachableOperation`.

Given this, the `ReachableOperation` will manage errors/executing the operation when the app is online.

Execution inside of a `RetryOperation` will support errors/executing the operation in the face of errors from CloudKit. Which leaves `CloudKitOperation` to support the configuration of the underlying target operation.

### Capabilities - `AuthorizedFor` condition.

It seems a bit silly that consumers should be adding an `AutorizedFor(Capability.Cloud())` condition to their `CloudKitOperation` instances. Instead, it might make sense to support initializing `CloudKitOperation` with permissions (with a default) and a container id (with a default). This way, we can automatically add a condition.

- ~~[ ] Automatically add condition for authorized access to the container.~~
- ~~[ ] Automatically manage authorized condition when subsequently setting the container.~~

**Not easily possible** without introducing another generic type, as `Capability.Cloud` cannot be created in a unit test environment. Might be able to eventually do with with some re-factoring of the `_CloudCapability`.

### Greedy Batch Processing

There are some `CKOperation` types which have the concept of "more coming" with "result limits". This means that the operation can be configured to reduce the amount of processing that must be done any any single operation instance. This can the useful to mitigate the risk of data taking longer to get to the device and then consequently either not being an accurate representation any more, or bogging down the network. So, setting a reasonable batch size (`resultsLimit`) and then re-requesting if there are more results is a sensible approach.

The framework should have automatic support for this - essentially repeating itself until `moreComing` is false.

- [x] Greedy Batch Processing
    - [x] `CKFetchNotificationChangesOperation`
    - [x] `CKFetchRecordChangesOperation `

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ DerivedData
 # http://guides.cocoapods.org/using/using-cocoapods.html#should-i-ignore-the-pods-directory-in-source-control
 #
 # Pods/
+Examples/*/Pods
 
 # Carthage
 #
--- a/Examples/Last Opened/Last Opened.xcodeproj/project.pbxproj
+++ b/Examples/Last Opened/Last Opened.xcodeproj/project.pbxproj
@@ -1,0 +1,371 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 46;
+	objects = {
+
+/* Begin PBXBuildFile section */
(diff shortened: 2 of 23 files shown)


danthorpe:
Okay, going to merge this before it gets too stale. Error Handling will be done in #167.

PR #168, opened 2016-01-13
Refactor RepeatedOperation & RetryOperation

danthorpe:
At the moment, `RepeatedOperation` uses two generators, one for delays and another for the operation, which in hindsight is a little bit limiting.

A better approach would be a generator of type `(NSTimeInternal?, T)` or `(NSDate?, T)`... so that the generator can be responsible for injecting it's own delay.

Another factor would be how to take latest and aggregate errors into account.

### DIFF
--- a/Sources/Core/Shared/DelayOperation.swift
+++ b/Sources/Core/Shared/DelayOperation.swift
@@ -8,6 +8,35 @@
 
 import Foundation
 
+public enum Delay {
+    case By(NSTimeInterval)
+    case Until(NSDate)
+}
+
+extension Delay: CustomStringConvertible {
+
+    public var description: String {
+        switch self {
+        case .By(let _interval):
+            return "for \(_interval) seconds"
+        case .Until(let date):
+            return "until \(NSDateFormatter().stringFromDate(date))"
+        }
+    }
+}
+
+internal extension Delay {
+
+    var interval: NSTimeInterval {
+        switch self {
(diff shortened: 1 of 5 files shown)


danthorpe:
Change in this PR:

> Now, the generator element for Repeated is a tuple: (Delay?, T) which means that at the lowest level, which has the most flexible, it is possible to provide a generator which does the following:

> 1. return .None to stop repeating, i.e. end the RepeatedOperation.
2. return (.None, operation) to add the operation to the queue immediately.
3. return (.By(1.0), operation) to add the operation to the queue, but depending on a 1 second delay.
4. return (.Until(date), operation) to add the operation to the queue, but depending until a date in the future.

Additionally, `RetryOperation` is a thin facade over this structure which accepts an error handler.The handler is this type:

```swift
typealias Handler = (RetryFailureInfo<T>, Delay?, T) -> (Delay?, T)?
```

The failure info value encapsulates error information related to all the previous failed operation instances. It can be inspected to adjust the return tuple. For example to modify the delay, or provide a completely new operation instance, or to add dependency operations.

For example:

```swift
typealias Handler = RetryOperation<MyOperation>.Handler
let handler: Handler = { info, delay, op in
    let extra = AnExtraOperation()
    op.addDependency(extra)
    // Will add another operation to the group's queue separate to the primary operation
    info.addOperations(extra)
    return (delay, op)
}
```

PR #167, opened 2016-01-13
CloudKitOperation: Error Handling

danthorpe:
The 4th phase of CloudKitOperation refactoring.

- [x] `CloudKitOperation` now subclasses `RetryOperation`.
- [x] Configuration still works. - possibly copy `addConfigurationBlock` concepts from `BatchedCloudKitOperation`.
- [x] `BatchedCloudKitOperation` should subclass `RepeatedOperation`?

### Default Error Recovery

I think it would be great if the framework provided some convenient error handling for some of the more recoverable errors. Things like, not logged into iCloud for example. My thinking is a type `CloudKitErrorHandler` say, which can be used in conjunction with `RetryOperation` as the "error recovery block".

- [x] Cloud Kit Error Handling

There are going to be some pretty standard error responses.

1. Fatal Log the error, and return `.None` (i.e. do not attempt to retry).
    - [x] `InternalError`
    - [x] `MissingEntitlement`
    - [x] `InvalidArguments`
    - [x] `ServerRejectedRequest`
    - [x] `AssetFileNotFound`
    - [x] `IncompatibleVersion`
    - [x] `ConstraintViolation`
    - [x] `BadDatabase`
    - [x] `QuotaExceeded`
    - [x] `OperationCancelled`

2. Re-request after a delay.
    - [x] `NetworkUnavailable`
    - [x] `NetworkFailure`
    - [x] `ServiceUnavailable` 
    - [x] `RequestRateLimited`
    - [x] `AssetFileModified`
    - [x] `BatchRequestFailed`
    - [x] `ZoneBusy`

These will require a bit more thought:

- [ ] `PartialFailure`
    Some items failed, but the operation succeeded overall
    Do not handle in anyway - should be handled by the framework consumer.

- [ ] `BadContainer`
    Un-provisioned or unauthorized container. Try provisioning the container before retrying the operation.
    TODO: Look into what provisioning the container means.
    Exit immediately - with fatal error logs, this is a compile time issue.

- [ ] `NotAuthenticated`
    Not authenticated (writing without being logged in, no user record)

- [ ] `PermissionFailure`
    Access failure (save or fetch)
    This sounds like we could handle it automatically, with an Operation/Block which is an alert telling the user to log into iCloud. So, we don't retry, but we do add an `AlertPresentation` operation.

- [ ] `UnknownItem`
    Record does not exist

- [ ] `ResultsTruncated`
    Query results were truncated by the server
    Automatically adjust the parameters to fetch the next set of results?

- [ ] `ServerRecordChanged`
    The record was rejected because the version on the server was different
    Automatically adjust the parameters and re-request?

- [ ] `ChangeTokenExpired`
    The previousServerChangeToken value is too old and the client must re-sync from scratch
    Trigger a re-sync operation?

- [ ] `ZoneNotFound`
    The specified zone does not exist on the server
    Add a new operation to create the Zone first, with dependencies setup.

- [ ] `LimitExceeded`
    The request to the server was too large. Retry this request as a smaller batch.
    Automatically re-request as multiple batches?

- [ ] `UserDeletedZone`
    The user deleted this zone through the settings UI. Your client should either remove its local data or prompt the user before attempting to re-upload any data to this zone.
    Retry with a UserConfirmationOperation?


danthorpe:
So, the way the error handling will work is something like this, (or will be)

```swift
// Fetch (all) Record Zones CloudKit Operation
let operation = CloudKitOperation { CKFetchRecordZonesOperation.fetchAllRecordZonesOperation() }

// Add custom error handlers
operation.setErrorHandlerForCode(.ConstraintViolation) { error, recommended in
    // Inspect the error
    // The recommended tuple is the "recommended" response. It might be .None.
    // This is the result of the default error handlers being applied to the "next"
    // delay & operation (using the generator provided at initialism).

    // For errors such as this (a constraint violation) you'll typically want to log the
    // info, and return .None (which is actually what the default error handler would do). 
}

// Add error handlers for any CKErrorCode 
operation.setErrorHandlerForCode(.PartialFailure) { error, recommended in
    // etc
}
```

danthorpe:
Okay, going to merge this now, and work on a CloudKit example in another branch.

PR #177, opened 2016-02-06
CloudKitOperation add header documentation

danthorpe:


Sources/Features/Shared/CloudKitOperation.swift
```diff
@@ -276,7 +585,97 @@ public class CloudKitRecovery<T where T: NSOperation, T: CKOperationType> {
...
+ ```swift
+ operation.setErrorHandlerForCode(.PartialFailure) { error, log, suggested in
+    return suggested
+ }
+ ```
+
+ Note that the error handler receives, the received error, a logger object, and the suggested tuple, which
+ could be modified before being returned. Alternatively, return nil to not retry.
+
+*/
```
danthorpe: @StephenDaedalus does this makes sense?

StephenDaedalus: Sounds pretty cool. I'm sure I'd almost always retry. 

Question: in my Objective C version I never fussed a lot with Reachability because I thought that I could use the CloudKit's own error codes to deal with setting up repeated attempts at the operations. Is that accurate?

Issue #175, opened 2016-02-03
Completion block isn't called in ~0.1% of cases when a condition is attached.


bastianschilbe:
If you have a operation and prevent it from running by a failed condition, the completion callback isn't always called. Try the following with an iOS test target:
```
    func test__multiple_operations_completion_blocks() {
        
        let q = OperationQueue()
        (0...5000).forEach { i in
            let e = expectationWithDescription("wait\(i)")
            let operation = BlockOperation(block: { _ in XCTFail() })
            operation.addCompletionBlock({ e.fulfill() })
            let condition = BlockCondition(block: { false })
            operation.addCondition(condition)
            q.addOperation(operation)
        }
        waitForExpectationsWithTimeout(8, handler: nil)
    
    }
```
I see about 1-10 expections unfulfilled.
Do I use conditions the wrong way?

danthorpe:
Hmm, interesting - I'll have a look into this, too.

There isn't anything special in the framework regarding `completionBlock` - it's just the same behavior as `NSOperation`. The documentation is a little vague about when the completion block is executed. Have you written this test because you've noticed some odd or random behavior?

I tend to use `FinishedObserver` instead of completion blocks, as it's more clearly defined when it is executed. E.g.

```swift
func test__multiple_operations_completion_blocks() {

    let q = OperationQueue()
    (0...5000).forEach { i in
        let e = expectationWithDescription("wait\(i)")
        let operation = BlockOperation { XCTFail() }
        operation.addObserver(FinishedObserver { _, _ in e.fulfill() })
        let condition = BlockCondition { false }
        operation.addCondition(condition)
        q.addOperation(operation)
    }
    waitForExpectationsWithTimeout(8, handler: nil)
}
```

But, regardless, your test should work. Might just be an issue with your `BlockOperation`.

bastianschilbe:
Yes I've tested it with the FinishObserver with the same result.
I've some nested GroupOperations and some operations in them have a condition which is failing and the finishedObserver isn't called. I've already spent some time to track down the issue but I'm kind of lost. It seems to be some sort of race condition. If I slow down the .forEach loop by a logging breakpoint with each iteration the chances that more expectations are fulfilled are raising. Hope you can help with that. If I can help you somehow let me know

danthorpe:
[...]

PR #180, opened 2016-02-09
Investigating completion block not firing bug.

danthorpe:
This is an investigation of #175 which is a bug report indicating that completion blocks are not always called.

### DIFF
--- a/Operations.xcodeproj/xcshareddata/xcschemes/OS X Operations.xcscheme
+++ b/Operations.xcodeproj/xcshareddata/xcschemes/OS X Operations.xcscheme
@@ -38,6 +38,11 @@
                BlueprintName = "OS X OperationsTests"
                ReferencedContainer = "container:Operations.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "CompletionBlockOperationTests/test__many_completion_blocks_are_executed()">
+               </Test>
+            </SkippedTests>
          </TestableReference>
       </Testables>
       <MacroExpansion>
--- a/Operations.xcodeproj/xcshareddata/xcschemes/iOS Operations.xcscheme
+++ b/Operations.xcodeproj/xcshareddata/xcschemes/iOS Operations.xcscheme
@@ -38,6 +38,11 @@
                BlueprintName = "iOS OperationsTests"
                ReferencedContainer = "container:Operations.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "CompletionBlockOperationTests/test__many_completion_blocks_are_executed()">
+               </Test>
+            </SkippedTests>
          </TestableReference>
       </Testables>
       <MacroExpansion>
--- a/Sources/Core/Shared/Operation.swift
+++ b/Sources/Core/Shared/Operation.swift
@@ -75,15 +75,15 @@ public class Operation: NSOperation {
 
     // use the KVO mechanism to indicate that changes to "state" affect other properties as well
     class func keyPathsForValuesAffectingIsReady() -> Set<NSObject> {
-        return ["state"]
+        return ["State"]
     }
 
     class func keyPathsForValuesAffectingIsExecuting() -> Set<NSObject> {
-        return ["state"]
+        return ["State"]
     }
 
     class func keyPathsForValuesAffectingIsFinished() -> Set<NSObject> {
-        return ["state"]
+        return ["State"]
     }
 
     class func keyPathsForValuesAffectingIsCancelled() -> Set<NSObject> {
@@ -105,13 +105,13 @@ public class Operation: NSOperation {
             return stateLock.withCriticalScope { _state }
         }
         set (newState) {
-            willChangeValueForKey("state")
+            willChangeValueForKey("State")
             stateLock.withCriticalScope {
                 assert(_state.canTransitionToState(newState, whenCancelled: cancelled), "Attempting to perform illegal cyclic state transition, \(_state) -> \(newState).")
                 log.verbose("\(_state) -> \(newState)")
                 _state = newState
             }
-            didChangeValueForKey("state")
+            didChangeValueForKey("State")
         }
     }
 
@@ -262,7 +262,7 @@ public class Operation: NSOperation {
         assert(state == .Pending, "\(__FUNCTION__) was called out of order.")
         assert(cancelled == false, "\(__FUNCTION__) was called on cancelled operation: \(operationName).")
         state = .EvaluatingConditions
-        OperationConditionEvaluator.evaluate(conditions, operation: self) { errors in
+        evaluateOperationConditions(conditions, operation: self) { errors in
             self._internalErrors.appendContentsOf(errors)
             self.state = .Ready
         }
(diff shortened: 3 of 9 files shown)


danthorpe:
With the code as it currently is, where there is a condition attached to the operation which fails - the test fails. Approximately 0.2% of the operations do not have their completion block executed after the operation finishes.

However, if the condition line is commented out, the test passes.

danthorpe:
Okay, so there is still something going on here. Definitely a race condition. Test passes on Mac OS X, Fails on iOS, about 0.1% of operations do not have their `state` proceed past `.Ready`.

The state moves to `.Ready` however - the operation does not begin...

danthorpe:
So, it's not a case of the evaluator not returning, as the state is set to `.Ready` - it's now just a case of the the KVO mechanisms not kicking in for some reason.

danthorpe:
So... for example, if we set the severity to `.Verbose`.

>Interation: 4590: Initialized -> Pending
>Interation: 4589: Finishing -> Finished
>Interation: 4590: Pending -> EvaluatingConditions
>Interation: 4590: EvaluatingConditions -> Ready
>Interation: 4591: Initialized -> Pending
>Interation: 4590: Ready -> Finishing
>Interation: 4591: Pending -> EvaluatingConditions
>Interation: 4590: Did finish with no errors.
>Interation: 4591: EvaluatingConditions -> Ready
>Interation: 4590: Finishing -> Finished
>Interation: 4592: Initialized -> Pending
>Interation: 4592: Pending -> EvaluatingConditions
>Interation: 4592: EvaluatingConditions -> Ready
>Interation: 4593: Initialized -> Pending

Where iteration 4591 is the failed iteration. If we separate out the iterations:

>Interation: 4591: Initialized -> Pending
>Interation: 4591: Pending -> EvaluatingConditions
>Interation: 4591: EvaluatingConditions -> Ready

And `Iteration: 4591` does not progress further from here.

>Interation: 4590: Initialized -> Pending
>Interation: 4590: Pending -> EvaluatingConditions
>Interation: 4590: EvaluatingConditions -> Ready
>Interation: 4590: Ready -> Finishing
>Interation: 4590: Did finish with no errors.
>Interation: 4590: Finishing -> Finished

So, the question is.... why would stop the KVO from kicking in?

danthorpe:
And yet sometimes.... it works.

@ansf, @jshier have you got any thoughts on this? I'm running out of ideas here.

Something that I have done, but didn't commit, was actually to also wrap up all the internal state (like, the conditions, observers, logger, cancelled flag, etc) in `Operation` in to a many-readers-one-writer thread safe structure. However, this actually caused way more problems and still didn't solve this original issue.

jshier:
Considering the same works fine on OS X it seems like an underlying KVO issue to me. Is KVO compliance necessary for triggering underlying `NSOperation` functionality? If not I'd almost say KVO isn't worth it. Could you clarify exactly how KVO is involved in this issue?

danthorpe:
`NSOperation` uses KVO to manage the state of itself and I expect to perform dependency management across queues.

We register "State" as the key path which affects "isReady" which is the key path observed by `NSOperation` for when it becomes ready to execute. When this changes (to true), `NSOperation` invokes `start()`. I guess the next thing which I can try to check is, what happens inside our `ready` override for this non-progressing operation instance. I guess ultimately, if there is a bug in `Operation` it would be there. However, I suspect more than likely this is an artifact of KVO when working with this number of objects - 5000 is quite a lot of operations to add to a queue.

@davedelong - can you offer any insight into this?

danthorpe:
I added more logging to see what is happening inside an operation which fails. After we set `state = .Ready`, it then checks `ready` and we correctly return `true` as the result of `super.ready || cancelled`. However, what should happen is then `start()`, `main()`, `finish()` - but instead nothing happens.

And just to clarify, without adding the condition, running 5000 iterations works okay. Performing the same test using `NSOperation` and `NSOperation` queue works okay up to around 7500 iterations - but beyond that things start to go off the rails, either with unfulfilled expectations or just stalling on the main thread.

So, in conclusion, I feel that this is mostly just reaching the limits to KVO, and I'm reluctant to spend any more time on it.

This issues was initially raised by @bastianschilbe, so will wait wait now to hear back from him.

jshier:
I don't have any additional suggestions, aside from using something like Hopper to debug into the system frameworks on iOS. I do want to add, though, that I think I've run into this issue, where some network operation in my app fails to finish because it gets stuck like this. 

A related question is, does this issue change in frequency depending on the iOS hardware it's being run on?

danthorpe:
Thanks for the feedback @jshier. Good questions - I've run the tests on a range of Mac hardware - haven't noticed a discernible difference, although on my 12" MacBook it's far more likely that the tests fail because the main thread stalls.

I think possible another thing to look into, is to write an example app which does something similar (but not a unit test) and then run it on some iOS hardware to see if this is an issue which only presents in the iOS simulator. I get the sense that it's less likely to occur in the Mac test target than the iOS one - but can't say so definitely.

And again, I'm a little bit reluctant to sweep the whole thing under the carpet as I'm pretty sure this issue can present itself with "normal" usage. I don't really see that running tight loops adding 5,000 or 10,000 operations to a queue as being normal.

Having said all of that.... if this is an issue with `NSOperation` or KVO, then I feel that it's largely out of my hands. I might as well begin work on v3.0 which breaks the inheritance from `NSOperation` and goes "pure Swift".

bastianschilbe:
@ansf and me are rather clueless regarding this one, too :( We've tested several scenarios and only observe missing completion callbacks when we introduced KVO with an additional state variable. Couldn't find a way to fix this so far. We totally agree that 5000+ operations is not a default use case, but can confirm that it seems to fail more often when URLOperations are involved... somehow. 
We put all our hope in the next version in "pure Swift"

ansf:
I changed the original reproducer to use a fresh OperationQueue in each iteration. This still triggers a wait timeout, often within a few hundred iterations:

```
    func test__multiple_operations_completion_blocks() {
        continueAfterFailure = false
        
        (0...500000).forEach { i in
            let q = OperationQueue()
            q.maxConcurrentOperationCount = 1
            
            let e = expectationWithDescription("wait\(i)")
            let operation = BlockOperation()
            operation.addObserver(FinishedObserver { _, _ in e.fulfill() })
            
            let condition = BlockCondition { false }
            operation.addCondition(condition)
            q.addOperation(operation)
            
            waitForExpectationsWithTimeout(10, handler: nil)
        }
    }
```

I think we should continue to investigate this issue.

As an aside, I also saw this test failing on the MacOs scheme, but it took several 100k iterations to trigger the bug.

danthorpe:
You have the `waitForExpectations` inside the loop? Is that right?

ansf:
Yes, basically I wanted to test if I can trigger the issue with one OperationQueue and one Operation.
The loop is just there to execute this test many times in sequence.

danthorpe:
@ansf @bastianschilbe @jshier okay, so definitely still looking into this, but for now, I'm going to merge this PR as there are some fixes in here which I don't want to go stale.

The original bug report #175 is still open, and can track progress, or until another PR is created.

My thinking is that other than bug fixing, I probably won't be adding any new "features" to the version 2.0 line in this project, and I'll probably start work afresh on an pure Swift implementation with `Operation` as a base class in the near future.

pronebird:
@davedelong @smileyborg I apologize for spamming, but I naively hope that maybe somebody from Apple could take a look at what's going on in NSOperation?

smileyborg:
@pronebird Please file a radar if you think there's an Apple bug. If you can also include a minimal sample project that clearly reproduces the issue, it is much more likely to get prompt attention.

I know you have good intentions, but tagging random Apple folks on GitHub isn't going to help if you haven't filed a bug report; it's a huge company and issues will only get properly screened, routed, & fixed if they are in our bug tracking system (Radar).

Thanks!

PR #170, opened 2016-02-11
CloudKitOperation should conform to ResultTypeOperation

danthorpe:
If the consumer doesn't set the completion block on a `CloudKitOperation`, then it will "work" but nothing will actually happen. Errors will not get caught, and the "result" will not get used.

So, perhaps there should be a default completion block which allows the operation to be used as a result.

i.e.

```swift
public protocol CKOperationType: class {
    // etc

    typealias OperationResult
}

public class CloudKitRecovery<T where T: NSOperation, T: CKOperationType>: ResultTypeOperation {
    public typealias Result = T.OperationResult

    public var result: Result? {
        return getResultBlock()
    }
    internal var getResultBlock: (() -> Result?)? = .None
}
```

This would allow each `CKOperation` type to define its own "Result" and set a block which returns the result inside a default completion handler.

### DIFF
--- a/Sources/Features/Shared/CloudKitOperation.swift
+++ b/Sources/Features/Shared/CloudKitOperation.swift
@@ -512,8 +512,12 @@ extension CKQueryOperation: CKQueryOperationType {
 
 // MARK: OPRCKOperation
 
-public class OPRCKOperation<T where T: NSOperation, T: CKOperationType>: ReachableOperation<T> {
-
+public class OPRCKOperation<T where T: NSOperation, T: CKOperationType>: ReachableOperation<T>, ResultOperationType {
+    
+    public typealias Result = T.Result?
+    
+    public var result: Result = nil
+    
     convenience init(operation op: T) {
         self.init(operation: op, connectivity: .AnyConnectionKind, reachability: ReachabilityManager(DeviceReachability()))
     }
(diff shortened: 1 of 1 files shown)


danthorpe:
Okay, so I don't think this is going to realistically be possible. Here's why:

1. Defining the `Result` typealias of each operation is fairly trivial.

2. Some operations have a completion block, but also have blocks which receive individual values/instances from an array of results. For example `CKFetchRecordChangesOperation` has a `recordChangedBlock` in addition to a completion block. To be useful, `CloudKitOperation` would need to collect all of these intermediary results and include them in a final "result". Less trivial, but in theory do-able.

3. It starts to get tricky when we consider how we should setup this default behavior. `OPRCKOperation` would need to effective set these completion handlers when it is initialized. This means that the protocol `CKOperation` will been to have a definition of something like:

```swift
func setupDefaultResultHandler() -> () -> Result
```

or something? In other words, in each `CKOperation` we implement a function, which will collect the final result type. This will involve setting intermediate result blocks and the completion block. We need to have a function which returns a completion block which does this, as we cannot add any properties to the `CKOperation` subclasses (the  classes).

4. So, it's already getting a bit complicated - then we need to consider, that we need to support this within the context of `RetryOperation` where the instance of the `CKOperation` is returned as needed (at execution time) from a closure. Therefore, we need to apply the default result collection behavior at this time.

5. And lastly, we need to make all of this work with the addition of allowing the user to set their own completion blocks & intermediate results.

So... gonna drop it for now. Increases the complexity of the class unreasonably.

Issue #174, opened 2016-02-03
GroupOperation with AlertOperation not finishing


bastianschilbe:
If I add an AlertOperation to a GroupOperation the finish observer of the AlertOperation gets called but the outer GroupOperation never finishes. It seems that the containing UIOperation contained in the AlertOperation is scheduled on the GroupOperation queue but is never finished so the group cannot finish. Do I miss something here?
```
let group = GroupOperation()
group.addOperation(AlertOperation(presentAlertFrom: self))
group.addObserver(FinishedObserver(didFinish: { _, _ in print("expected") }))
queue.addOperation(group)
```

danthorpe:
Hi @bastianschilbe - this in interesting, could be a bug. 

I'm assuming that you're dismissing the `UIAlertController` that is presented? If so, I think your expectations are justified.

I'll take a look into it tonight, hopefully get back to you.

danthorpe:
Hi @bastianschilbe, can you check this again now on the latest `development`? Thanks!

malhal:
The Alert Operation was designed by Dave Delong who is not a UIKit developer and instead of solving his problems using UI methods he has opted to make a super complicated wrapper via his NSOperation framework. This library should come with a warning that it shouldn't be used by programmers already proficient in UIKit and all beginners. In his talk on this at WWDC I remember him saying there are issues with alerts that they can appear on top of each other (false) and in fact with the new alert controller it is impossible to even present two alerts because the second gets dropped. Given his background in operations he has attempted to solve this using that but in fact the solution is to fix the alert controller to allow sequential alerts, how the original UIAlertView worked, if you know UIKit you know how to solve this problem.

If you run the eaethqaukes sample you'll experience all kinds of UI bugs, including if you tap an earthquake then tap more info, dismiss the safari controller then try to tap more info again you will find it won't work and the Operation queue is blocked. This is because he didn't know that the safari controller dismisses itself when the done button is pressed, thus the dismiss with completion has no effect and the completion does not fire thus the operation never ends. its crazy to me there is all this bloat and convoluted code and he claims it's so powerful yet it falls over in the simplest of cases. This I think the library should come with a severe health warning.

malhal:
And by the way same thing with CloudKit, it already has an internal queue that runs sequentially so there is no reason to wrap it up in more operations.

And same thing with NSURLSession.

jshier:
[...]

PR #185, opened 2016-02-11
Add UIOperation finishing to AlertOperation action callback

bastianschilbe:
The UIOperation inside the AlertOperation doesn't finish. This causes two bugs: 1) The group operation which contains an AlertOperation is never finished 2) The UIOperation never releases the PresentingViewController

### DIFF
--- a/Sources/Core/Shared/Operation.swift
+++ b/Sources/Core/Shared/Operation.swift
@@ -478,6 +478,7 @@ public class Operation: NSOperation {
     - parameter operation: a `NSOperation` instance.
     */
     public final func produceOperation(operation: NSOperation) {
+        precondition(state > .Initialized, "Cannot produce operation while not being scheduled on a queue.")
         log.verbose("Did produce \(operation.operationName)")
         didProduceOperationObservers.forEach { $0.operation(self, didProduceOperation: operation) }
     }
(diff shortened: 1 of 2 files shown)


bastianschilbe:
This was reportet in issue #174

danthorpe:
Arr! Awesome is that all that is needed? Tests are probably failing because `danthorpe:development` has updated the Gemfile / environment dependencies

bastianschilbe:
UIOperation seems to be designed to be finished from outside. So someone should call finish on it. Hope I understand this operation right. For me it looks like this is all to fix this issue.

danthorpe:
Oh, I see. You are certainly correct that `UIOperation` is not finishing, but I think that is how it should be fixed. Not really sure how this got past me - I'm fixing it now in #186.

danthorpe:
Really appreciate the input on this issue, but I think the fix needed to be made in `UIOperation` which I've done in #186. Will keep this here for now, but if #174 is now fixed, I think this can be closed.

bastianschilbe:
I also thought about fixing UIOperation the way you did but in case use don't present the view controller the completion callback is also never called (ViewControllerDisplaStyle.displayController kind of lies about it's completion parameter).

danthorpe:
Arr, yes, you are right. Okay, I'll sort this out.

danthorpe:
@bastianschilbe I'm going to close this now. Thanks for the input to the project!

PR #187, opened 2016-02-12
Adding precondition to produceOperation method

bastianschilbe:
Calling produceOperation without being on a OperationQueue will silently fail. If you try to use produceOperation in an operation initializer this situation might happen.


danthorpe:
Oh, this is a good catch!

danthorpe:
Thanks @bastianschilbe!

PR #188, opened 2016-02-14
UIOperation does not always finish

danthorpe:
Related to #174 & #185.

### DIFF
--- a/Examples/Permissions/Pods/SwiftyBeaver/README.md
+++ b/Examples/Permissions/Pods/SwiftyBeaver/README.md
@@ -1,35 +1,36 @@
 # SwiftyBeaver
 
-<img src="https://img.shields.io/badge/Platform-iOS%208%2B-blue.svg" alt="Platform iOS8+">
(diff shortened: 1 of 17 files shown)


danthorpe:
Think that this is properly fixed now @bastianschilbe. Inside `ViewControllerDisplayStyle`'s `displayController` method, we just ensure that the completion block is called.

Tested this using an `AlertOperation` in the Permissions example app. The table view gets its datasource set inside a `BlockOperation` which is dependent on an `AlertOperation` finishing. So, the app launches, shows an alert, and when you tap `Okay`, the alert dismisses and the table view has rows.

This should fix #174, although I've not specifically set up a test for this yet.

danthorpe:
@bastianschilbe - actually just re-wrote my changes to the Permissions project pretty much as you have in the bug report for #174:

```swift
let config = configure()
let group = GroupOperation(operations: createInfoAlert())
config.addDependency(group)

queue.addOperations(config, group)
```

And this works, meaning that the group is correctly finished.

PR #179, opened 2016-02-14
Reachable Operation - still has a bug

danthorpe:
There is still a bug in `ReachableOperation`. It works correctly when the device is already online, however when offline, the operation does not execute when the device comes back online.

The most recent change here remove the private singleton which was holding all the "observers" instead making them all individual. But this is not going to work.

So, back to the drawing board and need to re-consider how this can be unit tested. Possibly move away from the function registered in the run loop - as it's hard to make this non-global.

### DIFF
--- a/.ci/buildkite/pipeline.template.yml
+++ b/.ci/buildkite/pipeline.template.yml
@@ -2,13 +2,9 @@ steps:
   -
     name: ":fastlane: Build iOS Extension"
     command: .ci/scripts/build-ios-extension
-    # agents:
-    #   name: "$BUILDKITE_AGENT_META_DATA_NAME"
   -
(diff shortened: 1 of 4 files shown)


danthorpe:
This can be verified in a real app using code like this:

```swift
func doReachableOperationStuff() {
    (0..<5).forEach { i in
        queue.addOperation(createReachableOperationWithInfo("\(i): is Reachable!"))
    }
}

func createReachableOperationWithInfo(info: String) -> Operation {
    let operation = ReachableOperation(BlockOperation(block: { (continueWithError: BlockOperation.ContinuationBlockType) in
        print(info)
        continueWithError(error: nil)
    }))
    operation.addObserver(BackgroundObserver())
    return operation
}
```

PR #192, opened 2016-02-17
Make podspec dependencies more granular

itsthejb:
Hi there,

Great lib! We're just starting using this, and noticed that in spite of the current use of subspecs, it would be nice to make them just a shade more granular. In particular, we'd rather not CloudKit unnecessarily, which is currently required in the Standard subspec.

In any case, please let me know what you think.

Thanks!

### DIFF
--- a/Operations.podspec
+++ b/Operations.podspec
@@ -32,6 +32,12 @@ session Advanced NSOperations: https://developer.apple.com/videos/wwdc/2015/?id=
       'Sources/Features/Shared',
       'Sources/Features/iOS'
     ]
+    ss.ios.exclude_files = [
+      'Sources/Features/Shared/CloudCapability.swift',
+      'Sources/Features/Shared/CloudKitOperation.swift',			
+      'Sources/Features/iOS/PhotosCapability.swift',
+      'Sources/Features/iOS/PassbookCapability.swift'
+    ]
     ss.watchos.exclude_files = [
       'Sources/Core/iOS',
       'Sources/Features/iOS/LocationCapability.swift',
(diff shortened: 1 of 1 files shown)


danthorpe:
Hi! @itsthejb thanks a lot for the input! I totally agree with you on this, and that the subspecs can be a little bit more granular.

There is an existing issue #112 which is sort of related. Part of the issue is that we have stuff which is not API Extension compatible. And the goal for v2.7 is to combine all the code into one workspace/project where the core `Operations` module is Extension compatible, and have everything else broken up into modules. I think the CocoaPod specs should then mirror this.

Happy to merge this into `development` now, if you are already on the "bleeding edge". However, I would like to sort out all the project configurations as part of 2.7 to avoid so many continual breaking changes. If you can manage for now with v2.6 - then it might just be easier to wait until the changes for #112 are proposed and see what you think?

Operations.podspec
```diff
@@ -121,7 +129,47 @@ session Advanced NSOperations: https://developer.apple.com/videos/wwdc/2015/?id=
...
+  # Subspec which includes CloudKit functionality
+  s.subspec '+CloudKit' do |ss|
+    ss.platforms = { :ios => "8.0", :osx => "10.10" }
+    ss.dependency 'Operations/Standard'
+    ss.frameworks = 'CloudKit'
+    ss.source_files = [
+      'Sources/Features/Shared/CloudCapability.swift',
+      'Sources/Features/Shared/CloudKitOperation.swift',			
+    ]
+  end
```
danthorpe: This is a duplication in error I think which should be removed. 

Also CloudKit is available on tvOS platform, which will need to be added here, and updated in the excludes from Standard spec.

itsthejb:
Hey there,

Well we've forked for the time being, so no particular urgency at our end. Whatever you think best in the long run.

itsthejb:
Hi there,

Had a chance to make those changes you recommended. Please let me know what you think. I'm afraid a bit too snowed under to look into the changes for the upcoming release, though, so sorry if there's any redundancy.

danthorpe:
@itsthejb thanks for this - sorry it took a little while to merge. Preparing v2.7 now which will include these changes. Thanks!

Issue #152, opened 2015-12-29
Operation Observers in 2.5


jshier:
In 2.5 it was made a precondition failure to try to add an observer to an operation that had already progressed to the `.Ready` state. To me, this seems to greatly limit utility of observers, in some situations anyway. For instance, I can no longer decouple observation with operation creation. i.e. I can do this anymore:
```swift
var currentOperation: Operation? {
    didSet {
        guard let operation = currentOperation else { return }
        operation.addObserver(...)    
   }
}
...
func doSomething() -> Void {
    currentOperation = OperationManager.doSomething {
        print("did something")
    }
}
```
Instead, if I wish to add an observer, I must do it at operation creation, forcing me to tightly couple my observers and operations. This limits the scenarios in which I can use observers and the functionality of the observers to that which can be tightly coupled.

I'm not too familiar with the underly observer code, but perhaps it would be possible to, instead of checking to make sure the operation isn't yet ready before adding an observer, "play back" the state transitions the observer must logically have missed right after adding the observer to the operation. I realize this might not be possible for something like the finished or produced observers, depending on the states, but should be possible for general start observers and perhaps others.

danthorpe:
Hi @jshier, hmm, but doesn't the use of `precondition` make explicit what was already the expectation of `Operation`?

If an observer is added after the operation moves past the `.Ready` state, then there is the potential that they will be added after the relevant event has passed. I will admit however, that given that now `OperationObserver` is somewhere redundant, that this only applies to some of the observer types, and that it could be improved, i.e. it should be possible to add an `OperationDidFinishObserverType` right up until it goes to `.Finishing` state.

Can you explain your usage a little bit more? What sort of observers are you adding, and how does the `OperationManager` work with regards to adding these operations to a queue versus adding the observer?

jshier:
`OperationManager`s of the type I've been writing usually work like this:
```swift
class OperationManager : OperationQueue {
    private static let sharedManager = OperationManager()

    static func doSomething(completionHandler: () -> ()) -> SomeOperation {
        let operation = SomeOperation(completionHandler: completionHandler)
        sharedManager.addOperation(operation)

        return operation
    }
}
```
[...]

PR #193, opened 2016-02-18
Refactor Observers

danthorpe:
This is to address the points raised in #152 by @jshier. Plus a few things I've noticed subsequently while fixing bugs for v2.6.

- [x] Supports Will/Did Finish
        The reasoning behind this, is that typically, observers want to know when the operation did *actually* finish. This means that `XCTAssertTrue(operation.finished)` should pass. However... `GroupOperation` and its like, will often want to use this observer (or a similar event) to add new operations to the queue. Hence, it used to be posted before the state change. Now, we just have two observers, one before and one after.

- [x] Renames methods to a better Swift style
        For example, instead of `operationDidStart` it is `didStartOperation`. Changed consistently throughout including in `GroupOperation`, which now provides `willFinishOperation` instead of `operationDidFinish` - note that this takes the change above into account too.       

- [x] Adds and calls a method for did attach to operation.
        The base protocol defines `didAttachToOperation` and in an extension, we have a default implementation which currently just logs a `.Verbose` message. In a custom observer, this can be over-ridden as you like.

- [x] Make thread safe & remove asserts from `Operation` when adding an observer.

### DIFF
--- a/Sources/Core/Shared/BlockObserver.swift
+++ b/Sources/Core/Shared/BlockObserver.swift
@@ -88,7 +88,7 @@ public struct ProducedOperationObserver: OperationDidProduceOperationObserver {
 
 /**
  WillFinishObserver is an observer which will execute a
- closure when the operation finishes.
+ closure when the operation is about to finish.
  */
 public struct WillFinishObserver: OperationWillFinishObserver {
     public typealias BlockType = (Operation, [ErrorType]) -> Void
@@ -114,7 +114,7 @@ public struct WillFinishObserver: OperationWillFinishObserver {
 
 /**
  DidFinishObserver is an observer which will execute a
- closure when the operation finishes.
+ closure when the operation did just finish.
  */
 public struct DidFinishObserver: OperationDidFinishObserver {
     public typealias BlockType = (Operation, [ErrorType]) -> Void
--- a/Sources/Core/Shared/BlockObserver.swift
+++ b/Sources/Core/Shared/BlockObserver.swift
@@ -87,7 +87,7 @@ public struct ProducedOperationObserver: OperationDidProduceOperationObserver {
 
 
 /**
- FinishedObserver is an observer which will execute a
+ WillFinishObserver is an observer which will execute a
  closure when the operation finishes.
  */
 public struct WillFinishObserver: OperationWillFinishObserver {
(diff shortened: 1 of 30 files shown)


danthorpe:
@jshier - I've got some tests to write, but I think the changes here should make the use of Operations a lot more flexible. I'll probably merge this over the weekend, but gonna be largely directed by your feedback on this - so if you could let me know what you think, and if there is anything more to do, that would be great!

danthorpe:
Okay, I'm going to merge this now, as it has quite a few breaking changes over 2.6, and so is going to be hard to stop getting out of sync if I leave it.

jshier:
I'll give this a try right now.

jshier:
Looks great so far. I'll take a closer look at the changes but my initial implementation of an `UIActivityIndicatorView` extension to observe an `Operation` works well. I'll work up to some more advanced implementations perhaps tomorrow. 

A few small notes: 
1. I was disappointed that the default `Block` observers didn't offer the ability to use `didAttach`, but I can see why keeping them simple is a priority. I was able to create my own block-based observers using it just find.
2. The `BlockType`s used by the `BlockObserver`s could use parameter labels, so when I autocomplete them I don't have to fill them in myself. Also, you might be able to define some of them in the protocols (I'm not sure), so other implementers have access to the definitions and labels.

I'll let you know if I have more feedback later.

danthorpe:
Hi @jshier - thanks for trying this out. Did you have to re-write much code? I added `@available(*, unavailable, renamed="etc etc")` annotations where possible - so hopefully Xcode offered to fix things for you.

Perhaps rather than an extension on `UIActivityIndicatorView` a bespoke observer which receives one would be useful to add to the project?

Regarding your notes:

1. Yeah, I can fix this, although I'm thinking that instead of accepting more blocks in an initializer, we can have mutable properties `var didAttachToOperation: (Operation -> Void)? = .None` *and* `var willDetachFromOperation: (Operation -> Void)? = .None` where the latter property would be called after the observer's last event is performed.

    For example - lets say you attach an `OperationDidStartObserver` - after the operation starts, if this observer does not conform to any of the other observation protocols, then it can be deallocated - but we can inform it before that happens. This observer could then be used to potentially attach different observers to the operation.

2. Yeah, this is no problem to fix either.

I will do another ticket later today.

danthorpe:
See #195 for pt. 2

PR #195, opened 2016-02-21
Refactor Observers, part 2

danthorpe:
- [x] Add property for `didAttach` to all the block based observers.
- [x] Add labels to the closures used in the block based observers.
- [x] Added `events` option set property to Observer protocols. This isn't actually used yet, but figured it is actually quite handy to be able to inspect which events an observer is interested in without so much optional casting.

~~- [ ] Add `willDetachFromOperation` to the base protocol~~ - This is actually a little bit harder to figure out. Probably not worth the extra blocking calls to remove the observers. Especially as we'd have to figure out which ones are safe to remove.

~~- [ ] Prune observers from the `Operation` after their last event has been observed.~~

~~- [ ] Add property for `willDetach` to all the block based observers.~~

### DIFF
--- a/Sources/Core/Shared/BlockObserver.swift
+++ b/Sources/Core/Shared/BlockObserver.swift
@@ -195,8 +195,14 @@ public struct BlockObserver: OperationObserver {
     let willFinish: WillFinishObserver?
     let didFinish: DidFinishObserver?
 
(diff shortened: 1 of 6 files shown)


danthorpe:
@jshier - some more changes to improve after your initial feedback.

Issue #197, opened 2016-02-22
Errors in nested group operations don't aggregate up


bastianschilbe:
The error produced by an operation inside a group operation which is executed in another group operation isn't aggregated to the outer group operation. There is also a comment in code which says that it is explicitly avoided to add the errors to prevent duplicating all errors. But shouldn't this test run:

        let failingOperation = BlockOperation { (continueWithError) -> Void in
            continueWithError(error: NSError(domain: "", code: 1, userInfo: nil))
        }
        let innerGroup = GroupOperation(operations: [failingOperation])
        let outerGroup = GroupOperation(operations: [innerGroup])
        let queue = OperationQueue()
        
        let expectation = expectationWithDescription("Check Errors")
        
        outerGroup.addObserver(FinishedObserver(didFinish: { (op, errors) -> Void in
            XCTAssertTrue(!errors.isEmpty)
            expectation.fulfill()
        }))
        
        queue.addOperation(outerGroup)
        
        waitForExpectationsWithTimeout(1, handler: nil)

danthorpe:
@bastianschilbe - yeah, I think you might have discovered a regression here! I'll have a look into it. When I added that comment: the `GroupOperation` was automatically getting the errors from any nested `GroupOperation`, but maybe this has stopped happening.

But yeah - I will look into this - good shout!

danthorpe:
Hi @bastianschilbe - thanks for this bug report. You were totally correct! 👍

Anyway, I'm pretty sure that I've fixed it now, merged into `development`. If you have a chance to work on the dev that would be great. Let me know how you get on, and hopefully this can be closed 😀.

FYI - There are some breaking changes on `development` related to operation observers. Hopefully not too serious for you though.

danthorpe:
I'm doing some clean up - pretty sure that this is fixed. @bastianschilbe please re-open if not :)

PR #198, opened 2016-02-22
Branch investigating errors within nested GroupOperation instances.

danthorpe:
To address #197.

- [x] Write a test which exercises the bug/exercises the mistake.
- [x] Determine wether or not this is a regression.
- [x] Fix bug.

### DIFF
--- a/Sources/Core/Shared/ComposedOperation.swift
+++ b/Sources/Core/Shared/ComposedOperation.swift
@@ -18,8 +18,8 @@ public class ComposedOperation<T: NSOperation>: Operation, OperationDidFinishObs
     }
 
     init(operation op: T) {
-        self.target = op as? Operation ?? GroupOperation(operations: [op])
-        self.operation = op
+        target = op as? Operation ?? GroupOperation(operations: [op])
+        operation = op
         super.init()
         name = "Composed Operation"
         target.name = "Composed <\(T.self)>"
@@ -27,6 +27,7 @@ public class ComposedOperation<T: NSOperation>: Operation, OperationDidFinishObs
     }
 
     public override func cancel() {
+        target.cancel()
         operation.cancel()
         super.cancel()
     }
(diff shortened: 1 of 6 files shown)


danthorpe:
Copied from the commit message:

>  This is a little bit complicated, so I'm going to explain what is going on here.
> 
> 1. CloudKitOperation is a GroupOperation subclass.
> 2. The child operations in this group are OPRCKOperation instances, which are themselves subclasses of ComposedOperation.
> 3. The composed operation is Apple's CKOperation
> 4. When composing NSOperation (not Operation) subclasses, they get put into a GroupOperation.
> 5. When ComposedOperation, which composes an NSOperation executes, it "produces" the GroupOperation.
> 6. When producing an operation, it gets added to the receiver's queue.
> 
> Therefore, all of this together means we have... A group operation, which has children which produces another GroupOperation. Therefore, the outer GroupOperation (the CloudKitOperation<T>) gets notified of when the produced operation (the GroupOperation which is wrapping T) finishes, but also, when it's own child (the OPRCKOperation / ComposedOperation) finishes. And, ComposedOperation also collects the errors of its composed operation.
> 
> This means that by default, GroupOperation will collect the error twice.
> 
> In summary, when GroupOperations, contain ComposedOperations which produce GroupOperations, we ultimately have two observers of when that GroupOperation finishes, which means we collect errors twice.
> 
> To get round this, I have modified GroupOperation to make it cleaner - we call a method which indicates the child did finish either errors. All it does is add the errors to the aggregate errors.
> 
> In CloudKitOperation<T> we override this, and call super, but only if the child is not a OPRCKOperation. Essenitally we ignore any errors coming from the composed operation, as we'll have already collected the same error.

danthorpe:
Unfortunately it seems as though some other tests are now failing. So, will continue to look into this.

danthorpe:
FYI - still looking into this. Just got side-tracked.

Issue #199, opened 2016-02-24
Subclassing RepeatedOperation is very painful


jshier:
Due Swift's subclass initializer requirements (a subclass' designated initializer must call the superclass' designated initializer, not a convenience initializer, and a subclass must have a designated initializer) creating a subclass of `RepeatedOperation` is quite painful, as the required convenience generators and such are not public. In order to create my initializer, as such:
```swift
init(operationGenerator: () -> T?) {
    let generator = RepeatingGenerator(anyGenerator(operationGenerator))
    let strategy = WaitStrategy.Fixed(5)
    let delay = MapGenerator(strategy.generator()) { Delay.By($0) }
    let tuple = TupleGenerator(primary: generator, secondary: delay)
    
    super.init(maxCount: nil, generator: anyGenerator(tuple))
}
```
I had to make public the types, initializer, and `next()` of `RepeatingGenerator`, `WaitStrategy`'s `generator()`, `MapGenerator`, and `TupleGenerator`.

Obviously this isn't a viable solution. Could the library be updated to make the relevant types and properties public? Or is there some other way I'm supposed to be subclassing `RepeatedOperation`?

danthorpe:
Hi @jshier - this is great feedback - I've not tried subclassing this outside of the module. It can definitely be improved. :)

I think if I make the convenience initializers in `RepeatedOperation` all designated initializers (effectively), which will make the library code a little bit less neat, then it'll be much easier for consumers to subclass it. I was also thinking of making the initializer available in the extension (for `T: Repeatable`) available by default.

Also, I am a little bit torn between making some of these internal types public. E.g. `ReachabilityManager`, `MapGenerator`, `FiniteGenerator` etc. On the plus side - they're very helpful, on the con side, it means supporting them (which is restrictive when it comes to refactoring/rewriting). What do you think?

jshier:
I think if the initializers allow the flexibility the convenience initializers expose, then exposing the underlying types would be unnecessary. However, I can see how opening up the initializers like that would lead to some rather messy and duplicate code. I would lean towards the initializer path since once those are open the need for external developers to use the other types is essentially eliminated, unless they're creating other types of generators.

danthorpe:
Yep - I agree. Take the hit on `RepeatedOperation` becoming a little messy by "opening up" (i.e. remove `convenience` and *just making it work*), instead of making public all the helper generators and other things.

Can always make public later and refactor the internals of `RepeatedOperation` later.

[...]

PR #203, opened 2016-02-25
Refactor RepeatedOperation initializers

danthorpe:
At the moment, it's very tricky to subclass `RepeatedOperation` from outside the module.

- [x] Remove `convenience` initializers - as these cannot be used by subclasses.
- [x] Try to minimize the messy initialization code.
- [x] Unify / Cleanup / ~~Remove~~ the `Repeatable` & `RepeatedOperation` stuff - I think this has got a bit confused since it was first written.

This is to address issues raised in #199.

### DIFF
--- a/Sources/Core/Shared/RepeatedOperation.swift
+++ b/Sources/Core/Shared/RepeatedOperation.swift
@@ -495,17 +495,17 @@ public protocol Repeatable {
     func shouldRepeat(count: Int) -> Bool
 }
 
-class RepeatingGenerator<G: GeneratorType where G.Element: Repeatable>: GeneratorType {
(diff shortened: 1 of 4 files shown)


danthorpe:
@jshier - I've made a start on these changes. Currently everything is passing tests (CI is currently failing only because coverage has dropped).

Let me know what you think - hopefully should be much easier to subclass these.

danthorpe:
This is all looking good now. I renamed `RepeatedGenerator` to `RepeatableGenerator` and made it public. This is because I cannot get around having the initializer for when `T: Repeatable` a designated initializer, because it's in a constrained extension.

Issue #205, opened 2016-03-02
Sources/Features/Shared/CloudCapability.swift should be included in tvOS PodSpec


tbaggett:
tvOS doesn't currently support retrieving a user's contacts information via discoverAllContactUserInfosWithCompletionHandler. However, it does support checking the user's iCloud account status with [accountStatusWithCompletionHandler](https://developer.apple.com/library/prerelease/tvos/documentation/CloudKit/Reference/CKContainer_class/index.html#//apple_ref/occ/instm/CKContainer/accountStatusWithCompletionHandler:). 

The CloudCapability.swift file is currently excluded from the tvOS target in the PodSpec file. It should be included for tvOS as well, possibly with some "#if os(tvOS) checks" to exclude the permissions portion of the code.

danthorpe:
Arr - good spot, thanks @tbaggett. Will look into this today as a 2.6.0 bug, and release it as 2.6.1.

danthorpe:
@tbaggett this is now fixed on this branch: `hotfix/OPR-205_tvOS_cloud_capability`. If you can try it, that would be great. Change your Podfile so that you have:

```ruby
pod 'Operations', :git => 'https://github.com/danthorpe/Operations', :branch => 'hotfix/OPR-205_tvOS_cloud_capability'
```

I did create a sample app to check that it all compiles, but I've not got entitlements set up, so can't *really* test it right now. Hopefully should be easier for you. 

Let me know how it goes, and I'll create 2.6.1 release.

tbaggett:
Awesome, thanks for the quick response! I will test this today and get back to you with the results.

tbaggett:
Its working great, no problem. I set breakpoints in the CloudCapability verifyAccountStatus method and confirmed that it was being called and that the "Available" CKAccountStatus was returned when I passed in my custom container name.

danthorpe:
Great. It's about midnight here in London. Will merge and do v2.6.1 tomorrow morning (in about 8 hours). 

Thanks for the feedback!

danthorpe:
Okay, this has been merged as a hot fix. Will be in `2.6.1`, `master` and `development` later today.

PR #206, opened 2016-03-02
CloudCapability should be available in tvOS

danthorpe:
This is to address #205.


danthorpe:
Assuming that the application has the required CloudKit entitlements, this now is possible in tvOS.

```swift
let operation = BlockOperation { print("Logged into CloudKit") }
operation.addCondition(AuthorizedFor(Capability.Cloud()))
queue.addOperation(operation)
```

The above will check that that the user is signed into the default container with default permissions. To changes the defaults, pass arguments into `Capability.Cloud()`.

PR #208, opened 2016-03-02
Default argument for designated initializer due to internal type

itsthejb:
Hi there,

Thanks again for the great work on this library - it continues to impress us with its capabilities.

Minor one here, I think: we want to implement a throttling condition which will fail if we exceed a certain number of network operations *on cell connection* in a particular period. Therefore the logic (seemingly) is:

* `ReachabilityCondition` with `Connectivity.ViaWiFi`
* If this condition fails, we can assume we're on cell connection.
* Inspect the number of requests from a particular period to see if we've hit our threshold, and succeed/fail based on that.

This sort of condition composition doesn't appear to be supported (and would be great condition!), but for this simple usage, I can subclass `ReachabilityCondition`, and override `evaluateForOperation()`. Not super clean, but looks to be enough. Currently subclassing is an issue, though, because the designated initializer uses the internal type `HostReachabilityType`. Since we don't care about this argument, a better solution seems to be to drop the convenience initializer and just have a default argument here. Since the type is internal it can't be used.

In any case, if there's a better solution for the problem we're solving here, also very keen to hear about it.

Thanks!

### DIFF
--- a/Sources/Features/Shared/ReachabilityCondition.swift
+++ b/Sources/Features/Shared/ReachabilityCondition.swift
@@ -27,11 +27,7 @@ public class ReachabilityCondition: OperationCondition {
     let connectivity: Reachability.Connectivity
     let reachability: HostReachabilityType
 
-    public convenience init(url: NSURL, connectivity: Reachability.Connectivity = .AnyConnectionKind) {
-        self.init(url: url, connectivity: connectivity, reachability: ReachabilityManager(DeviceReachability()))
-    }
-
-    init(url: NSURL, connectivity: Reachability.Connectivity = .AnyConnectionKind, reachability: HostReachabilityType) {
+    public init(url: NSURL, connectivity: Reachability.Connectivity = .AnyConnectionKind, reachability: HostReachabilityType = ReachabilityManager(DeviceReachability())) {
         self.url = url
         self.connectivity = connectivity
         self.reachability = reachability


itsthejb:
Hmmm... Actually, no this doesn't work since can't have a public initializer that uses an internal type at all :grinning: 

I suppose no big deal - will just make our initializer `convenience`. It's an internal type, so doesn't matter much. Please let us know if you have any better suggestions, though. Thanks!

jshier:
Putting this here since this is where you explained your use case.

Just wanted to point out that you may not want to use (or be able to reliably use) reachability for this. Not only does it fly in the face of Apple's recommended best practices, namely not using reachability as a precondition for network requests, but there is no guarantee that your logic (if reachability reports only Wifi, you're not on cellular) will hold all (or really any) of the time. There's no inherent connection between the interfaces reachability reports at any given moment for a particular domain and what your requests are going over, so using it to build some sort of throttling is probably a bad idea. Apple's documentation on this is terrible, but I recommend watching Apple's [WWDC 2012 Session 706, "Networking Best Practices"](https://developer.apple.com/videos/play/wwdc2012-706/) video. At best I think your intended use would work some of the time, but failures (inaccurate counts of requests over cellular) would likely be common and invisible, since you haven't nothing to check it against.

Just wanted to point this out.

itsthejb:
Great point! I had forgotten about that :) ok, well then we'll have to see how it goes, and most likely do a million iterations!

Thanks!

danthorpe:
This is a really good point - thanks @jshier. I've been aware that `ReachabilityCondition` does sort of fly in the face of Apple's recommendation.

And at the moment, `ReachabilityOperation` is just a `ComposedOperation` subclass - where it only run the operation "when connected". However, it's on my list of things to do that this can be refactored to be a `RetryOperation` subclass with a `TimeoutObserver`, and is one of the many reasons why I'm hesitant to make any of the Reachability stuff public at the moment.

But, yeah, not sure what to do about `ReachabilityCondition`. Perhaps we should update the docs on this type to warn people of Apple's recommended practices? If you want to use it - go ahead.

itsthejb:
Well, considering our particular needs, not so much of a concern. We just want to have a very conservative mode for WWAN, and it doesn't need to be particular accurate. It will probably suffice for now, although the proper retry mode @danthorpe mentions sounds great!

PR #209, opened 2016-03-02
Move HostReachabilityType dependency to a var, to allow for subclassing

itsthejb:
Originally https://github.com/danthorpe/Operations/pull/208

Sorry for the noise. Turns out we *do* still need to do something, since we need to store properties. Another solution, which I suspect you won't like!, is to make the `reachability` property into an (internal) var, defaulting to `ReachabilityManager(DeviceReachability())`, and remove it from the initializer entirely. This is still testable, by setting the var.

Not super elegant, but I think this solves our immediate need. Please let me know what you think.

Thanks!

### DIFF
--- a/Sources/Features/Shared/ReachabilityCondition.swift
+++ b/Sources/Features/Shared/ReachabilityCondition.swift
@@ -25,16 +25,11 @@ public class ReachabilityCondition: OperationCondition {
 
     let url: NSURL
     let connectivity: Reachability.Connectivity
-    let reachability: HostReachabilityType
+    var reachability: HostReachabilityType = ReachabilityManager(DeviceReachability())
 
-    public convenience init(url: NSURL, connectivity: Reachability.Connectivity = .AnyConnectionKind) {
-        self.init(url: url, connectivity: connectivity, reachability: ReachabilityManager(DeviceReachability()))
-    }
-
-    init(url: NSURL, connectivity: Reachability.Connectivity = .AnyConnectionKind, reachability: HostReachabilityType) {
+    public init(url: NSURL, connectivity: Reachability.Connectivity = .AnyConnectionKind) {
         self.url = url
         self.connectivity = connectivity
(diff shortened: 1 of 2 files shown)


danthorpe:
Hi @itsthejb - I'm just catching up on this - looks really interesting, this is actually the sort of sophisticated stuff that I have hoped this framework would enable!

Okay, basically, the issue here is that I *really* want to keep these "reachability" types internal at the moment, and I need to be able to set them on things to make them testable.

There are a bunch of types which use these types, `ReachabilityCondition`, `ReachabilityOperation` and `CloudKitOperation`. I know that there is an issue with using the designated initialiser to accomplish these things. So... I guess the ultimate goals is...

1. Keep some of these reachability types internal.
2. Fully support subclassing & configuration by consumers
3. Unify the injection process for setting & getting the reachability types across the framework.

So... I actually think your solution here is pretty pragmatic and ticks most of the boxes. But what I'd like to do is apply it consistently across the framework.

So... if at the moment you're okay to work off your fork & branch, then you're not blocked by this - which is great for me. I can go through your changes, which I'll merge and then quickly update all the other types to keep consistency. I'm super busy this week unfortunately so probably won't get done until the weekend.

Let me know how you get on with your throttling stuff. It sounds really interesting. Also, not sure if this will help, but `NegatedCondition` might help a little bit - it'll reverse the condition for you? Wondering if that will help with detecting WWAN?

Also - you've certainly raised an interesting idea of some like of `CompoundCondition` which has got me thinking!

itsthejb:
Hi!

Yes, no problem working off our fork. Since we're working pretty fast over here, main priority is any (minimal!) changes necessary to serve any urgent needs. Tidy up can come later!

danthorpe:
Thanks for this @itsthejb  - sorry it took so long to merge!

itsthejb:
:+1:

PR #210, opened 2016-03-02
Reachabiility should pass when constrained to wifi only, and wifi is available

itsthejb:
Hi, again!

So another change related to the feature mentioned in https://github.com/danthorpe/Operations/pull/209: As stated there, we implement a `ReachabilityCondition` for WiFi only. We then test for throttling, and possibly have another condition for WWAN only.

In any case, I noticed that the pattern matching doesn't appear to accept the case where we want Wifi only, and only wifi is available. I think this is based on the idea that in the real world, we're more likely to have WWAN, although WiFi is higher priority. In any case, added this as a succeeding condition, including a test.

Not sure this is the most elegant matching, either: I don't claim to be the greatest Swift-er (yet!).

Please let me know what  you think.

Thanks!

### DIFF
--- a/Tests/Features/ReachabilityConditionTests.swift
+++ b/Tests/Features/ReachabilityConditionTests.swift
@@ -98,6 +98,18 @@ class ReachabilityConditionTests: OperationTests {
 
         XCTAssertEqual(error, ReachabilityCondition.Error.NotReachableWithConnectivity(.ViaWiFi))
     }
+
+    func test__condition_succeeds_when_only_wifi_accepted_and_only_wifi_available() {
+        network.flags = [.Reachable]
+
+        let operation = TestOperation()
+        operation.addCondition(ReachabilityCondition(url: url, connectivity: .ViaWiFi, reachability: manager))
+
+        waitForOperation(operation)
+
+        XCTAssertTrue(operation.didExecute)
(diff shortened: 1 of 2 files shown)


danthorpe:
Don't worry about the CI - there are a couple of flaky tests in there which look to be falling over for some reason.

Tests/Features/ReachabilityConditionTests.swift
```diff
@@ -98,6 +98,17 @@ class ReachabilityConditionTests: OperationTests {
 
         XCTAssertEqual(error, ReachabilityCondition.Error.NotReachableWithConnectivity(.ViaWiFi))
     }
+
+    func test__condition_succeeds_when_only_wifi_accepted_and_only_wifi_available() {
+        let operation = TestOperation()
+
```
danthorpe: Looks like you need to attach the condition here right? I guess you're waiting for the other PR to be merged though?

itsthejb: Ah good spot - total rushed blunder there! Will fix tomorrow

itsthejb:
Testing blunder fixed :wink:

danthorpe:
Thanks for this one too @itsthejb  - sorry it took so long to merge!

itsthejb:
:+1:

PR #223, opened 2016-03-10
Fix suspected retain cycles in result injection

itsthejb:
Hi there,

Courtesy of @estromlund. These appear to be retain cycles; the `previous` operation is added as a dependency of the `subsequent` operation, but the block retains the `subsequent` operation, so we would seem to have a clear retain cycle. Certainly adding this fixed some cycles for us. Please let us know if you think we might be using things incorrectly.

Thanks!


danthorpe:
Good spot! Thanks @itsthejb and @estromlund!

itsthejb:
:+1:

PR #224, opened 2016-03-10
Use log.warning for all error cases

itsthejb:
Hi again ;)

A minor one: previously using `verbose` for all logging, which means that genuine error cases can get lost in a verbose log (currently we're looking at everything). Changed to `warning` for when there are errors. With coloured output this should mean that errors don't get lost in the log.

Subjective of course. Let us know what you think.

### DIFF
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -112,7 +112,7 @@ public class GroupOperation: Operation {
      - parameter error: an ErrorType to append.
     */
     public final func aggregateError(error: ErrorType) {
-        log.verbose("Aggregated error: \(error)")
+        log.warning("Aggregated error: \(error)")
         _aggregateErrors.append(error)
     }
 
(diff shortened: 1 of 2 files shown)


danthorpe:
I did use to have it like this. I changed it because I just sort of thought that it could be a bit much. I think before I can merge it though I'll need to change the unit tests to set the default log level to `.Fatal` as there are some tests which run 1000s of failing operations.

The OS X tests seem to be failing on some odd things - so I'll look into this tonight and will merge these changes.

danthorpe:
Also, thanks so much for the feedback - I really appreciate the fine-tuning, definitely feel like the project is getting better and better.

itsthejb:
Well on our project we're moving from the proof of concept stage to the polishing stage, so we're now turning down the logging - but we *do* want to see logs of anything going wrong, so I think this would generally be welcomed. Let me know if you'd like me to tweak anything here.

itsthejb:
PS. And no idea what we'd be doing without this framework :+1: :wink:

danthorpe:
Yeah, no worries - will merge this, or at least will make the same changes today.

Currently working on preparing the last bits and pieces for v2.8.0 and will then branch that off as `swift_2.1` and `development` will move  to Xcode 7.3 & Swift 2.2.

itsthejb:
Ah, yes, we have *that* to look forward to... :skull: :stuck_out_tongue:

PR #229, opened 2016-03-13
Inject results with WillFinishObserver to avoid race condition

estromlund:
When injecting results using a `DidFinishObserver`, there is a race condition when the dependent operation becomes ready and executes *after* its dependency finishes but *before* the dependency's `DidFinishObserver` is triggered. This means the dependent operation occasionally executes before its dependency's result is injected and fails due to a missing `requirement`.

Switching this to a `WillFinishObserver` ensures that the result is injected before the dependency finishes, and before the dependent operation is allowed to execute.


danthorpe:
Great catch @estromlund. Thanks!

I think would probably have been a regression in 2.7.0, as originally the did finish event was actually triggered where will finish is now triggered.

PR #232, opened 2016-03-14
Always create background tasks for background observer

itsthejb:
Hi again :wink: 

Here's a suggested change: from reading [this](https://spin.atomicobject.com/2015/05/14/ios-background-task-reactivecocoa/) article, I started wondering if `BackgroundObserver()` shouldn't always generate a background task identifier if attached, and respond as before only to the operation finishing. There doesn't seem to be any harm in doing so, and essentially this observer can be freely attached to any operation that would want to progress in the background - and be sure it's been assigned an identifier. It does not appear that `beginBackgroundTaskWithName()` cares when the identifier is created.

Context: in our case we basically want *all* our operations to continue in the background for as long as possible, so I'm currently attaching this condition at the queue level.

Let me know what you think :+1:

### DIFF
--- a/Sources/Core/iOS/BackgroundObserver.swift
+++ b/Sources/Core/iOS/BackgroundObserver.swift
@@ -25,8 +25,6 @@ if the app goes in the background.
 */
 public class BackgroundObserver: NSObject {
 
-    static let backgroundTaskName = "Background Operation Observer"
-
     private var identifier: UIBackgroundTaskIdentifier? = .None
     private let application: BackgroundTaskApplicationInterface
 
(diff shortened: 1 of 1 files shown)


itsthejb:
Expected test failures :wink: If you think this is interesting, I can tidy those up

danthorpe:
Yeah, make sense to me.

I'm thinking that it would be even better to just make it part of `Operation` directly, with a boolean which defaults to true. That was you don't need to attach the `BackgroundObserver` yourself, it just gets one in the `Operation`'s `init`.

danthorpe:
So, yeah, I think this is interesting :+1:

Perhaps the best thing to do is for you to "tidy up the test cases" - they actually fail to compile, because the static property has been removed.

And I will look into making any changes on `Operation` myself. Sounds good?

itsthejb:
Sounds good!

danthorpe:
Going to close this, as it's implemented in #233

PR #234, opened 2016-03-15
Fix race condition when cancelling group operation

MrAlek:
Currently, there's a race condition in `GroupOperation` when cancelling. Since the group first cancels its composing operations before setting its own `_cancelled` flag, sometimes the internal queue will finish and the group operation will finish before `super.cancel()` is run, resulting in the `_cancelled` flag never being set.

Look at 431dd07 for a test proving the race condition exists by inserting a sleep between `queue.cancelAllOperations` and `super.cancel()`

This PR remedies the problem by calling `super` first and setting the groups own state before cancelling its children.

### DIFF
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -55,10 +55,10 @@ public class GroupOperation: Operation {
 
     /// Override of public method
     public override func cancel() {
+        super.cancel()
         queue.cancelAllOperations()
         queue.suspended = false
         operations.forEach { $0.cancel() }
-        super.cancel()
     }
 
     /**
--- a/Tests/Core/GroupOperationTests.swift
+++ b/Tests/Core/GroupOperationTests.swift
@@ -26,6 +26,37 @@ class GroupOperationTests: OperationTests {
             XCTAssertTrue(op.cancelled)
         }
     }
+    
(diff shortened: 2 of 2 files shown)


danthorpe:
Oh great - thanks for this - great spot and fix. :+1:  Will look into it shortly!

Right off the bat - is there anyway to reproduce it without using `sleep(1)` - don't want to introduce stuff like that into the tests as it slows things down. :)

MrAlek:
Cool :)

Well race conditions are always a bit iffy to test. Even the current test doesn't really help if someone changes the implementation since the race occurs just between row 58 and 61 and there's no easy way to inject a sleep there even with subclassing. I used it mainly to prove the issue.

danthorpe:
One thing which I've considered doing with `GroupOperation` is to allow injection of the queue into it, or make it an internally settable property.

In such a scenario you'd then be able to replace the queue with a `SleepingOperationQueue` and put the delay into the `cancelAllOperations` implementation. In other words, modify the queue instead of the class under test.

But to be honest - the core change is the moving of `super.cancel()`. So, I'm happy to merge your PR as it is and I might refactor `GroupOperation` and the test in another ticket.

MrAlek:
Sounds awesome :+1: Thanks!

danthorpe:
Thanks a lot for this @MrAlek! Will pull this into the v2.8.0 release, which will land later this week.

PR #235, opened 2016-03-15
Operation Profiler

danthorpe:
This is an idea prompted by some stuff done by @itsthejb in his fork.

The aim is to provide a way to extract performance metrics from `Operation` instances in a way which is amenable to feeding into analytics and tooling. The purpose for this might be to provide end-to-end integration metrics or for full system diagnostics, or even just to set some idea of the performance of a particular `Operation`.

Now, I hasten to add, that there is a strong argument that shipping apps should not include any kind of performance metrics code. The flip side of the coin, especially if you have an app which is getting 10k-100k or more DAUs is that it is very tricky to diagnose performance problems across all of your user's environments. From different devices, to networks, carriers and countless others. The reality is that performance tests in your unit test bundle will not help diagnose or report the real world performance of your application.

So, here is the idea. We have `OperationProfiler` which itself is just an `OperationObserverType`. It's job is simply to record the timing split between all the events the Operation encounters, so it's a little bit like the deprecated `OperationLogger`.

However, it will be initialized (or is type constrained) so that when the operation finishes, it will report the metrics to a logger. For example, the profiling could just be sent onto the console. More likely however, framework consumers would have their own logger for this information, to e.g. store it and then send it into their analytics systems. The framework will not provide these, just the profiler, and a specialized logger.

### DIFF
--- a/Operations.xcodeproj/project.pbxproj
+++ b/Operations.xcodeproj/project.pbxproj
@@ -693,13 +693,13 @@
 				652617151C11F5EB00654091 /* NetworkObserverTests.swift */,
 				652617161C11F5EB00654091 /* NoFailedDependenciesConditionTests.swift */,
 				652617171C11F5EB00654091 /* OperationTests.swift */,
+				6565FFE91C98C860008B5248 /* ProfilerTests.swift */,
 				65E4CDE61C33F97900B9F9D8 /* RepeatedOperationTests.swift */,
 				65691C9D1C21A29C00AF06B4 /* ResultInjectionTests.swift */,
 				65AE14961C34865E00F592D2 /* RetryOperationTests.swift */,
 				652617181C11F5EB00654091 /* SilentConditionTests.swift */,
 				652617191C11F5EB00654091 /* TimeoutObserverTests.swift */,
 				6526171A1C11F5EB00654091 /* UIOperationTests.swift */,
-				6565FFE91C98C860008B5248 /* ProfilerTests.swift */,
 			);
 			path = Core;
 			sourceTree = "<group>";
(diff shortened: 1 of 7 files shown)


danthorpe:
Things to consider... currently the profiler has the lifespan of an Observer - it is attached to the `Operation`. Therefore, if the operation produces another `Operation`, the profiler needs to produce another profiler, yet, it cannot own or in anyway retain it.

Therefore, each profiler, will need to have it's own unique identifier which gets generated when created. If the attached operation produces another operation, that should get logged in the metrics as a event, with a reference to the produced profiler's identifier.

More than likely, in addition to profilers & reporters we will actually want to have a `ProfileManager` which will collect the profilers to correctly tie together produced and/or group operations.

danthorpe:
So, if an operation, which is being profiled produces another operation using `produceOperation`. This now also gets profiled. This works as a sort of daisy chain, where the child profiler, actually must hold onto its parent profiler (which acts as its reporter). This way, the first operation can finish (and get deallocated) before the produced operation finishes. In which case, the first profiler (or parent) would be deallocated (as it's owned by the operation instance). However, because the child/spawned operation is running, with its own profiler, which owns the first profiler, it doesn't get deallocated until all the operations finish.

So, when this finally reports at the first profiler, it looks like this:

The log output for this looks a bit like this:

```
One #0AED6596-5998-4716-B8D2-CD2F67937489: finished with perfomance metrics:
+0.000547945499420166 Attached
+0.00269895792007446 Started
+0.00420397520065308 Produced Two #EFB859E8-AA35-4C78-AD2F-DB44EF7FA65D
  +2.30073928833008e-05 Attached
  +0.000799000263214111 Started
  +0.00531101226806641 Produced Three #E88AFB71-7F48-4854-A952-3B26E24F143E
    +2.10404396057129e-05 Attached
    +0.000181019306182861 Started
    +0.326790034770966 Finished
  +0.298131048679352 Finished
+0.192564964294434 Finished
```

The above scenario is with the first operation `One` which finishes after 0.1 seconds, produces a 2nd operation, `Two` which finishes after 0.2 seconds, which produces a 3rd operation, `Three` which finishes after 0.3 seconds.

So, looking at the log, you can see the nested timings of each operation as it moves through its states.

Still to work on:

- [ ] Unit tests
- [ ] Special handling for `GroupOperation`.

@itsthejb - thoughts?

itsthejb:
Well this certainly looks great! Glad to have given some inspiration! Can't contribute much more right now, but I'll definitely be keen to try it out at a later date. :+1:

PR #236, opened 2016-03-17
Finish AlertOperation after call to handler

felix-dumit:
Is there a reason why it needed to be the other way around?

Currently I have a `GroupOperation` where the `AlertOperation` is the only one created at startup. Then depending on the option selected by the alert I add another operation to the group. 

However because finish is called before the alert handler, I get an error saying I can't add an operation to the group after it has finished

### DIFF
--- a/Sources/Core/iOS/AlertOperation.swift
+++ b/Sources/Core/iOS/AlertOperation.swift
@@ -89,8 +89,8 @@ public class AlertOperation<From: PresentingViewController>: Operation {
     public func addActionWithTitle(title: String, style: UIAlertActionStyle = .Default, handler: AlertOperation -> Void = { _ in }) {
         let action = UIAlertAction(title: title, style: style) { [weak self] _ in
             if let weakSelf = self {
-                weakSelf.finish()
                 handler(weakSelf)
+                weakSelf.finish()
             }
         }
         alert.addAction(action)


danthorpe:
Hi @felix-dumit thanks for this!

It's a really good point. Looking at the code, I don't see why the handler is not called before finish. So making this change would allow for adding `Operation` instances after the alert is dismissed directly without using a group - so will merge it.

However... depending on how you've written your `GroupOperation` subclass, you could still get it to work. As with most things there is more than one way! :) So, for example:

```swift
class MyGroupOperation: GroupOperation {

    override func willFinishOperation(operation: NSOperation, withErrors errors: [ErrorType]) {
        if operation === alert {
            addTheNextOperationAfterAlertNow()
        }
    }
}
```

And the above is the standard approach for adding new operations to a group. However, I can totally appreciate that you want to keep your code organized together, and adding the next operation in the callback from the alert also makes a lot of sense.

felix-dumit:
Ok. One question related to this. The operation I produce after the alert click is presenting a view controller, which has the `MutuallyExclusive<UIViewController>` condition. 
It would sometimes happen that it would evaluate conditions while the alert operation had not finished and so it would fail. 
So what I did was add the alert operation as a dependency of this new operation. Is that generally the way to go?

danthorpe:
Hi @felix-dumit yes, that is exactly what you should do.

So you have an alert, and then depending on the result you add a new operation, which is mutually exclusive on `UIViewController`....

So, you are going to want to setup a dependency, as it's likely that you'll trigger your operation before the alert is finished.

PR #244, opened 2016-03-22
Fix two bugs.

difujia:
1. Composed operation is clearly causing a retain cycle (composed op <-> target).
2. Mutually exclusive operations may not execute, please see the added test.

### DIFF
--- a/Sources/Core/Shared/ComposedOperation.swift
+++ b/Sources/Core/Shared/ComposedOperation.swift
@@ -8,26 +8,28 @@
 
 import Foundation
 
-public class ComposedOperation<T: NSOperation>: Operation, OperationDidFinishObserver {
+public class ComposedOperation<T: NSOperation>: Operation {
 
     public let target: Operation
     public var operation: T
 
-    public convenience init(_ operation: T) {
-        self.init(operation: operation)
+    public convenience init(_ op: T) {
+        self.init(operation: op)
     }
 
-    init(operation composed: T) {
-        target = composed as? Operation ?? GroupOperation(operations: [composed])
-        operation = composed
+    init(operation op: T) {
+        self.target = op as? Operation ?? GroupOperation(operations: [op])
+        self.operation = op
         super.init()
         name = "Composed Operation"
-        target.name = "Composed <\(T.self)>"
-        target.addObserver(self)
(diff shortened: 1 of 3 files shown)


Sources/Core/Shared/ComposedOperation.swift
```diff
@@ -8,26 +8,28 @@
...
-    init(operation composed: T) {
-        target = composed as? Operation ?? GroupOperation(operations: [composed])
-        operation = composed
+    init(operation op: T) {
+        self.target = op as? Operation ?? GroupOperation(operations: [op])
+        self.operation = op
         super.init()
         name = "Composed Operation"
-        target.name = "Composed <\(T.self)>"
-        target.addObserver(self)
```
danthorpe: Arrrr!!!! Great spot @difujia!

danthorpe:
Looks good - currently I've got a bit of an issue with my build agents. After updating to the latest build kite agent, OS X 10.11.4, and Xcode 7.3, there are some tests (`ContactsOperation`) which are triggering an auth alert to access Contacts on the build machine - and this is pausing the execution of the tests.

Unfortunately i'm now at work, and I don't have remote access to the build machines. So, will have to wait until this evening to sort this out.

But thanks a lot for the bug fixes - I think that mutually exclusive one has been an issue for a while which I've not been able to properly track down!

Sources/Core/Shared/ComposedOperation.swift
```diff
@@ -8,26 +8,28 @@
...
+        target.name = target.name ?? "Composed <\(T.self)>"
```
danthorpe: Unfortunately this means that there is now a failing test in `ReachableOperationTest`.

danthorpe:
@difujia I have cherry picked your commits into separate branches, fixed the lint issues and a failing unit test, and they've now been pull requested into the `swift/2.1` branch, from where a new release 2.8.1 will be made.

Thanks a lot for the bug fixes - pretty bad that I hadn't spotted these 😞.

difujia:
Good to hear. I have just switched to xcode 7.3 and swift 2.2, and some tests are failing for no reason (bad access in the test code, e.g. PassbookCapabilityTests), so I didn't try out all the tests while making these changes. I guess the latest xcode has some trouble with xctest.

PR #250, opened 2016-03-24
Cancel Group Operation after cancelling Child Operations

felix-dumit:
Again, is there a reason it needs to be the other way around?

I have an instance where after one group operation finishes I need to look at the results of that group's child operations.
If the group operation is cancelled before the children, it finishes before the children have a chance to cancel and update their status.

### DIFF
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -55,10 +55,10 @@ public class GroupOperation: Operation {
 
     /// Override of public method
     public override func cancel() {
-        super.cancel()
         queue.cancelAllOperations()
         queue.suspended = false
         operations.forEach { $0.cancel() }
+        super.cancel()
     }
 
     /**


jshier:
Slightly related question, @danthorpe, but why cancel all of the internal queue's operations, and then tell each individual operation to cancel? Isn't that redundant?

danthorpe:
@felix-dumit - CI is failing at the moment, because I'm in the middle of a change to Swift 2.2.... So just bear with me.

@jshier it's a really good question really. I don't know / remember. This is probably just directly from @davedelong's original sample code. However, I guess I will add that, while technically all the operations in that array are likely in the queue - there isn't actually any guarantees about this. So, it's perhaps some overly defensive coding - but that's not necessarily a bad thing in a general purpose framework.

felix-dumit:
@danthorpe One more thing here, currently if you cancel a group using `group.cancelWithError(error)` the child operations are cancelled without the error. 
Do you think it makes sense to propagate that error to the children?

danthorpe:
@felix-dumit hmm, yeah, I guess so. Perhaps something like

```swift
enum GroupOperationError: ErrorType {
   case ParentGroupCancelledWithError(ErrorType)
}
```

So, that we can propagate the error but wrapped up so that it is clear that it's been cancelled due to the group.

felix-dumit:
@danthorpe Ok, how about this in the `GroupOperation.cancel()`

```swift

enum GroupOperationError: ErrorType {
    case ParentGroupCancelledWithErrors([ErrorType])
}

    /// Override of public method
    public override func cancel() {
        queue.cancelAllOperations()
        queue.suspended = false
        operations.forEach {
            if let op = $0 as? Operation {
                op.cancelWithError(GroupOperationError.ParentGroupCancelledWithErrors(aggregateErrors))
            } else {
                $0.cancel()
            }
        }
        super.cancel()
    }
```

danthorpe:
Yeah, that's what I was thinking.

danthorpe:
I created a new issue #256 for this.

MrAlek:
Sorry to be the party pooper here but this just broke my code related to #234

I guess the real issue is that observers shouldn't be notified until the operation has finished updating all it's state.

danthorpe:
@MrAlek - oh no! I'll look into it. 

Can you get a diff of the change which has introduced the regression for you? Or, what was the last good commit from Operations? the v2.8.0 release?

MrAlek:
@danthorpe It's this one commit and this one line right here :) 2.8.1 is the latest release that still works for me.

It's a classic problem though with composite structures. The group operation must be able to update all of its internal state (which include cancelling child operations) before any reasoning is done about it's external state which triggers observer callbacks. I don't have a proposed fix at the top of my mind but could look at it when I have time. Maybe create an issue out of it? For me I'm good with running 2.8.1 meanwhile.

danthorpe:
@felix-dumit @MrAlek well, I resolved this situation for "finish" state by adding `willFinish` & `didFinish` observers.

It actually makes sense to do the same for cancellation...

danthorpe:
@MrAlek just to clarify what you're saying here though: namely that the group should update all of its "internal" state (including cancelling child operations) before any "external" state is changed which triggers observer callbacks.

This is what this change *actually does*. It's a little bit confusing because the "external" state you refer to, is that inherited from `Operation` and is changed by calling `super.cancel()`, which is done after it cancels the child operations.

So, I'm not sure I agree with your assertions here...

Anyway, I've created a pull request with some changes to look at this, can we move the discussion over to #293?

@felix-dumit in terms of your original issue here. I think the thing to ensure you're doing is, that you check the `cancelled` state of your group - just because it has finished does not mean it wasn't cancelled or finished due to errors or conditions not being met.

PR #241, opened 2016-03-24
Make changes for Swift 2.2

danthorpe:


danthorpe:
So, looks like the overall architecture of `CapabilityType` could be a problem with Swift 2.2 - the tests on iOS seem to fail for the same reason that @gvarela has an issue with `Capability.Location`.

Basically there is crash in Swift:

swift 2 2 crash

Not really sure how to address this, other than to refactor `CapabilityType` - which I have been considering.

danthorpe:
Going to close this, and merge these changes into a "working development" branch for Swift 2.2: `swift/2.2`

PR #251, opened 2016-03-25
Refactor Capabilities

danthorpe:
There seems to be a pretty serious issue with the way Capability has its Registrar types. Causing a EXC_BAD_ACCESS somewhere.

Now, as it happens, the `Registrar` protocol stuff can probably be simplified and removed entirely from the constructors. This would be similar to how the `HostReachabilityType` was refactored into a variable property with a default value.

This would grant nicer public interfaces.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,6 +1,5 @@
 excluded:
   - Tests
-use_nested_configs: true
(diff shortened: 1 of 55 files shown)


danthorpe:
Damnit - wrong target branch.

PR #255, opened 2016-03-25
Pass NetworkOperation error if present when operation finishes

estromlund:
Hi again! I noticed that when wrapping a `NetworkOperation` in a `RetryOperation`, it would never retry because the error is not passed. Passing the `task` error in `finished()` seems most straightforward to me (it is `nil` for successful operations) but I'm happy to consider other approaches.


danthorpe:
Hi @estromlund - yeah, seems to be a bit of an obvious one! 👍

Is there any chance you could send this PR to `swift/2.1` though? The `development` branch has moved on to Swift 2.2. Alternatively can you pull `development` into your fork and update your branch?

estromlund:
No problem, just created #260 on `swift/2.1`

PR #256, opened 2016-03-29
Group operation children are cancelled with parent error

danthorpe:


felix-dumit:
I wonder if the wrapping of the parents error to pass to the child is a bit of over-engineering and will actually make it more painful to handle the errors.

Here's the example where I wanted to use this scenario:

I have a `GroupOperation` in which all the child operations are saving a file to disk. What I want is that if any of the child operations fail because the device is out of space, I want to cancel all the pending/remaining operations with the same error.
Here is the relevant method of the `GroupOperation`

```swift
    override func willFinishOperation(operation: NSOperation, withErrors errors: [ErrorType]) {
        
       // is there an easier way to do this?
        let outOfSpace = errors.flatMap { $0 as? SavingError }.filter { $0 == .NoSpace }.count > 0
        if outOfSpace {
            self.cancelWithError(SavingError.NoSpace)
        }
    }
```

jshier:
@felix-dumit I like to wrap logic like that in a `CollectionType` extension:
```swift
extension CollectionType where Generator.Element == ErrorType {
    var isOutOfSpaceError: Bool {
        return flatMap { $0 as? SavingError }.contains { $0 == .NoSpace }
    }
}
```
So it can be used like:
```swift
if errors.isOutOfSpaceError {
    cancelWithError(SavingError.NoSpace)
}
```
There are other things you can do to make it cleaner, like having a property on your `SavingError` that determines whether it should be passed to child operations, etc.

Otherwise, it seems necessary to wrap the errors being passed in, otherwise there'd be no way for child operations to tell the difference between being cancelled by their own error or a parent error.

danthorpe:
Hi @felix-dumit - I really agree with everything @jshier has put here. Great tip on using an extension on `CollectionType` to make it clearer to read the error handling.

I'll also add, that I think there is something to be said for how Apple recommend handling CloudKit errors, which is to only worry about the first error in a list of errors. Their logic being that subsequent errors are either knock on effects of the first error, or redundant given the first error. With this in mind, I approach `GroupOperation` and finished `[ErrorType]` in a similar way, either using `errors.isEmpty` to follow the happy path, or switching on `errors.first` to follow the sad path.

Additionally, as I've said in a few other questions on this project - using an `enum` is a great way to group related errors together as an `ErrorType` but using a more sophisticated error type should not be discouraged.

Lastly - I think I will change this so that, instead of adding a new `GroupOperationError`, instead we have a new case in the existing `OperationError` type. This sort of makes sense to me, as framework consumers should not need to hunt through different kinds of `ErrorType`s for "core" level operations. It also means that `ComposedOperation` can and should use the same mechanism.

I'll update this PR later in the day.

danthorpe:
@felix-dumit - I've not heard back from you yet on this yet, but I'm going to merge it.

felix-dumit:
@danthorpe Sorry about that, I think your points make sense, specially when only handling the first error.
I'll definitely use the `CollectionType` extension by @jshier as well!

danthorpe:
@felix-dumit - yeah, that's nice. But also, if your `ErrorType`s conform to `Equatable` then can't you just use the built in `contains` method?

Any, lets face it, there is no reason why all your value types should not conform to `Equatable` 😬

felix-dumit:
The `contains` method would only work if the array was typed with my `[CustomError]` type instead of just `[ErrorType]`.

danthorpe:
Right - yeah, of course. Duh.

PR #260, opened 2016-03-30
Pass network error if present when operation finishes

estromlund:
Rebasing #255 onto `swift/2.1`


danthorpe:
Hi @estromlund - gonna merge this.

I know CI is failing the unit tests - but it's only because I cannot run it for Swift 2.1 when building from forks. And the tests fail on Swift 2.2 because the branch doesn't have some fixes for Swift 2.2 on it...

So.... not to worry. Pretty sure that it'll be okay once i merge into Swift 2.1.

Thanks again for the changes!

Issue #263, opened 2016-04-05
GroupOperation - User Initiated


ryanjm:
I just added this to one of a `GroupOperation` in my project and I was wondering you thought it would make sense in the project itself:

```
    override var userInitiated: Bool {
        get {
            return super.userInitiated
        }
        set {
            for operation in operations as! [Operation] {
                operation.userInitiated = newValue
            }
            super.userInitiated = newValue
        }
    }
```

The reasoning behind this is that often times in a group operation your `init` method will create many additional operations to go into the group. But you can't set the `userInitiated` property until later and you'll usually want that to apply to all operations in the group.

Example: I have a group operation for downloading a list of files. I have another operation, `DownloadFileOp`, which downloads one file so the group operation creates a lot of these. `DownloadFileOp` has a conditional that says to only download if on Wifi OR the user initiated the download. Therefore I want those operations to have their `userInitiated` property updated when the group operation's is set.

Is there a better / more appropriate way of handling this? Is it better to create a connivence `init` that is used to set `userInitiated` before the other operations are created?

If this was to be added to the project I think it'd need to handle the following:

- If additional operations are added to the queue later then they should get the same `userInitiated` value.
- The loop above should handle the case that some of the operations might not be of `Operation` type.

ryanjm:
I tried the `convince init` route and the compiler won't let me due to:

> Use of 'self' in property access 'userInitiated' before self.initialized self.

Here was my code

```swift
convenience init(attachments: [Attachment], userInitiated ut: Bool) {
    userInitiated = ut
    self.init(attachments: attachments)
}
```

danthorpe:
Hi @ryanjm - that for the feedback, quite a subtle yet interesting point here.

First up, yep, I definitely think that `GroupOperation` should set the *quality of service* for its child operations to match its own - which would require an override similar to what you propose here. 

But stepping back a bit... this `userInitiated` property is directly from @davedelong's sample code from WWDC. At the moment, it sort of conflates the concept of *intent* with the quality of service (which is a mechanism to indicate to the queue the priority of the block). And thinking about this - I'm not a huge fan.

So I propose doing something like this instead:

```swift
enum UserIntent {
    case Implicit, Explicit

    var qualityOfService: NSQualityOfService {
        switch self {
        case .Implicit: return .Default
        case .Explicit: return .UserInitiated
        }
    }
}

class Operation: NSOperation {
   // etc

    var userIntent: UserIntent = .Implicit {
        didSet {
[...]

PR #265, opened 2016-04-06
Operation User Intent

danthorpe:
This is to address issue #263.

### DIFF
--- a/Sources/Core/Shared/Operation.swift
+++ b/Sources/Core/Shared/Operation.swift
@@ -153,7 +153,7 @@ public class Operation: NSOperation {
             precondition(state < .Executing, "Cannot modify userIntent after execution has begun.")
         }
         didSet {
-            qualityOfService = userIntent.qos
+            setQualityOfServiceFromUserIntent(userIntent)
         }
     }
 
@@ -671,6 +671,10 @@ extension NSOperation {
     */
     public func addDependencies(dependencies: [NSOperation]) {
         dependencies.forEach(addDependency)
+    }
+
+    internal func setQualityOfServiceFromUserIntent(userIntent: Operation.UserIntent) {
+        qualityOfService = userIntent.qos
     }
 }
 
--- a/Sources/Core/Shared/Operation.swift
+++ b/Sources/Core/Shared/Operation.swift
@@ -150,7 +150,7 @@ public class Operation: NSOperation {
      */
     public var userIntent: UserIntent = .Background {
         willSet {
-            precondition(state < .Executing, "Cannot modify userInitiated after execution has begun.")
+            precondition(state < .Executing, "Cannot modify userIntent after execution has begun.")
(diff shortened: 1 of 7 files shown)


Sources/Core/Shared/Operation.swift
```diff
@@ -76,6 +76,24 @@ public class Operation: NSOperation {
...
+    public enum UserIntent: Int {
+        case Background = 0, SideEffect, Initiated
+
+        internal var qos: NSQualityOfService {
+            switch self {
+            case .Initiated, .SideEffect:
+                return .UserInitiated
+            default:
+                return .Default
+            }
```
ryanjm: I wonder if `.Background` is misleading since that doesn't actually correlate to a `.Background` QoS (and shouldn't since that would just be duplicating QoS). I wonder if `None`, `Implicit`, `Indirect`, or `Unintended` might be better. My vote would be for `None`, as in "the user had no intention about this operation."

Also, what is the intention behind `SideEffect`?

jshier:
I guess my big question is, why is this necessary? Isn't setting the QoS on the operation directly enough to get this effect? This just seems to be adding a layer of API above the QoS that has no real gain. What is this API adding?

danthorpe:
Well, the idea behind this is to make it possible to mark up in a more descriptive way what the operation is being used for. I agree that it is a just a thin layer on top of the quality of service. Perhaps we've over discussed it, as it's pretty trivial.

Sources/Core/Shared/Operation.swift
danthorpe: The `.SideEffect` case is to describe operations which are triggered as a side effect of user actions. So in your Post example, after the `DownloadPostOperation` finishes, perhaps a parent groups goes off and downloads all the Comment items - `DownloadCommentOperation` - you might consider these to be side effects.

My thinking was that semantically, they might have different cancellation, early termination or error handling.

danthorpe: But yeah, will rename `.Background` to `.None`.

danthorpe:
Also, while setting the `userIntent` will now set the appropriate `qualityOfService`, because of the way that `NSOperation`s will get their QoS promoted automatically if the queue's QoS is promoted, or another operation is added to the queue the same QoS, you cannot reliably infer the user's intent from the `qualityOfService` property of an operation.

jshier:
True enough. I guess my only question then, since I don't have any feedback on the enum names, is why is it `Int` based?

danthorpe:
It's an Int because it's the easiest way to make it Comparable. When a Group is initialised with an array of operations, we set the groups user intent to the max of those in the array.

ryanjm:
I like the changes and think this looks great.

danthorpe:
Okay - gonna merge! Thanks @jshier @ryanjm

PR #270, opened 2016-04-09
Disable automatic KVO notifications for state

pronebird:
I think KVO wraps setters in will/didChangeValueForKey automatically. Switching `automaticallyNotifiesObserversOfState` produces a different call stack when running from Objective-C. This is pretty much invisible in Swift, but I think this piece is hidden somewhere within "thunk".

`automaticallyNotifiesObserversOfState = YES`

screenshot 2016-04-09 09 57 43

`automaticallyNotifiesObserversOfState = NO`

screenshot 2016-04-09 09 59 05

### DIFF
--- a/Sources/Core/Shared/Operation.swift
+++ b/Sources/Core/Shared/Operation.swift
@@ -94,6 +94,10 @@ public class Operation: NSOperation {
         }
     }
 
+    class func automaticallyNotifiesObserversOfState() -> Bool {
+        return false
+    }
+
     // use the KVO mechanism to indicate that changes to "state" affect other properties as well
     class func keyPathsForValuesAffectingIsReady() -> Set<NSObject> {
         return ["State", "Cancelled"]


danthorpe:
Hi @pronebird - had a look at this, and then read through the documentation for KVO / NSObject again. Not sure why you think this will help - perhaps you can explain your reasoning here.

Anyway, I've pulled this change of yours into another branch where I've also turned on my unit test which re-produces the issue faithfully. Test still fails.

pronebird:
@danthorpe I apologize if that looked like a fix, I didn't advertise it that way. It's more of tidying up of the code.

danthorpe:
Oh, I see - it's just to reduce the number of KVO notifications being sent?

pronebird:
@danthorpe 

Yeah, technically willChange/didChange is called twice in our case since KVO generates them automatically for setter methods, until we explicitly opt-out.

However, notifications should not be produced for incremental calls anyway. This is really tidying up, it has no impact whatsoever.

see: http://petersteinberger.com/blog/2012/dont-call-willchangevalueforkey/

danthorpe:
@pronebird okay, sorry maybe I'm missing something? What makes you say that KVO for `State` is called twice?

pronebird:
@danthorpe weird but what I see is observes are being automatically notified if `automaticallyNotifiesObserversOfState` returns `YES` in Objective-C, will/didChangeValueForKey  is not used in custom setter. Ignore, looks like black magic to me.

danthorpe:
Hi @pronebird - yeah, I spent a bit of time debugging this, could not discern that KVO notifications were being sent multiple times. And, I feel that your change, while not making a difference, adds unnecessarily to the KVO magic going on with `NSOperation`.

PR #273, opened 2016-04-12
Add preferredStyle to AlertOperation initializer

felix-dumit:
I'd like to be able to use alert operation for showing an ActionSheet as well:

`let alertOp = AlertOperation(presentAlertFrom: self, style: .ActionSheet)`

### DIFF
--- a/Sources/Core/iOS/AlertOperation.swift
+++ b/Sources/Core/iOS/AlertOperation.swift
@@ -67,8 +67,8 @@ public class AlertOperation<From: PresentingViewController>: Operation {
     - parameter from: a generic type conforming to `PresentingViewController`,
     such as an `UIViewController`
     */
-    public init(presentAlertFrom from: From) {
-        let controller = UIAlertController(title: .None, message: .None, preferredStyle: .Alert)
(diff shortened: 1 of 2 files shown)


danthorpe:
Looks good. Happy to merge. There isn't actually a test for the style, the only thing I would add would be an assert in to check that the preferred style is set.

danthorpe:
Hi @felix-dumit - leave a comment when you're finished / ready to merge.

felix-dumit:
@danthorpe But it will always be set because of the default parameter?

I also want the `addActionWithTitle` method to return the `UIAlertAction`, so I can customize it or set the `preferredAction` property of the `UIAlertController`, just added a commit for that

danthorpe:
Thanks for the additions! 👍😄

danthorpe:
Yeah, it currently looks like it will always be set from the default parameter. But the test isn't just to ensure that now, it's to ensure that in the future this isn't changed or is broken.

felix-dumit:
@danthorpe Ok I added the tests, is that what you had in mind?

danthorpe:
Great - thanks for adding the test! Will merge shortly.

PR #278, opened 2016-04-15
Refactor OperationCondition to work as an Operation

danthorpe:
The aim here is to take `OperationCondition` and change it so that the evaluation of the condition which already happens asynchronously before the operation executes is just an `Operation` instance which becomes a dependency of the target operation.

Internally, this will be managed using a `GroupOperation` so that can evaluate multiple conditions.

`OperationCondition` supports the concept of optional dependent operations - this will naturally just work out as standard dependencies.

At the same time, this will solve the longstanding issue with `.Ready`, see discussion in #175.

- [x] ~~Define `OperationType`~~ Not technically needed right now
- [x] ~~Define `ConditionOperationType` (refines `OperationType`)~~ Created `ConditionOperation`
- [x] Figure out each condition will evaluate for given the operation. -- We inject the operation into `ConditionOperation` before it is evaluated.
- [x] Create internal subclass of `ConditionOperation` to wrap up existing `OperationCondition` values.
- [x] Create internal custom `Operation` subclass to evaluate all the conditions, more than likely a `GroupOperation` subclass - this is actually incredibly trivial, we just use a plain `GroupOperation`.
- [x] Redefine dependency scheduling:
  - [x] Update `OperationQueue` to add evaluator if there are any conditions. -- It's actually a little bit more complicated than that.
  1. If there are conditions
  2. If there are conditions which are mutually exclusive
    1. Get the mutually exclusive category
    2. Add the operation to the manager for this category
    3. This method now returns the previous operation for that category if it exists
    4. All previous operations are added to a `Set`. This is for later dependency management.
  3. Create the condition evaluator.
    1. Each conditions gets injected with a weak reference to the target operation
    2. The evaluator (a `GroupOperation`) is added as a dependency (but not a *direct* dependency)
    3. It gets a `WillFinishObserver` to cancel the target queue if there are errors.
    4. Every *direct dependency* is added as a dependency of the evaluator. This means that we don't start evaluating conditions (i.e. run the group) until all the normal dependencies have finished.
  4. Get the *indirect* dependencies - these are the *direct* dependencies of the conditions.
  5. Iterate through the *indirect* dependencies, each one is:
    1. Add all the previous mutually exclusive operations as dependencies.
    2. Add all the *direct* dependencies as dependencies
    3. Add it as a dependency of the evaluator.
  6. Add the *indirect* dependencies to the queue
  7. Add the evaluator to the queue

  This means that the scheduling works like this:

  1. Any previous mutually exclusive operation finish (note that an operation can be mutually exclusive against multiple categories).
  2. Note that *direct* dependencies are not mutually exclusive - they do not wait, and should have been added to the queue separately.
  3. After all *direct* dependencies finish, the *indirect* dependencies start.
  4. After all *indirect* dependencies finish, all the conditions are evaluated. Essentially this just means that they are executed, which in turn calls the evaluate function.
  5. If any condition failed, the target operation is cancelled with errors.
  6. If all conditions succeed, the target operation is executed.
- [x] Refactor `Operation` to remove unnecessary `state` and readiness overrides.
- [x] Unit tests
- [ ] Stress tests
- [ ] Rewrite framework conditions


Sources/Core/Shared/Operation.swift
```diff
@@ -379,11 +379,6 @@ public extension Operation {
             }
         })
 
-        // The evaluator waits for all condition dependencies
-        conditions
```
pronebird: Evaluation or any of its dependencies should not run until after previous mutually exclusive operation has finished.

danthorpe: Yeah, still working on this. Might just draw a diagram to explain the scheduling.

Sources/Core/Shared/ConditionOperation.swift
```diff
@@ -0,0 +1,78 @@
...
+ Additionally, conditions are evaluated asynchronously, and indicate
+ failure by passing an ConditionResult enum back.
+
+ */
+public class ConditionOperation: Operation {
+    public typealias CompletionBlockType = ConditionResult -> Void
+
+    public var isMutuallyExclusive: Bool = false
+
+    internal weak var operation: Operation? = .None
```
danthorpe: Technically I think this should be `unowned` but Swift won't allow me to use it.

danthorpe:
FYI - @pronebird ok - I think I've got it all working.

danthorpe:
This introduces some really interesting possibilities. Consider that the following scenario is now possible.

An operation may have normal (*direct*) dependencies. It can also have multiple conditions. Conditions are `Operation` subclasses, so they can have multiple dependencies and conditions of their own. 

Therefore, a subset of the conditions may have dependencies (which are referred to a *indirect* in relation to the target operation). A subset of the conditions may be mutually exclusive.

Given the above, the following will happen:

1. All the direct dependencies will finish before any dependencies of the conditions start.
2. Any dependencies of any mutually exclusive conditions will wait for the previous operation which it's mutually exclusive with to finish. Note that conditions can now have multiple dependencies.
3. Once all dependencies of all conditions have finished, the conditions will be evaluated.
4. If any conditions fails, the target operation will be cancelled with the errors.
5. If all conditions succeed, the target operation will be executed.

pronebird:
@danthorpe that's pretty rad! Glad to see stress tests are making it into Operations too :)

danthorpe:
Damnit - accidentally closed this...

PR #202, opened 2016-04-16
UIActivityIndicatorViewObserver

danthorpe:
Seems like it would be sensible to have a `UIActivityIndicatorViewObserver` class.

### DIFF
--- a/Operations.xcodeproj/project.pbxproj
+++ b/Operations.xcodeproj/project.pbxproj
@@ -238,6 +238,8 @@
 		6526197B1C11FE6D00654091 /* AddressBookConditionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6526171D1C11F5EB00654091 /* AddressBookConditionTests.swift */; };
 		6526197C1C11FE6D00654091 /* AddressBookTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6526171E1C11F5EB00654091 /* AddressBookTests.swift */; };
 		6526197D1C11FE6D00654091 /* ContactsOperationsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 652617201C11F5EB00654091 /* ContactsOperationsTests.swift */; };
+		653D40221CC2591700893746 /* ActivityIndicatorViewObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 653D40211CC2591700893746 /* ActivityIndicatorViewObserver.swift */; };
+		653D40241CC2594800893746 /* ActivityIndicatorViewObserverTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 653D40231CC2594800893746 /* ActivityIndicatorViewObserverTests.swift */; };
 		6565FFE51C98B7AC008B5248 /* Profiler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6565FFE41C98B7AC008B5248 /* Profiler.swift */; };
 		6565FFE61C98B7AC008B5248 /* Profiler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6565FFE41C98B7AC008B5248 /* Profiler.swift */; };
 		6565FFE71C98B7AC008B5248 /* Profiler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6565FFE41C98B7AC008B5248 /* Profiler.swift */; };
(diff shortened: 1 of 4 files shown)


danthorpe:
Oh, damnit - shouldn't have done that. I want you to create the PR from your fork. Okay, well, not to worry, we can start a new PR.

In your fork, if your fork, if you have a remote which points to this repo - fetch from it, and then checkout the branch `feature/OPR-202_uiactivityindicatorview_observer`. On this branch you will see that I've added new files in the correct places and I've just created the stub of an observer. When you add some code to that branch in your fork, create a pull request to my `development` branch in GitHub.

Regarding the code you pasted in above. It all looks very fancy - but, not at all what I'm looking for. The framework already has a mechanism for observing the life cycle of operations with blocks. Introducing a another generic system to perform blocks when operations start/finish which forward to a wrapped object is not good. Why? Because it's not clear (to framework consumers) what it's doing and how it should be used.

Regarding `UIRefreshControl` and `UIActivityIndicatorView` these are quite different. The former is an "active" control which takes user input and can trigger an operation. So, control flow is `UIRefreshControl` -> `Operation`. Whereas `UIActivityIndicatorView` (and `NSProgress`) are the other way around, they are "passive" and should be updated when something happens, the control flow is `Operation` -> `UIActivityIndicatorView`.

This is why, the observer type just needs to implement the existing protocols, and then manage the state of the `UIActivityIndicatorView`. Take a look at how the `NetworkObserver` works too.

I'm going to close this PR now - we'll pick it up again in the PR that you create.

matt-holden:
Thanks!  I think I still had some misunderstandings about the intent.  I re-watched Apple's "Advanced NSOperations" video and started doing what you suggested above.

I'll get it wrapped up in the next day or so (time permitting).

matt-holden:
@danthorpe 

Hi!

I've spent more time diving into the framework and am starting to work on the pull request.   I have a question, though about part of your original comments on this task:

> Now, the interesting thing, which would be great if you could do, would be to handle "child operations". There is `GroupOperationWillAddChildObserver` and `OperationDidProduceOperationObserver` which the observer could listen for too.

 
Hopefully I'm not completely off-base again, but I don't see (or understand) why we'd do either:

**`GroupOperationWillAddChildObserver`:**

Because the group operation's cancel & finish observers  won't be called until *all* child operations have been completed, I'm not understanding why we'd concern ourselves with the state of each individual child operation.   ...?

**`OperationDidProduceOperationObserver`:**  This one definitely makes more sense to me, but  I think it could also have some subtle, unintended consequences.  In the contrived example that follows, if `UIActivityIndicatorViewObserver` were to attach itself to "produced" operations, our activity indicator will animate for longer than the developer likely intended:


```swift

// **Client Code**
// Runs `AddImageFiltersToPhotoOperation` and displays its resulting image in the UI.
// While the image is processing, the activity indicator will begin animating. 
func showOurCoolUI() {
    let imageOperation = AddImageFiltersToPhotoOperation(myImage) { newImage, err in
        dispatch_async(Queue.Main.queue) {
            imageView.image = newImage
        }
    }

    // Observe with the new `UIActivityIndicatorViewObserver`
    imageOperation.addObserver(UIActivityIndicatorViewObserver(activityIndicatorView: self.activityIndicator))

    queue.addOperation(imageOperation)
}

/// Applies filters to an image (imaginary ultra-complex and slow image filtering)
/// - note: Also quietly caches the result to disk after calling the completionHandler.
/// ^^^^^ (This is would probably be considered an anti-pattern, but I don't have a better example off-hand.)
class AddImageFiltersToPhotoOperation: Operation {

    typealias CompletionBlockType = (UIImage, NSError) -> ()

    private let originalImage: UIImage
    private let completion: CompletionBlockType

    init(image: UIImage, completion: CompletionBlockType) {
        self.originalImage = image
        self.completion = completion
    }
    

    override func execute() {

        let newImage = myImageTransformationFunction(originalImage)

        // Quitely produce an operation to cache the resulting image data to disk
        produceOperation(MyCacheImageOperation(image: newImage))


        completionBlock(newImage)
        finish()
    }
}

```

I think that if I were writing the above code, my desired behavior would be for the activity view to stop animating as soon as the new, filtered image appeared in the UI... However, the actual behavior will be to wait for the produced "caching" operation to finish.

**Final thought:** The current implementation of `NetworkObserver` does not pay attention to produced operations... so adding that behavior to `UIActivityIndicatorViewObserver` would be inconsistent with `NetworkObserver`.. which is in many ways a similar class.

---------------

**The usual disclaimer** I think I have a basic understanding of most of the framework's moving parts; I still may be completely wrong here 😛 


Thanks!

danthorpe:
Yeah - likely it might not make sense - it was just an idea off the top of my head - 😁 - but great that you'd considered it.

Also - this is one of the challenges of writing a framework vs writing application code. We can't tell what the framework consumer (i.e. application developer who is using Operations) will want to do or how they are using the classes. The idea is to create functionality which is clear, easy to use, flexible & stable. At a first pass I'd suggest leaving these off for now, just focus on starting/stopping a single instance.

Also, I have a Slack team where we can discuss this more (and there have already been some discussions). Not sure if you use Slack, but I'll invite you.

matt-holden:
Yeah I can definitely open up Slack. Bot-scraping-proof email coming up 😄 :

invite_me_to_slack_jftholden at yahoo  <-- but, remove "invite_me_to_slack_" from the beginning  ;)

PR #293, opened 2016-04-20
Provide will & did observers for changes to cancellation & start.

danthorpe:
There is a bit of a fundamental problem at the moment, where observers either get notified before or after state changes. And this is a little bit arbitrary.

However, framework consumers will need or want to observe didn't changes to state for different reasons and scenarios. So, we'll add before and after for `start` and `cancel` state changes (to match what is there already for `finish`).

- [x] `didStart` -> `willExecute`
- [x] `willCancel`
- [x] `didCancel`
- [x] ~~make~~ prepare for `cancel` to be final.


Sources/Core/Shared/GroupOperation.swift
```diff
@@ -52,30 +52,26 @@ public class GroupOperation: Operation {
...
-
-    /// Override of public method
-    public override func cancelWithErrors(errors: [ErrorType]) {
-        queue.suspended = false
-        let (nsops, ops) = operations.splitNSOperationsAndOperations
-        nsops.forEach { $0.cancel() }
-        ops.forEach { $0.cancelWithError(OperationError.ParentOperationCancelledWithErrors(errors)) }
-        super.cancelWithErrors(errors)
-    }
-
```
danthorpe: We use a `WillCancelObserver` instead of overriding the `cancel` methods. I'd be tempted to make `cancel` a final method in `Operation`.

MrAlek: Agree on making `cancel` final 👍

MrAlek:
I think this is a great addition to the framework and unless I'm mistaken, it also solves the race condition for `didFinish` when cancelling group operations by suspending the internal queue in `willCancel` so that the `finishOperation` is never executed. 👏

danthorpe:
Oh right yeah - I didn't think about that situation - good point.

felix-dumit:
@danthorpe  I was testing this branch in my code and I found an issue, which might be unrelated.
Basically I was cancelling a group operation and not all child operations were being cancelled. The ones that were not cancelled were added after initialization with `group.addOperation(op)`.
I see in `GroupOperation.addOperation` that it only adds new operations to the queue, but does not update the internal `operations` array, which is used in the willCancelBlock.
Was this intended behaviour?

danthorpe:
Thanks @felix-dumit, I think that's a bug - not intended behaviour.

- All operations should be cancelled
- The `operations` property is supposed to be the the operations which the group is initialised with. 

I think if you create a group with an operations argument, that should be what you get back when you ask from the operations property. For operations added as children - I don't think they should be inside that array. However, it would make sense if they were available somehow - though another array perhaps?

felix-dumit:
@danthorpe Actually it seems the call to `queue.cancelAllOperations()` which previously was in the `cancel` method, now it not being called from the `WillCancel` block. That might fix it without having to keep track of the added properties

danthorpe:
@felix-dumit - yeah, that's the bug, I think we need to keep track of them to pass the error on.

felix-dumit:
@danthorpe Maybe something like this:

```swift
  var extraOperations = [Operation]()
    public func addOperations(operations: [NSOperation]) {
        if operations.count > 0 {
            operations.forEachOperation { $0.log.severity = log.severity }
            queue.addOperations(operations)
        }
        extraOperations.appendContentsOf(operations.filter { !self.operations.contains($0) })
    }
```

Then in `WillCancel` block use `operations + extraOperations`

MrAlek:
@danthorpe @felix-dumit Adding extra state with another variable would make the code even harder to reason about.

Why not make `operations` a `public private(set) var` and let the queue be private? Then `GroupOperation` would add the initial operations to its queue during `init` and in `addOperations` you would append to the var, configure the severity and add to the queue. It also protects the queue from getting its delegate changed and from API consumers accessing the `finishingOperation`.

danthorpe:
@MrAlek - how important do you think it should be that the `operation` property is the same as the initialiser argument?

You make a good point r.e. *even more* state. It's a tradeoff between managing complexity inside the class, and what class consumers should reasonably expect from the public interface. Not to mention breaking existing behaviour...

danthorpe:
Also, when I originally wrote the class, `queue` was private. It got made public as part of some work on a "CollectionOperation" which I later deleted. So yeah, it should be made public.

MrAlek:
@danthorpe Personally, I think it's unimportant for `operations` to be a `let`. Subclasses can even add operations past init-time in their own initializers that API consumers might be unaware of.

danthorpe:
@felix-dumit - thoughts?

MrAlek:
Example of above:

```swift
class GroupSubclass: GroupOperation {
    init() {
        super.init(operations: [])
        addOperation(NSOperation())
    }
}


let groupOperation: GroupOperation = GroupSubclass()
```
In this case, `groupOperation.operations` would be empty, which is kind of weird from a consumer's point of view.

felix-dumit:
I think it makes more sense for the `operations` array to be updated with all the operations in the group.
In fact, as a consumer of the class I thought it worked that way until I "looked under the hood".

danthorpe:
Okay - will make the changes to stick new operations in `operations`, `queue` becomes private.

danthorpe:
@felix-dumit @MrAlek if either of you want to join discussions on Slack about this framework too - ping your email to me either via a DM on Twitter or my github user at me dot com. 😀

danthorpe:
My Twitter DMs are now open.

felix-dumit:
@danthorpe Sometimes I want my groupOperations to perform only one operation at a time, currently I do this with `group.queue.maxConcurrentOperations = 1`, how would I implement that if the `queue` becomes private?
Maybe we need a public property that will get/set from the queue?

danthorpe:
@felix-dumit hmm, maybe. Let me think on this.

One way that we could do it, is to allow you to inject your own queue into the group. For example:

```swift
init(queue: OperationQueue = OperationQueue(), operations: [NSOperation]) {
    // etc
}
```

felix-dumit:
@danthorpe There is one scenario where if my app gets a memory warning I switch from multiple operations at a time to just one, so I think it would still be interesting to have the ability to change it.

danthorpe:
Yeah - it's a property which is run-time editable, so perhaps that is something which the group operation should expose.

I'll have another look the `NSOperationQueue` API.

danthorpe:
Okay - going to merge this now. There will be an intermediate release shortly (v2.10.0 #298) which does not contain these changes. 

This will be released as part of v3.0.

PR #305, opened 2016-04-29
Creating a UI which reflects CLLocationAuthorizationStatus

J-Swift:
I'm not sure why it thinks this is the right thing to do, but in some (but perhaps not all?) cases when asking CLLocationManager to authorize us for Location updates, it will immediately respond with 'NotDetermined' and then show the user the dialog asking for permission. After the dialog is dismissed, it proceeds to call us back with the proper status value.

Without this fix, adding a condition like this: 

    addCondition(AuthorizedFor(Capability.Location()))

will immediately fail our operation due to the premature callback.

### DIFF
--- a/Sources/Features/iOS/LocationCapability.swift
+++ b/Sources/Features/iOS/LocationCapability.swift
@@ -183,6 +183,10 @@ public class LocationCapability: NSObject, CLLocationManagerDelegate, Capability
      - parameter status: the CLAuthorizationStatus
      */
     @objc public func locationManager(manager: CLLocationManager, didChangeAuthorizationStatus status: CLAuthorizationStatus) {
+        guard status != .NotDetermined else {
+            return
+        }
+
         authorizationCompletionBlock?()
     }
 }


danthorpe:
Hi @J-Swift - thanks for this! I think this must be a regression either from the Swift 2.2 changes to capabilities or the conditions as operations stuff.

I can't get into it right now - but will do later this weekend!

J-Swift:
Hmm, I think I must have found myself in between that big refactor for 'conditions as operations'. I was hotfixing my code locally with this based on a v2.10.0 release, then did a quick pull/update/push to get the PR submitted, not realizing that I wasn't operating on the latest version of the code.

So, this might/should be able to be put on hold until I verify that what I was seeing is still an issue or not with the refactorings you mention. Were you planning on cutting that release soon?

danthorpe:
Not releasing v3 for a couple of weeks - but will fix this as v2.10.1 if needed, and then have the fix on development too.

Not had a chance to look into this - but will do tonight / a few hours.

Can you confirm if the issue exists on 2.10? 🤔😀

J-Swift:
Switching back to v2.10.0 and testing on the sim I can't reproduce on a clean install, but can if I perform the following:

1. run on clean install
1. ask for permissions (this correctly does not trigger the callback early)
1. while the dialog is still on screen (i.e. without allowing/disallowing the permission), stop the sim
1. rerun the app
1. ask for permissions

This causes the premature failure I mentioned due to a .NotDetermined status. Perhaps the device is returning some sort of cached value that would have been created by that initial run-through?

I'll try and test on a device as well, but xcode is not allowing me to run on one at the moment so I can't test that unfortunately. I'll update here if I end up getting around said issue.

danthorpe:
Okay, great. I'm back at my desk now, so looking into this now.

I think for this, I'm most interested in behaviour of while on device in an app. Previously, while it would essentially always query to see what the authorisation status was, the OS would only present the dialog if it were not determined. However, if the user had already responded to the challenge, then it would not re-present the dialog.

This is quite important, as the user could accept the authorisation request, and then later on, disable the permissions in their Settings.

So, I don't think you're change is an issue - but also, curious to see why it is happening - hoping it's just a result of the whole conditions as operations refactor.

J-Swift:
So I still wasn't able to run through the xcode debugger, but building and installing as an adhoc, I was able to reproduce on the device in a similar fashion:

1. run on clean install
1. ask for permissions (this correctly does not trigger the callback early)
1. while the dialog is still on screen (i.e. without allowing/disallowing the permission), turn off the device
1. turn the device back on
1. rerun the app
1. ask for permissions

So, in a contrived but existing scenario, it is unfortunately still possible to be called back prematurely by the system.

danthorpe:
Hi @J-Swift - can you share a bit of your code for how you're using the Location operations?

I've tried re-producing on a device this using my the [Location example](https://github.com/danthorpe/Examples/blob/development/Operations/Permissions/Permissions/LocationViewController.swift) in my Permissions.app - but behaves pretty much as expected. This is using v2.10.

danthorpe:
Looking at `development` now.

danthorpe:
So, yes - definitely an issue with `development` as it runs the dependency immediately.

J-Swift:
To be clear, this is regarding the LocationCondition, not LocationOperations. Here is a pared down example of my class (please disregard poor swift abilities, I have very limited experience with it). It is triggered when a user enables location tracking within a settings page:

```
public class RequestUserLocationTrackingOperation : Operation {
  private let _onComplete: (Bool -> Void)?
  
  public init(onComplete: (Bool -> Void)? = nil ) {
    _onComplete = onComplete
    super.init()
    
    addCondition(AuthorizedFor(Capability.Location(.WhenInUse)))
    addCondition(MutuallyExclusive<RequestUserLocationTrackingOperation>())
  }
  
  override convenience init() {
    self.init(onComplete: nil)
  }
  
  // MARK: Operation
  
    Capability.Location().authorizationStatus { (status) in
      switch (status) {
      case .AuthorizedWhenInUse, .AuthorizedAlways:
        self.finish()
      case .Restricted, .Denied:
        self.finish(OperationError.ConditionFailed)
      default:
        // no-op
        print("DEFAULT")
      }
    }
  }
  
  public override func finished(errors: [ErrorType]) {
    super.finished(errors)
    dispatch_async(dispatch_get_main_queue()) {
      self._onComplete?(errors.count == 0);
    }
  }
}
```

Here is a run with logging for the buggy case:

```
Unnamed Operation:  >>> MutuallyExclusive<RequestUserLocationTrackingOperation>
Authorize Location.WhenInUse:  >>> MutuallyExclusive<LocationCapability>
Authorize Location.WhenInUse: Initialized -> Pending
Authorize Location.WhenInUse: Pending -> EvaluatingConditions
Unnamed Operation: Initialized -> Pending
Authorize Location.WhenInUse: EvaluatingConditions -> Ready
Authorize Location.WhenInUse: Ready -> Executing
Authorize Location.WhenInUse: Will Execute
Authorize Location.WhenInUse: Executing -> Finishing
Authorize Location.WhenInUse: Finishing with no errors.
Authorize Location.WhenInUse: Finishing -> Finished
Unnamed Operation: Pending -> EvaluatingConditions
Authorize Location.WhenInUse:  <<< MutuallyExclusive<LocationCapability>
Unnamed Operation: EvaluatingConditions -> Ready
Unnamed Operation: Ready -> Finishing
Unnamed Operation: Finishing with errors: [Operations.CapabilityError<Operations.LocationCapability>.AuthorizationNotGranted(__C.CLAuthorizationStatus, Optional(Operations.LocationUsage.WhenInUse))].
Unnamed Operation: Finishing -> Finished
Unnamed Operation:  <<< MutuallyExclusive<RequestUserLocationTrackingOperation>
```

And one with the fix:

```
Unnamed Operation:  >>> MutuallyExclusive<RequestUserLocationTrackingOperation>
Authorize Location.WhenInUse:  >>> MutuallyExclusive<LocationCapability>
Authorize Location.WhenInUse: Initialized -> Pending
Authorize Location.WhenInUse: Pending -> EvaluatingConditions
Unnamed Operation: Initialized -> Pending
Authorize Location.WhenInUse: EvaluatingConditions -> Ready
Authorize Location.WhenInUse: Ready -> Executing
Authorize Location.WhenInUse: Will Execute
Authorize Location.WhenInUse: Executing -> Finishing
Authorize Location.WhenInUse: Finishing with no errors.
Authorize Location.WhenInUse: Finishing -> Finished
Unnamed Operation: Pending -> EvaluatingConditions
Authorize Location.WhenInUse:  <<< MutuallyExclusive<LocationCapability>
Unnamed Operation: EvaluatingConditions -> Ready
Unnamed Operation: Ready -> Executing
Unnamed Operation: Will Execute
Unnamed Operation: Executing -> Finishing
Unnamed Operation: Finishing with no errors.
Unnamed Operation: Finishing -> Finished
Unnamed Operation:  <<< MutuallyExclusive<RequestUserLocationTrackingOperation>
```

danthorpe:
And this is using v2.10?

J-Swift:
This is based on me cherry-picking my PR onto master at the moment, then commenting/uncommenting the tweak:

https://github.com/J-Swift/Operations/tree/master

danthorpe:
And, can you clarify - how are you responding to the dialog? It looks like you're denying permission?

J-Swift:
In the 'fixed' log I'm allowing permission.

In the 'buggy' log I'm not responding, that is the condition failing prematurely which causes the operation to finish without executing. The dialog is still on screen at this point. If I then allow/deny nothing else gets logged regardless of my choice (but the system does correctly store that choice as evidenced by future retrieval of the permission).

danthorpe:
Okay, so sorry about this - but I don't quite follow what the issue is - how is it not working? 

If you turn off the device or quite the app before responding, is it not unreasonable that the operation finishes - more than likely because the `NSOperationQueue` is deallocated when the app is quit - and it'll cancel all the operations.

The dialog is out of process - if you quit the app while the dialog is open, and then respond, the OS remembers the authorisation permission - and it gets retrieved correctly. Which seems reasonable...

What am I missing?

danthorpe:
Perhaps you can explain what you're trying to do too? I think maybe there is some mis-understanding about what `AuthorizedFor` does/is for. Given the `execute` method of your operation - are you just trying to check the `CLLocationAuthorizationStatus`? Or do you want to execute an `Operation` *only* if authorised?

J-Swift:
The usecase is that I want to allow enabling location tracking as a setting in a settings page.

So, to do that I wrap the AuthorizedFor capability as a condition of an operation. The intended flow is then, run the condition and check in our own `execute` block what the result was so that we can callback to our settings UI to let them know the result.

The issue arises due to a situation where the condition will prematurely fail due to the system calling it back with .NotDetermined immediately upon being asked to requestAuthorization. We correctly handle the 'happy' path of this being returned as the [current system status](https://github.com/danthorpe/Operations/blob/f7b021f04d2f633d4b656fd8059e76bdee915748/Sources/Features/iOS/LocationCapability.swift#L170) but the issue is that this is then also returned immediately upon [calling `requestAuthorization`](https://github.com/danthorpe/Operations/blob/f7b021f04d2f633d4b656fd8059e76bdee915748/Sources/Features/iOS/LocationCapability.swift#L173). This is due to the delegate method not being prepared for .NotDetermined as a possible status code when [being called](https://github.com/danthorpe/Operations/blob/f7b021f04d2f633d4b656fd8059e76bdee915748/Sources/Features/iOS/LocationCapability.swift#L186).

So, while we expect the following to happen:

1. Check status
1. See that we are .NotDetermined
1. Request authorization
1. (system presents auth dialog)
1. (user answers dialog)
1. Get called back with either .Denied or .Authorized
1. Check if the callback status means we were granted permission. If .Denied then we were denied and if .Authorized then we were granted

We are actually having this happen:

1. Check status
1. See that we are .NotDetermined
1. Request authorization
1. **Get called back immediately with .NotDetermined**
1. **Interpret .NotDetermined as being denied permission**

Meanwhile during the last two steps, the following is also occurring after we have completed our AuthorizedFor capability check:

1. (system presents auth dialog)
1. (user answers dialog)

Does this make sense? To be clear, I think this is a bug with CLLocationManager, but it is something that I'm trying to work around by being prepared for it.

danthorpe:
Okay - I think I can see what is going on. When a condition is attached to an operation, it is either satisfied, or it fails with an error. An operation can have multiple conditions. If any fail, the operation does not execute - it finishes immediately with an error.

In other words, the operation will only run if all its conditions are satisfied.

If you want to run an operation, and respond to the `CLLocationAuthorizationStatus` attaching a `Capability.Location` is not the way to do it - as you'll never know if the user denied access. The condition will fail if authorisation is denied.

Instead, use `GetAuthorizationStatus` and `Authorize` which are two `Operation` subclasses which can be used to query for the authorization status, and explicitly request authorization of any capability.

So, you can use them like this:

```swift
let getStatus = GetAuthorizationStatus(Capability.Location()) { enabled, status in
    // setup UI to reflect the status - consider writing this as a function.
}
queue.addOperation(getStatus)
```

To explicitly request authorization:

```swift
let authorise = Authorise(Capability.Location()) { enabled, status in
    // setup UI to reflect the status - consider writing this as a function.
    // Notice that the closure signature is the same as GetAuthorizationStatus
    // This is so that the same function can be used.
}
queue.addOperation(authorize)
```

You can use combinations of `SilentCondition` and `NegatedCondition` in order to show or hide different UI elements depending on the authorisation status.

I strong recommend you look at the [Permissions.app](https://github.com/danthorpe/Examples/blob/development/Operations/Permissions/Permissions/LocationViewController.swift#L76) sample code to see how this is done.

Perhaps also read the [programming guide on Capabilities](https://operations.readme.io/docs/conditions) if you haven't had a chance yet.

J-Swift:
Dan, I really appreciate you taking the time on this. I don't think you quite understand my point I'm trying to illustrate however. My use case, while maybe being a slightly wonky use of the available tools you provide, is tangential to the buggy CLLocationManager behavior.

I've gone ahead and recreated the issue on the Permissions.app to hopefully give a bit more clarity:

https://www.dropbox.com/s/xzdxzrtuiu5uhwr/cllocation_issues.mov?dl=0

Note a few things in this. First, the initial run shows that the operation is waiting for our input while the alert is on screen. Second, the next run immediately finishes the operation without waiting for our input. Third, no further operation actions are logged after allowing the permission because the operation has already completed at this point.

Now, in this specific setup that you have with the Permissions.app it turns out that the premature failure doesn't actually have any side-effects that might affect the experience. However, you might imagine a situation where mutual exclusivity guarantees are violated by the operation finishing but still being 'in flight', or in my case the UI callback is structured in a way that interprets the callback as being 'denied' permission even though no permission has actually been given at the point of callback.

danthorpe:
Hi @J-Swift - I guess I just don't really see what the issue is. I watched the video - don't see what the issue is. You're gonna have to explain it better. Sorry! 🤔

J-Swift:
Watch the logging while the app is being used.

In the first run through, notice how when the alert is on screen the logging indicates that we are in the 'executing' phase and remains that way.

In the second run through, notice how when the alert is on screen the logging indicates that we _immediately_ go from 'executing' to 'finishing'. The alert is still on screen. We haven't made a choice yet. The operation has already completed and the alert is _still on screen_.

danthorpe:
Hi @J-Swift - sorry for the delay, bit crazy at work.

But yep - been able to reproduce this - got there in the end 😁. Seems like a bit of a undefined behaviour about how `CLLocationManager` reports the status changes to its delegate.

Will merge this PR, but I will look at cherry picking the change into a point release for `v2.10.1`. I also want to go through all the other capabilities which wrap system permissions to check them too.

J-Swift:
Thanks Dan, and thanks for the library. I tried PSOperations for a brief bit but converted to your version after seeing how much time you've put into your documentation and examples. That has made a big difference in getting things going, so thanks for putting all your time into it.

PR #312, opened 2016-05-10
WebpageOperation support for entersReaderIfAvailable

blg-andreasbraun:
Extended WebpageOperation with the ability to configure entersReaderIfAvailable at initialization.

It is possible now to show an SFSafariViewController with entersReaderIfAvailable set to false.


danthorpe:
Merged! @blg-andreasbraun thanks!

PR #315, opened 2016-05-12
Bugfix: Retaincycle in WebpageOperation 

blg-andreasbraun:
I notices that WebpageOperation has an retain cycle in its current implementation.

I changed it to be a GroupOperation, because it actually is even if it only has one UIOperation.
Added a WillExecuteObsersever to setup the delegate property of SFSafariViewController.

### DIFF
--- a/Sources/Features/iOS/WebpageOperation.swift
+++ b/Sources/Features/iOS/WebpageOperation.swift
@@ -10,25 +10,19 @@ import Foundation
 import SafariServices
 
 @available(iOS 9.0, *)
-protocol WebpageController: class {
-    weak var delegate: SFSafariViewControllerDelegate? { get set }
-    init(URL: NSURL, entersReaderIfAvailable: Bool)
-}
+public class WebpageOperation<From: PresentingViewController>: GroupOperation, SFSafariViewControllerDelegate {
 
-@available(iOS 9.0, *)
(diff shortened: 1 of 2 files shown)


danthorpe:
Hi @blg-andreasbraun thanks for this!

I think this bug has been fixed more generally as part of `ComposedOperation<T>` and so perhaps it would make sense to just subclass that?

blg-andreasbraun:
Hi @danthorpe,

you could be right about ComposedOperation<T>, didn't think about that.
I'll have a look into that matter and post a pull request if I can implement it nicely :)
I think I can add some documentation to this class while I'm at it.

blg-andreasbraun:
@danthorpe 

I think I have an implementation that looks good as an ComposedOperation that also passes the existing tests.

I have one suggestion though I'd like to have your opinion on because I don't know if there are guidelines present.

the first parameter "url" is in lowercase notation, hat contradicts the way Cocoa does it, that is "URL".
ls it planned at some stage to have a code review session where someone applies guidelines?

danthorpe:
@blg-andreasbraun - that is a really really great point about the `URL` thing. I think it makes a lot of sense to have a sanity checking and cleaning ticket for the 3.0 release which is what I'm currently working towards. So, I will create a ticket for this, and it can also cover adding in any missing documentation too.

PR #317, opened 2016-05-14
Implemented OpenInSafariOperation

blg-andreasbraun:


blg-andreasbraun:
Hi @danthorpe,

I finally had time to implement and test this operation.
I hope the tests are ok like this, I don't have much experience writing those :')

For some reason the checks are failing but I can't check why.
Can I have some info on that to fix it? I can't see any problem from my site, no conflicts or anything.

danthorpe:
Can you install [SwiftLint](https://github.com/realm/SwiftLint) for me?

It is usually just as easy as

```
$ brew install swiftlint
```

if you use Homebrew.

Then, when you re-build, you will have a bunch of automatic modifications from the linting tool. This is why CI is failing your builds at the moment.

danthorpe:
Can you also check that you haven't added iOS files to the Mac OS X framework target. If in doubt, check that the OS X test pass. CI runs these tests too, and that is failing because Swift Lint is failing, which should not happen if you have only added the files to the iOS framework target.

danthorpe:
Also - sorry, I should have opened with this - thanks so much for this contribution! I haven't looked though the code yet, but its great to have people contribute to open source!

Sources/Features/iOS/OpenInSafariOperation.swift
```diff
@@ -0,0 +1,129 @@
...
+ - If the current iOS version does not support SFSafariViewController, it opens the URL the old school way.
+ */
+
+public class OpenInSafariOperation<From: PresentingViewController>: GroupOperation {
+    /// The URL to open.
+    public var URL: NSURL
+    /// A flag that determines whether the `SFSafariViewController` should open the `URL` in reader mode or not.
+    public var entersReaderIfAvailable: Bool
+    /// The presenting `ViewControllerDisplayStyle`
+    private var displayControllerFrom: ViewControllerDisplayStyle<From>?
```
danthorpe: I think these should be constants `let` yeah?

Sources/Features/iOS/OpenInSafariOperation.swift
```diff
@@ -0,0 +1,129 @@
...
+        UIApplication.sharedApplication().openURL(URL)
+        
+        finish()
+    }
+}
+
+/**
+ An `DecideWhereToOpenOperation` object represents an `Operation` that decides on some circumstances whether it's parent `OpenInSafariOperation` should open a `URL` in the `SFSafariViewController` or in the Safari app.
+ */
+class DecideWhereToOpenOperation: Operation {
```
danthorpe: So, what's this operation for? Can you explain your reasoning a bit here?

blg-andreasbraun:
Hi there!

Thanks for the help, installed Swift Lint and it was a forced downcast it complained about.

Do you mean the DecideWhereToOpenOperation?
It is primarily for the Unit Tests so there can be a test that executes the WebpageOperation and a second one that just opens the URL in Safari (not performing WebpageOperation).

danthorpe:
Okay, so `DecideWhereToOpenOperation` is only to facilitate the tests of `OpenInSafariOperation`. But this means that there is now a new public interface `DecideWhereToOpenOperation` which isn't relevant outside of `OpenInSafariOperation`.  More importantly, this operation is not really tested, and all of this is a little unnecessary, as it's actually pretty easy to test 😃

There are a couple of ways you can do this, both of them involve having an internal variable with a default value. The default value is the real way to perform the functionality, however in the tests, after creating the operation, we replace the default behaviour with the testing behaviour which we need.

This can either be done by defining protocols which expose the behaviour that we want, e.g.

```swift
internal protocol ApplicationIsFullscreenPresentation {
    var isFullscreenPresentation: Bool { get }
}

internal protocol ApplicationOpenURL {
    func openURL(_: NSURL) -> Void
}
```

Then you would just have properties in the operation of something which conforms to this protocol, and make `UIApplication` conform to the protocol in the code, and testable types conform to the protocol in the tests. This means we don't actually test using `UIApplication` which is a good thing. This way is a little bit more setup.

The other option, which I prefer, is just to define a block instead of a protocol as the property. Now, we will probably still need to subclass `GroupOperation` but all of the logic is in the `init` method:

Like this:

```swift
extension UIApplication {

    internal var isFullscreenPresentation: Bool {
        if let window = self.sharedApplication().keyWindow {
            return CGRectEqualToRect(window.frame, window.screen.bounds)
        } else {
            return true
        }
    }
}

public class OpenInSafariOperation<From: PresentingViewController>: GroupOperation {

    public let URL: NSURL
    public var entersReaderIfAvailable: Bool

    internal var shouldOpenInSafariViewController: () -> Bool = { UIApplication.sharedApplication(). isFullscreenPresentation }
    internal var openURL: NSURL -> Void = { UIApplication.sharedApplication.openURL($0) }

    private let displayControllerFrom: ViewControllerDisplayStyle<From>
    private let sender: AnyObject?

    public init(URL: NSURL, displayControllerFrom from: ViewControllerDisplayStyle<From>, entersReaderIfAvailable: Bool = false, sender: AnyObject? = .None) {
        self.URL = URL
        self.entersReaderIfAvailable = entersReaderIfAvailable
        self.displayControllerFrom = from
        self.sender = sender
        super.init(operations: [])
        name = "Open in Safari"
    }

        if shouldOpenInSafariViewController() {
           addChildOperation(WebpageOperation(url: URL, displayControllerFrom: displayControllerFrom, sender: sender))
        }
        else {
            addChildOperation(BlockOperation { [unowned self] in self.openURL(URL) }})
        }
        super.execute()
    }
}
```

Make sense?

danthorpe:
Then if you write the operation as I have above, in the tests you can do something like this:

```swift
func test__when_should_not_open_in_safari_etc_etc() {
    var didOpenURL: NSURL? = .None
    let operation = OpenInSafari() // etc etc
    operation.shouldOpenInSafari = { false }
    operation.openURL = { didOpenURL = $0 }
    // run test etc
    // assert didOpenURL is equal to the expected URL
}
```

blg-andreasbraun:
That sounds like a really good refactoring.

One single thing is that

```
internal var shouldOpenInSafariViewController: () -> Bool = {
   UIApplication.sharedApplication().isFullscreenPresentation
}
```

should also check for the iOS version, because it can only open the Safari View Controller on iOS 9.

Otherwise we would get a repetation of the openURL block in execute.

On the other hand that might not be as confusing as checking the OS version in the block.

That will result in the following execute method

```
   if shouldOpenInSafariViewController() {
      if #available(iOS 9.0, *) {
         addOperation(WebpageOperation(url: URL, displayControllerFrom: displayControllerFrom, sender: sender))
      } else {
         addOperation(BlockOperation { [unowned self] in self.openURL(self.URL) })
      }
   } else {
      addOperation(BlockOperation { [unowned self] in self.openURL(self.URL) })
   }
        
   super.execute()
}
```

It's not nice, but it does emphesize that it opens the URL on older OS Versions.

danthorpe:
Oh, yeah, sure. Well, it's probably nicer to do this:

```swift
    if #available(iOS 9.0, *) && shouldOpenInSafariViewController() {
        addOperation(WebpageOperation(url: URL, displayControllerFrom: displayControllerFrom, sender: sender))
    }
    else {
        addOperation(BlockOperation { [unowned self] in self.openURL(self.URL) })
    }

    super.execute()
}
```

blg-andreasbraun:
Done :D

danthorpe:
Merged! Thanks @blg-andreasbraun - great stuff! 👌😁

blg-andreasbraun:
Thanks to you for the great suggestions and corrections!
It's a great Framework. I'll try to contribute more in the future :)

Issue #288, opened 2016-04-18
Best way to handle CKErrorLimitExceeded in CloudKit modification operations


swiftlyfalling:
According to [Apple's docs](https://developer.apple.com/library/ios/documentation/CloudKit/Reference/CKModifyRecordsOperation_class/index.html#//apple_ref/occ/cl/CKModifyRecordsOperation):

> To ensure the speed of fetching and saving records, the server may reject large operations. When this occurs, a block reports the CKErrorLimitExceeded error. **Your app should handle this error, and refactor the operation into multiple smaller batches.**

(Emphasis mine.)

Such refactoring is complicated by CloudKit features like atomic commits, and is thus very application-specific.

This applies to at least _CKModifyRecordsOperation_ (but could theoretically apply to other operations that take multiple items for input, such as the other _CKModify*_ operations, _CKMarkNotificationsReadOperation_, etc). 

I noticed that you have **BatchedCloudKitOperation** for fetch operations (_CKFetchNotificationChangesOperation_ / _CKFetchRecordChangesOperation_), which is fantastic. But what would be the best way to handle the other side of this - for batched modifications - when the server returns an error of CKErrorLimitExceeded and the client needs to handle refactoring the operation into smaller batches? (Presumably including preserving order.)

danthorpe:
Hi @swiftlyfalling - thanks for the excellent feedback & question.

First off, the `BatchedCloudKitOperation` class only supports usage with `CKFetchNotificationChangesOperation` and `CKFetchRecordChangesOperation`, as these are the only two `CKOperation` classes which have support for receiving `moreComing` and `resultsList`. So... I guess it's for when the client asks for potentially lots of data, and it will receive it in batches. I think I could definitely make the names clearer for this. For `CKModifyRecordsOperation`, where the client is sending potentially large amounts of data to the server, I guess this class won't be relevant currently.

However, we should still be able to handle this error type. It just might be a bit tricky, as CloudKit doesn't give you any indication of what a smaller batch size should be. And it might be a bit fiddly too. So, lets look at `CKModifyRecordsOperation` as an example. Our basic strategy will be to:

1. Add an error handler for this code, inside the error handler, we'll capture (`unowned`) the operation.
2. Bisect the data being sent, into left and right halves. The left is the first half, and we want to send it first.
3. Create a new operation, configured as the current operation. Set the left half of data on it.
4. Implement the configure block (which will be applied when new attempts are made) for the right half.
[...]

PR #294, opened 2016-05-17
Simplistic support for CKLimitExceeded error handling

danthorpe:
See the discussion in #288.


danthorpe:
Hey @swiftlyfalling - this is an initial start at providing built in support for Cloud Kit Limit Exceeded error. We were discussing this previously in #288. 

Since that discussion, I have also merged #304 which made it possible for the operation type to define its own type for the errors, and so enables passing back full information to the error handler (as the `error` parameter).

This PR builds on that by automatically adding a suitable handler for `.LimitExceeded`. This method in turn receives another handler from the user, where it gives the error, log and a suggested response, returning a response. The response is how the remaining modifications should be sent, with a `left` and `right` pair.

Currently, this is supported by any `CKModify*` operation. So, for example, lets assume we have the following operation:

```swift
let operation = CloudKitOperation { CKModifySubscriptionsOperation() }

operation.addCondition(AuthorizedFor(Capability.Cloud()))

// Configure the container & database
operation.container = container
operation.database = container.privateCloudDatabase

operation.subscriptionsToSave = toSave
operation.subscriptionIDsToDelete = toDelete
```

we can add the default error handler for limit exceeded like this:

```swift
operation.setErrorHandlerForLimitExceeded()
```

This will basically figure out what is left to save or delete from the original operation, bisect those into left and right sets. Create a new operation for the left hand side which it will add as a child operation to the group, and set this as a dependency of the right hand side. This should respect the ordering of the arrays. The right hand becomes the next attempt when performing the retry.

Thoughts? I'm sure I'm missing 2nd level retry/error behaviour.

danthorpe:
To perform the bisection manually, you can provide your own handler. The signature is like this:

```swift
operation.setErrorHandlerForLimitExceeded { error, log, suggested in
    // the suggested argument is a tuple like this
    let leftHandSideToSave = suggested.left.toSave
    let leftHandSideToDelete = suggested.left.toDelete
    let rightHandSideToSave = suggested.right.toSave
    let rightHandSideToDelete = suggested.right.toDelete
    return ((leftHandSideToSave, leftHandSideToDelete), (rightHandSideToSave, rightHandSideToDelete))
}
```

I think I'll make it so that the response is optional though, which would allow for not attempting a retry.

swiftlyfalling:
In the event of a bisection, the current implementation would result in the CKOperation's specialized completion block getting called multiple times, correct? (Once for each child operation.)

(For example: the `modifyRecordsCompletionBlock` in CKModifyRecordsOperation)

If so, I would suggest that this is confusing behavior. According to the docs (`modifyRecordsCompletionBlock`):

> This block is executed only once and represents your last chance to process the operation results. It is executed after all individual progress blocks have completed but before the operation’s completion block. 

So it should either be heavily stressed in the documentation that the CKOperation's specialized completion block may be called more than once if you use .setErrorHandlerForLimitExceeded(), or the CloudKitOperation should accumulate:
savedRecords
deletedRecordIDs
operationError*

And pass them up to the user-set modifyRecordsCompletionBlock once. (When the last child operation's modifyRecordsCompletionBlock is called.) 

*operationError must accumulate errors for items. To match the documentation, this would need to be set as CKErrorPartialFailure if some fail but some succeed, and:

> The userInfo dictionary of the error contains a CKPartialErrorsByItemIDKey key whose value is an NSDictionary object. The keys of that dictionary are the IDs of the records that were not saved or deleted, and the corresponding values are error objects containing information about what happened.

swiftlyfalling:
And actually, I just came across evidence that a simple bisection may not be a safe default:

> If you're doing a batch save and you've got a jumble of records in this graph, and they have delete self references between them, CloudKit is going to do an automatic topological sort for you.

> We're going to upload the records in order so that all the records that are referenced will be up there by the time the record reference in them is uploaded.

> This is really great in the public database where you don't have atomic commits but you still want to be able to upload data that looks consistent at any point when a client downloads it.

Source: WWDC 2014 Session 231 - Advanced CloudKit ([transcript](http://asciiwwdc.com/2014/sessions/231))

So, according to Apple, CKModifyRecordsOperation does a topological sort in the event of CKReferences, to ensure the CKRecords in a single operation are uploaded in the right order. A simple bisection of the original, unsorted data could break this.

danthorpe:
Hi @swiftlyfalling - really fantastic feedback. 😁

1. Great points about accumulating the results and only sending the specialised completion handler once.

2. Might be okay on this point though? We take the original unsorted items, and we remove any that were actually sent (which we get from the error object), then we bisect the result of that. And then `CKModifyRecordsOperation` does its thing?

swiftlyfalling:
@danthorpe:

1. To clarify on the userInfo[CKPartialErrorsByItemIDKey] dictionary - this is different for each different type of operation. 
(CKModifyRecordsOperation has it as `[CKRecordID : NSError]`, CKModifySubscriptionsOperation has it as "Subscription ID" -> NSError - i.e. `[String : NSError]` - etc.)

1. We're splitting up the operation into multiple operations, and CloudKit can only do its topological sort within a single CKModifyRecordsOperation. If any members of a dependency graph end up on different sides of the split (i.e. in different CKModifyRecordsOperations), we've potentially introduced an error *caused* by the simple split (depending on the order). There's nothing CloudKit can do to fix it - it only has part of the dependency graph in each new operation.

Unfortunately, the only safe way to do this is to handle CKReference dependency graphs, cycles, and topological sorting in the framework itself. And even then, there are edge cases from which the framework can't safely recover. (i.e. If an operation fails with CKErrorLimitExceeded, and the attempt to do a topological sort results in a directed cycle containing all the original records, no smaller attempt can be made - those records cannot be automatically split into sub-operations - and the framework consumer would need to figure out if/how they can split up the records.)

I acknowledge that a simple split works in several simple cases:

- If the CKRecords in recordsToSave contain no CKReferences
- If the CKRecords in recordsToSave contain only CKReferences to records already in CloudKit
- If the order of records in recordsToSave happens to be such that the split places dependencies before the records that contain CKReferences to them, and there are no directed cycles spanning the split.
etc.

But I am concerned that while trying to solve an error case, the simple split is brittle in ways that may not be obvious to framework consumers.

(On the plus side of all of this, the changes made in #304 make it possible to handle this - and other - error cases properly.)

The current side project in which I'm testing this framework + CloudKit doesn't currently utilize CKReferences, so I have a working implementation using the simple split model for that project. Posts on Stackoverflow would seem to indicate that lots of people use CKReferences, though.

danthorpe:
@swiftlyfalling yeah, totally agree with everything you've addd here. The default "simple" split is almost certainly too naive - but it helps get stuff going.

The API here supports an (optional) block argument which has an optional return type for the left and right sides. So, totally possibly for framework consumers to override the default split.

But yes, it does make sense that the framework performs it's own topological sorting of the arguments and divides the remaining items accordingly. Quite a bit still to do here I think.

Thanks for your feedback, and great that #304 is working out well.

FYI - currently in Europe following Radiohead around, so not doing much code on this right now.

leotumwattana:
@danthorpe Thanks for this wonderful framework. I've been trying it out and contemplating switching over from PSOperations. The API is just so much cleaner and more readable.

@swiftlyfalling Thanks for all the work and thinking on this branch. 

Even though the default "simple" split may not work on more complex cases, I think it'll certainly help clean up a lot of boiler plate. For my case, I do use some CKReferences, single directionally, and what I do is make sure I push the referenced items first, then on up the chain.

Anyhow, I've been trying to toy around with this branch, but keep running into the following `crash`:

```
fatal error: 'try!' expression unexpectedly raised an error: Operations.Reachability.Error.FailedToSetDispatchQueue
```

Any ideas? Thanks!

danthorpe:
@leotumwattana thanks! 😍

That error is from deep inside the `Reachability` stuff. I'm not immediately sure what the cause, more than likely trying to set a dispatch queue from a queue which doesn't have a run loop (or something equally weird).

One thing which I am going to do for `CloudKitOperation` before v3 is done is remove the reachability stuff - as it's not really how  says this sort of stuff should be done. Instead, we'll have a default timeout observer.

So, if you are already on a fork - feel free to rip out all that reachability stuff anyway?

leotumwattana:
@danthorpe thanks for the quick reply. 👏 

I've ripped out the `Reachability` stuff. Ran the tests, seems like nothing relating to CloudKit operation is failing.

Now the crash is at adding an operation after the group operation has finished. Any pointers for where I should look? 😄 

`*** -[Operations.OperationQueue addOperation:]: operation is finished and cannot be enqueued`

danthorpe:
@leotumwattana hmm, well that is a more fundamental issue. Sounds like the operation is finishing before the retry handler is properly invoked.

Can you turn `.Verbose` logging on? You can do this by...

```
let operation = CloudKitOperation { CKModifySubscriptionsOperation() } // etc
operation.log.severity = .Verbose
``` 

And then you should see what is going on in terms of lifecycle.

leotumwattana:
Of course. Turned on `.Verbose`. It's as it says. We're trying to add the operation after the `Group Operation: Did finish with errors:`. I'll continue digging into this for a bit. 😉  

```
Group Operation: Aggregated error: ModifyRecordsError<CKRecord, CKRecordID>(underlyingError: <CKError 0x15888ec00: "Limit Exceeded" (27/1020); "Your request contains 1363 items which is more than the maximum number of items in a single request (400)">, saved: Optional([]), deleted: Optional([]))
******** <CKModifyRecordsOperation: 0x158d60d40; operationID=014E9BC5-12F6-45F1-8FD3-E16491FE9262, stateFlags=finished, metrics=<CKOperationMetrics: 0x158df0480; cloudKitMetrics=<CKMetric: 0x1575f0070; startDate=2016-06-01 16:50:40 +0000, duration=0.232, queueing=0.082, executing=0.246, bytesUploaded=0, bytesDownloaded=0, connections=1, connectionsCreated=0>>, requestIDs=(6B8A1C6F-EF20-4416-82F9-AEE5D51C9D39)>
Group Operation: Will finish with 1 errors.
OPRCKOperation<CKModifyRecordsOperation>: Will finish with 1 errors.
PushOperation for ORReminder: Received CloudKit Limit Exceeded error: ModifyRecordsError<CKRecord, CKRecordID>(underlyingError: <CKError 0x15888ec00: "Limit Exceeded" (27/1020); "Your request contains 1363 items which is more than the maximum number of items in a single request (400)">, saved: Optional([]), deleted: Optional([]))
OPRCKOperation<CKModifyRecordsOperation>: Did finish with errors: [Operations.ModifyRecordsError<CKRecord, CKRecordID>(underlyingError: <CKError 0x15888ec00: "Limit Exceeded" (27/1020); "Your request contains 1363 items which is more than the maximum number of items in a single request (400)">, saved: Optional([]), deleted: Optional([]))].
Group Operation: Did finish with errors: [Operations.ModifyRecordsError<CKRecord, CKRecordID>(underlyingError: <CKError 0x15888ec00: "Limit Exceeded" (27/1020); "Your request contains 1363 items which is more than the maximum number of items in a single request (400)">, saved: Optional([]), deleted: Optional([]))].
*** -[Operations.OperationQueue addOperation:]: operation is finished and cannot be enqueued
```

danthorpe:
@leotumwattana - hmm, interesting - I didn't know that there was actually an internal limit of 400.

Well, one of the reasons why this isn't merged yet, is because I've not finished all the unit testing, and I want to trigger it in an app against CloudKit - so, I guess it's still not quite right. If you find any more info, please feedback to me. I'll hopefully get it finished off soon - but can't promise that it'll be that soon.

Also, I have a Slack team for discussion, DM me on Twitter your email I can invite you, and we can have a more free flowing discussion.

leotumwattana:
@danthorpe As discussed, here's a sample project: [OperationCloudKit](https://github.com/leotumwattana/OperationCloudKit) that will crash with `[Operations.OperationQueue addOperation:]: operation is finished and cannot be enqueued`

Thanks!

leotumwattana:
@danthorpe I've updated the sample project. I think right now any retry of any CloudKitOperation will cause the same `operation is finished and cannot be enqueued` crash. 

You mentioned something about a possible regression from changes to `Will/DidFinishObservers`? Which commit(s) might you be referring to?

danthorpe:
@leotumwattana well, we've had `willFinish` and `didFinish` for a while. But I added `willCancel`, and renamed `didStart` to `willExecute` in https://github.com/danthorpe/Operations/pull/293.

I'm going to step through your samples now. One question that I do have is why you've added the `CloudKitOperation` into another `GroupOperation`? What is the logic/reason for this?

leotumwattana:
@danthorpe Adding the `CloudKitOperation` into another `GroupOperation` is probably not necessary for the sample project. I was just modelling it based on the more complex operation chains from my other projects. Such as doing some other operation after the `CloudKitOperation` is done.

In the case of a `CloudKitOperation<CKModifyRecordsOperation>` I might push certain `CKReferenced` records up first, then push other records up to avoid issues as discussed with @swiftlyfalling.

`GroupOperation` seems like a good way to package up such higher level operations, but I'm all ears if you have a better suggestion?

swiftlyfalling:
I can reproduce this crash with the sample project, with both this branch and the development branch.

I can also reproduce it if I simplify CreateRecordOperation's setErrorHandlerForCode(.LimitExceeded) to:
```
createRecordOperation.setErrorHandlerForCode(.LimitExceeded) { (error, log, suggested) -> (delay: Delay?, configure: OPRCKOperation<CKModifyRecordsOperation> -> Void)? in
    //return (Delay.By(1), suggested.configure)   // also fails
    return (suggested.delay, suggested.configure)
}
```

swiftlyfalling:
@leotumwattana:

This code in your OperationCloudKit example seems like it might be part of the problem:

```
let op = CKModifyRecordsOperation(
            recordsToSave: records,
            recordIDsToDelete: nil)
        
        op.savePolicy = .ChangedKeys
        op.qualityOfService = .UserInitiated
        
        let createRecordOperation = CloudKitOperation { op }
        createRecordOperation.name = "CreateRecordCloudKitOperation"
        createRecordOperation.container = CloudKit.container
        createRecordOperation.database = CloudKit.database
```

If I replace it with:

```
let createRecordOperation = CloudKitOperation { CKModifyRecordsOperation() }
        createRecordOperation.name = "CreateRecordCloudKitOperation"
        createRecordOperation.container = CloudKit.container
        createRecordOperation.database = CloudKit.database
        createRecordOperation.recordsToSave = records
        createRecordOperation.recordIDsToDelete = nil
        createRecordOperation.savePolicy = .ChangedKeys
        createRecordOperation.qualityOfService = .UserInitiated
```

I can't reproduce the crash.

danthorpe:
@swiftlyfalling @leotumwattana of course! Sorry, I should have picked up on this earlier.

The closure `{ op }` is going to try to re-run the same operation, but `NSOperation` is a "one shot" class. Once it is run, it cannot be run again. This is why `RepeatedOperation` (superclass of `CloudKitOperation`) takes a generator of an operation.

So, the replacement from @swiftlyfalling is the "nicest" way to write it. But you could also do this:

```swift
let createRecordOperation = CloudKitOperation {
    let op = CKModifyRecordsOperation(recordsToSave: records, recordIDsToDelete: .None)
    op.savePolicy = .ChangedKeys
    op.qualityOfService = .UserInitiated
    return op
 }

createRecordOperation.name = "CreateRecordCloudKitOperation"
createRecordOperation.container = CloudKit.container
createRecordOperation.database = CloudKit.database

queue.addOperation(createRecordOperation)
```

and it should work okay.

I will definitely update the documentation/wiki to make this clear, as it's such an easy mistake to make!

Sources/Features/Shared/CloudKit/CloudKitError.swift
```diff
@@ -38,3 +52,148 @@ public struct CloudKitError: CloudKitErrorType {
...
+internal extension BatchModifyOperationType where Save: Equatable, Save == Error.Save, Delete: Equatable, Delete == Error.Delete {
+
+    typealias ToModify = (toSave: [Save]?, toDelete: [Delete]?)
+    typealias ToModifyResponse = (left: ToModify, right: ToModify)
+
+    func bisect(error: Error) -> ToModifyResponse {
+        var left: ToModify = (.None, .None)
+        var right: ToModify = (.None, .None)
+
+        let remainingToSave = toSave?.filter { error.saved?.contains($0) ?? false }
```
swiftlyfalling: Condition should be something like: 
`let remainingToSave = toSave?.filter { !(error.saved?.contains($0) ?? false) }`

Also, just using the `contains` method on `error.saved` won't work properly. In the event that `saved` is an array of CKRecords (as in CKModifyRecordsOperation), what ought to be compared is the CKRecord.recordIDs. The saved CKRecords will _always_ be different from the initial CKRecords (as CloudKit returns the records-as-saved, which includes the updated system fields).

Thus, some ability to pass in a comparator function is needed?

Sources/Features/Shared/CloudKit/CloudKitError.swift
```diff
@@ -38,3 +52,148 @@ public struct CloudKitError: CloudKitErrorType {
...
+        let remainingToDelete = toDelete?.filter { error.deleted?.contains($0) ?? false }
```
swiftlyfalling: Condition should be something like: 
`let remainingToDelete = toDelete?.filter { !(error.deleted?.contains($0) ?? false) }`

swiftlyfalling:
On the topic of this PR:

I tested it with CloudKit, and some funky things are going on.

1. bisect() wasn't working (see code comments above on the filter conditional, but there are other problems depending on the type of `Save`)

2. Even with bisect() fixed, it executes endlessly with the error.

I tried with CKModifyOperation with 500 recordsToSave (just like the OperationCloudKit example code). It fails with CKErrorLimitExceeded as expected, bisects (with the fixes) to 250/250, sets lhs.toSave, and rhs.toSave.... 

and then the left operation succeeds, and the right operation fails trying to save the original 500 records. (Or so it seems?)

danthorpe:
Thanks so much for all the testing and support on doing this! I will get back to it tomorrow (need to write this talk over the weekend)...

I suspect what is happening is that the configure block is not getting applied correctly.

swiftlyfalling:
Actually, it looks like the configure block **is** being applied correctly. The settings are just being overwritten by the original configure block in RepeatedOperation.swift : line 278:
```
public func addNextOperation(@autoclosure shouldAddNext: () -> Bool = true) {
    if let (delay, op) = next() {
        if shouldAddNext() {
👉           configure(op)
```

The execution flow seems to be as follows (starting at the return of recoverWithInfo):

CloudKitOperation.Swift:
```
let handler: Handler = { [weak _recovery] info, payload in
➡️  guard let _recovery = _recovery, (delay, configure) = _recovery.recoverWithInfo(info, payload: payload) else { return .None } // this works, and returns the proper configure block
    let (_, operation) = payload
✅  configure(operation) // values of operation are properly set by this configure block (retrieved ultimately from the errorLimitExceeded handler)
    return (delay, operation) ⤵️
}
```
RetryOperation.swift:
```
class RetryGenerator<T: NSOperation>: GeneratorType {
    typealias Payload = RetryOperation<T>.Payload
    typealias Handler = RetryOperation<T>.Handler

    internal let retry: Handler
    internal var info: RetryFailureInfo<T>? = .None
    private var generator: AnyGenerator<(Delay?, T)>

    init(generator: AnyGenerator<Payload>, retry: Handler) {
        self.generator = generator
        self.retry = retry
    }

    func next() -> Payload? {
        guard let payload = generator.next() else { return nil }
        guard let info = info else { return payload }
➡️        return retry(info, payload) ⤵️
    }
}
```
RepeatedOperation.swift:
```
public func addNextOperation(@autoclosure shouldAddNext: () -> Bool = true) {
➡️    if let (delay, op) = next() {
        if shouldAddNext() {
💥            configure(op)
```

It occurs to me that it might be useful for the RepeatedOperation's generator to be able to return that it's already configured the next operation. (And that because it doesn't (?) this might affect other cases?)

leotumwattana:
@swiftlyfalling Ah, nice catch 👏 and my bad 😅. I updated the sample project with the corrected code.

swiftlyfalling:
@leotumwattana:
See @danthorpe's related message in Issue #326 :

> Just a quick update on this - after further discussions in the CloudKitOperation work that is being done at the moment. So, will definitely make some changes so that GroupOperation subclasses can indicate whether the error was handled successfully or not.

> Haven't figured out the details, will evolve it further in a PR.

(Simply erasing the error may not be ideal.)

And just to clarify:
Initializing a `CreateRecordCloudKitOperation` with this code:
```
let createRecord = CreateRecordOperation()
queue.addOperations([createRecord])
```
ultimately finishes successfully, albeit with an error of CKErrorLimitExceeded (despite the child CloudKitOperation having handled it properly internally).

The endless retrying comes into play with the `RetryOperation<CreateRecordCloudKitOperation>`.

swiftlyfalling:
As far as I can tell, after some additional testing, what's left to get this all working is:

- [ ] "Make some changes so that GroupOperation subclasses can indicate whether the error was handled successfully or not" 
> So that CloudKitOperation error handlers can optionally mark errors as handled.
> _Random musings:_
> Maybe a separate handledErrors array/accessor so that subclasses (or, say, CloudKitOperation error handlers) can review what's already been "handled", in case it matters?
> Certainly, RetryFailureInfo.aggregateErrors should also contain information on non-handled / "handled" errors?
> Presumably, handled errors shouldn't be considered in the places that normally check for errors to determine success/failure and such, but should still be accessible.

- [ ] Fix RepeatedOperation [executing the original configure block on the possibly-already-configured operation](https://github.com/danthorpe/Operations/pull/294#issuecomment-223723818).
> (What level should this be fixed at? RepeatedOperation / RetryOperation / CloudKitOperation? Does this affect other things?)

- [ ] Add comparator functions / blocks / etc. [for T.Save / T.Delete ? ](https://github.com/danthorpe/Operations/pull/294#discussion_r65786397)

- [ ] Integrate fix for bisect() filter conditionals.

Having done some (admittedly horrible and temporary) workarounds to test this, the rest seems to be working great. So, it's close! 😎

danthorpe:
I've made a start with #326, to address the top item here. I'm thinking an `Errors` type in `Operation` instead of an array. And then it can support the concept of "handled" errors.

leotumwattana:
@swiftlyfalling A little off topic, but I gotta ❤️  your comment formatting on Github. 😀 

Noted on #326. Can't wait to swap out my PSOperation code in my project. I have a mostly completed branch waiting to be committed once .LimitExceeded is ready, but no pressure! 😁

danthorpe:
@swiftlyfalling - agreed! 😍 @swiftlyfalling comments!

We will hopefully get all this sorted out in the next few days. I'd be working on it now, but I've got to write this presentation for [CocoaHeads Stockholm](http://www.meetup.com/CocoaHeads-Stockholm/events/231428668/) which is on Tuesday.... 😁

diagrams

leotumwattana:
@danthorpe No worries. Good luck on the presentation. I'm sure you don't need it, but have it anyway! 😉

leotumwattana:
Woot. Looks like things are moving along. Wasn't expecting it during WWDC week. Thanks @danthorpe and @swiftlyfalling.

danthorpe:
I'm actually currently in the iCloud labs - going to go through this with the  CloudKit engineers now...

leotumwattana:
Ah. Nice. I so want to be at WWDC. Have fun!

danthorpe:
To be honest @leotumwattana - the stuff from  has been a little underwhelming, but meeting people has been brilliant! Shame you couldn't be here 😔.

leotumwattana:
Well. Probably hard for Apple to keep the breakneck pace. Let's see if they are saving anything for next year. Fingers crossed.

danthorpe:
@swiftlyfalling @leotumwattana okay, there is still quite a lot to do for a bullet proof `CloudKitOperation` wrapper than can handle `CKLimitExceeded` and similar errors, but I think we've reached a good place here. This branch has been going on a long time, and it offers some significant improvements which support handling these kinds of errors.

Ultimately, there are still the rather tricky issues:
1. Ensuring that the "completion block" set on the original `CloudKitOperation` is not executed multiple times when bisecting the result.
2. Sorting the object/reference graph correctly so that bisecting does not save incomplete graphs

Recommendation from Apple is not even attempt to handle this sort of thing automatically, and it should be left to the application developer, as they will *always* understand their object graph more accurately then this framework can. The best that we can do, is to provide convenience hooks.

So - with that, going to merge this now and will continue look into point 1 above after v3.0.

PR #321, opened 2016-05-25
CLLocationManager runloop issues

danthorpe:
Me again with another CLLocationManager bug/issue/headache.

I am triggering a LocationCapability authorization from within a group operation like this:

```
let op = Authorize.init(Capability.Location(.WhenInUse), completion: weakSelf.cb)
weakSelf.addOperations(op)
```

This is basically taken from your Examples repo project. The issue I run into is that this correctly _asks_ for permissions, but the Capability delegate never gets the callback. I've tracked it down to being an issue with the thread the manager is created on, and the fact that it doesn't have a runloop. I came to this conclusion after reading [this helpful answer](http://stackoverflow.com/a/19041925/1273175) and then verifying [in the documentation](https://developer.apple.com/library/ios/documentation/CoreLocation/Reference/CLLocationManagerDelegate_Protocol/index.html).

The interesting bit is this

> The methods of your delegate object are called from the thread in which you started the corresponding location services. That thread must itself have an active run loop, like the one found in your application’s main thread.

Is this something you are aware of/familiar with working around from an `NSOperationQueue`? If so, what is the suggested approach? I've simply done the following hack:

```
weakSelf.queue.suspended = true
NSOperationQueue.mainQueue().addOperationWithBlock({ 
  let op = Authorize.init(Capability.Location(.WhenInUse), completion: weakSelf.cb)
  weakSelf.addOperations(op)
  weakSelf.queue.suspended = false
})
```

The `suspended` nonsense is to make sure the `GroupOperation` stays around until I am done dispatching the block. It just feels like the wrong answer, and I would rather the `LocationCapability` be smart enough to handle this on it's own by ensuring the `CLLocationManager` is configured on a thread which complies with the documentation.

Thanks for any suggestions you might have, or any insight into how I might be doing things wrong!

### DIFF
--- a/Operations.xcodeproj/project.pbxproj
+++ b/Operations.xcodeproj/project.pbxproj
@@ -266,6 +266,10 @@
 		659A55101CD328CC00D9D311 /* CloudKitInterface.swift in Sources */ = {isa = PBXBuildFile; fileRef = 659A54EC1CD3285000D9D311 /* CloudKitInterface.swift */; };
 		659A55111CD328CC00D9D311 /* CloudKitOperation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 659A54ED1CD3285000D9D311 /* CloudKitOperation.swift */; };
 		659A55121CD328CC00D9D311 /* CloudKitOperationExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 659A54EE1CD3285000D9D311 /* CloudKitOperationExtensions.swift */; };
+		65A592991D1C28E5001F9229 /* Queue.swift in Sources */ = {isa = PBXBuildFile; fileRef = 65A592981D1C28E5001F9229 /* Queue.swift */; };
+		65A5929A1D1C28E5001F9229 /* Queue.swift in Sources */ = {isa = PBXBuildFile; fileRef = 65A592981D1C28E5001F9229 /* Queue.swift */; };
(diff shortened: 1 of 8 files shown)


danthorpe:
@J-Swift - sorry this took so long. But it's now fixed. To verify it... I've used it in a modified version of my Permissions example app, with the following code:

```swift
let getLocation = UserLocationOperation { location in
    self.state = .Completed
    self.location = location
}
let group = GroupOperation(operations: getLocation)
group.log.severity = .Verbose
queue.addOperation(group)
```

This produces output in the log like this:

> Group Operation: Initialized -> Pending
> Group Operation: Pending -> Executing
> Group Operation: Will Execute
> User Location: >>> MutuallyExclusive<CLLocationManager>
> User Location: Initialized -> Pending
> User Location: Pending -> Executing
> User Location: Will Execute
> User Location: Updated last location: <+51.30952903,-0.12569460> +/- 65.00m (speed -1.00 mps / course -1.00) @ 23/06/2016, 17:32:29 British Summer Time
> User Location: Executing -> Finishing
> User Location: Will finish with no errors.
> User Location: Finishing -> Finished
> User Location: Did finish with no errors.
> Group Operation: Executing -> Finishing
> Group Operation: Will finish with no errors.
> User Location: <<< MutuallyExclusive<CLLocationManager>
> Group Operation: Finishing -> Finished
> Group Operation: Did finish with no errors.

Which, I hope shows that here we're running location operations inside other operations on their own dedicated queue. The bug here is ultimately solved by ensuring that the `CLLocationManager()` is always created on the main queue (which has a run loop 😀).

In terms of adopting this at the moment - it's going to be on `development` and released with v3 (hopefully very soon!)

PR #330, opened 2016-06-03
Override NSOperationQueue's mainQueue() to return the main queue as an OperationQueue

gtchance:


danthorpe:
Hi @gtchance - thanks for this!

Just one thought, the [documentation](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/NSOperationQueue_class/index.html#//apple_ref/occ/clm/NSOperationQueue/mainQueue) states: "The returned queue executes one operation at a time on the app’s main thread." I'm thinking that we should also set `maxConcurrentOperationCount` in addition to the underlying queue.

gtchance:
@danthorpe, nice catch! I made the adjustment.

danthorpe:
@gtchance great! Thanks again!

PR #326, opened 2016-06-04
Need to suppress error in group operation

danthorpe:
I have a group operation with a child operation that does a network request to our API. Sometimes, the request will return a 401, in which case I run an authentication operation, and when that succeeds, run the original operation again. When this succeeds, the group operation should finish and return no error, but it returns the 401. How can I suppress this error since I handled it? I'd like to be able to just remove it from _aggregateErrors, but that's private and there's no removeError() function. Is there a better way to architect this? Or should I fork and add a removeError() function?

Thanks,
Glen

### DIFF
--- a/Sources/Core/Shared/Operation.swift
+++ b/Sources/Core/Shared/Operation.swift
@@ -61,6 +61,35 @@ public class Operation: NSOperation {
         }
     }
 
+    public class Errors {
+        public var all = Array<ErrorType>()
+        public var handled = Array<ErrorType>()
+
+        public var count: Int {
+            return all.count
+        }
+
+        public var isEmpty: Bool {
+            return all.isEmpty
+        }
+
+        public var first: ErrorType? {
+            return all.first
+        }
+
+        var didFinishWithErrors: Bool {
+            return count > 0 && count != handled.count
+        }
+
+        public func addErrors(errors: [ErrorType]) {
+            all += errors
+        }
+
+        public func handledErrors(errors: [ErrorType]) {
+            handled += errors
+        }
+    }
+
     /**
      Type to express the intent of the user in regards to executing an Operation instance
 
@@ -97,7 +126,7 @@ public class Operation: NSOperation {
     private let stateLock = NSLock()
     private lazy var _log: LoggerType = Logger()
     private var _state = State.Initialized
-    private var _internalErrors = [ErrorType]()
+    private var _internalErrors = Protector(Errors())
     private var _hasFinishedAlready = false
     private var _observers = Protector([OperationObserverType]())
 
@@ -115,8 +144,8 @@ public class Operation: NSOperation {
         willSet {
             willChangeValueForKey("Cancelled")
             if !_cancelled && newValue {
-                operationWillCancel(errors)
-                willCancelObservers.forEach { $0.willCancelOperation(self, errors: self.errors) }
+                operationWillCancel(errors.all)
+                willCancelObservers.forEach { $0.willCancelOperation(self, errors: self.errors.all) }
             }
         }
         didSet {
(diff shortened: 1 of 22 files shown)


Sources/Core/Shared/Operation.swift
```diff
@@ -61,6 +61,35 @@ public class Operation: NSOperation {
...
+
+        public func addErrors(errors: [ErrorType]) {
+            all += errors
+        }
+
+        public func handledErrors(errors: [ErrorType]) {
+            handled += errors
+        }
+    }
+
```
swiftlyfalling: Seems to me it would be quite useful to have an `Errors`.`unhandled` accessor too? Thoughts?

swiftlyfalling:
On the commit above:

It seems to me that what would be most useful is to be able to easily get an array of the "unhandled" errors. Thoughts?

danthorpe:
@swiftlyfalling yeah, I agree. But it'll probably have to be a readonly property with an internal API, for it cannot really be determined from `all` and `handled` without jumping through some some serious type erasure and generic constraints.

However, I'll probably add an `internal func addUnhandledErrors(errors: [ErrorType])` which `GroupOperation` would then take advantage of in some way. For example, if a received error is handled, we can add it to the handled array, else it is presumably "unhandled". Still thinking about this. But got to head out now....

If you have any more ideas - please comment 🙂

swiftlyfalling:
Would it be possible to delay aggregating errors until handled status is known?

swiftlyfalling:
Been digging into this a little bit more.

My proposal is: there should be a way for GroupOperation subclasses to "handle" errors, prior to them being "set" (i.e. aggregated). This would open up RetryOperation, for example, to adjust the non-handled / handled errors in its RetryGenerator.

swiftlyfalling:
Or what about the notion of committing error sets.

Group Operation could separately store the set of the "errors from the prior operation" (all as "unhanded"), and then provide some method for subclasses to commit a different set of errors (handled + unhanded) for the prior operation.

If another child operation finishes, and the previous error set hasn't been committed, commit it as-is to aggregateErrors. (Since nothing "handled" it, this should be fine.) Because of the finishingOperation, this will handle any errors from the last "real" child operation too.

swiftlyfalling:
Please take a look at PR #331.

It's just a quick and dirty sample of the change I'm proposing.

It doesn't include any of the logic here for storing handled vs unhandled errors. But what it does do is extend GroupOperation with the logic of uncommitted vs committed errors. This enables RetryOperation to support modification of previous errors. And, thus, fixes the issue with CloudKitOperation and its recovery handlers being able to "handle" previous errors. It also doesn't change existing behavior (minus the changes to RetryOperation's retry handler) - if you don't deal with the uncommitted vs committed errors, things should be the same as they were.

danthorpe:
@gsimmons @jshier @swiftlyfalling - I'm pretty sure that I've resolved this. If anyone gets a chance, could they have a look through the changes?

In summary:

- [x] Refactored `GroupOperation` to have two methods for happy path and error handling.
- [x] API in `GroupOperation` to manage "fatal" errors (i.e. errors which were not handled), and "recovery attempted" errors. With support for removing errors which were successfully recovered from. `GroupOperation` subclasses can use these APIs to manage the group's final errors 
- [x] `RepeatedOperation` now maintains an optional instance of the previous `Operation` instance - this can be used to remove any errors which were retried for it, if the current operation finished with any errors. For `RepeatedOperation` there is no explicit error handling - the operation just gets repeated. 
- [x] In `RetryOperation` we can now use the previous `Operation` instance to perform effective error recovery, by overriding the methods from `GroupOperation.

gsimmons:
It'll be a week or so before I can get back to this, but looking at the API changes, this looks like exactly what I need. Thanks!

PR #331, opened 2016-06-05
Sample: GroupOperation with committed / uncommitted errors

swiftlyfalling:
Allows subclasses to commit changes to prior errors - for example, if
they’ve “handled” them.

**Note:** This is just a quick proof of concept. 

But it does fix this particular issue with CloudKitOperation and the recovery handlers. (_Not
a complete implementation, though - just erases the “handled” errors, instead of committing them as handled._)


danthorpe:
Hi @swiftlyfalling - really appreciate the work and ideas that you put into this. I feel like the solution, now merged with #326 works quite nicely. So, will close this now.

PR #334, opened 2016-06-06
Added missing functionality of UIAlertController to AlertOperation.

blg-andreasbraun:
The AlertOperation now support 'addTextFieldWithConfigurationHandler(_:)'

### DIFF
--- a/Sources/Core/Shared/OperationQueue.swift
+++ b/Sources/Core/Shared/OperationQueue.swift
@@ -189,15 +189,15 @@ public extension NSOperationQueue {
 
 
 public extension OperationQueue {
-    
+
     private static let sharedMainQueue = MainQueue()
 
     /**
      Override NSOperationQueue's mainQueue() to return the main queue as an OperationQueue
-     
+
      - returns: The main queue
      */
-    public override class func mainQueue()->OperationQueue {
+    public override class func mainQueue() -> OperationQueue {
         return sharedMainQueue
     }
 }
--- a/Sources/Core/iOS/AlertOperation.swift
+++ b/Sources/Core/iOS/AlertOperation.swift
@@ -39,27 +39,6 @@ public class AlertOperation<From: PresentingViewController>: Operation {
         return uiOperation.controller
     }
 
-    /// The title of the presented `UIAlertController`.
-    public var title: String? {
-        get {
-            return alert.title
-        }
-        set {
-            alert.title = newValue
-            name = newValue
-        }
-    }
-
-    /// The message body of the presented `UIAlertController`.
-    public var message: String? {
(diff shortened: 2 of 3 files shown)


danthorpe:
Hi @blg-andreasbraun - looks good! 

I'll need to refresh my memory on `AlertController` - but is it possible to add some unit tests here for the new functionality?

blg-andreasbraun:
I don't know if that would be necessary, because the API I added just calls into the UIAlertViewController API.

A coworker is using the framework in a project and missed the ability to add a UITextField to the presented UIAlertController, so I thought it would be good to enable that by just calling the appropriate API on the internal UIAlertController :)

blg-andreasbraun:
Actually I think now that the Operation has the complete functionality of UIAlertController, the internal 'alert' property could actually be annotated as private.

danthorpe:
Yeah, makes sense - this is going to go into `development` and v3 which is a major version change with breaking changes.

I'll wait for you to push the change.

blg-andreasbraun:
Done.

danthorpe:
Very nice! @blg-andreasbraun 💚😀

felix-dumit:
Can we have the `alert` property public again? If using this in an iPad with the alert style `.Sheet` the app will crash:

>
Fatal Exception: NSGenericException Your application has presented a UIAlertController () of style UIAlertControllerStyleActionSheet. The modalPresentationStyle of a UIAlertController with this style is UIModalPresentationPopover. You must provide location information for this popover through the alert controller's popoverPresentationController. You must provide either a sourceView and sourceRect or a barButtonItem. If this information is not known when you present the alert controller, you may provide it in the UIPopoverPresentationControllerDelegate method -prepareForPopoverPresentation.


So we need access to the `alert` property to do something like:

`alertOperation.alert.popoverPresentationController?.barButtonItem = myBarButtonItem`

blg-andreasbraun:
Wouldn't it be easier to expose those properties on the operation itself?

so you would have something more like 

`alertOperation.barButtonItem = myBarButtonItem`

and of cause the other properties like `sourceView` and `sourceRect`.

felix-dumit:
Those properties wouldn't always be available, as they're really properties for the `UIPopoverPresentationController`, which is only present in the `.Sheet` variation.

I feel like we'd be wrapping too deep if we do that, and also what if I want to change some other property of the popover controller? I'd be a lot of work to add all the possible properties to the alert operation. 
It would also add extra dependencies to this library in that if UIKit changes some api we'd also have to reflect it here.

blg-andreasbraun:
OK, on that thought, yeah that could be complicated over time...
I'll post a pull request.

PR #338, opened 2016-06-08
Network observer flickering

### LINKED ISSUE #335: NetworkObserver flickers when called multiple times
I've got a beautiful set of operations which create many more operations, but each of them make `NetworkObserver` an observer of their operation. As far as I can tell, because I'm calling it multiple times, that triggers strange animations.

I've created a GIF from a video that shows the network indicator restarting each time it is called (at least I believe that is what is happening):

It looks like the `NetworkObserver` was (possibly) loosely based on this [SO answer](http://stackoverflow.com/a/6978817/130556) or at least they are using a similar strategy. The main difference between the two is:

[NetworkObserver 66-68](https://github.com/danthorpe/Operations/blob/ecf45018e8e10cbd20b14493b929189cd8b3dd83/Sources/Core/iOS/NetworkObserver.swift#L66-L68):

```swift
if activityCount > 0 {
    networkIndicatorShouldShow(true)
}
```

SO Answer:

```objc
if (activityCount == 0) {
    [self setNetworkActivityIndicatorVisible:YES];
}
```

By doing `> 0` it is actually triggering `networkActivityIndicator.networkActivityIndicatorVisible` multiple times. I think by changing that to `== 0` it will make sure it only gets triggered once and that the timer is actually updated more than just when the timer finishes.

ryanjm:
Added tests for **multiple** operations having network observers.

Fixes #335 

I'm not 100% a fan of this implementation. This biggest change here is simply having tests in place that make sure that `networkActivityIndicatorVisible` doesn't get set multiple times unless it is a change in value.

Please feel free to give pointers on what I can do to improve this. I really appreciated seeing the way these tests were originally setup, nicely done. I'll add some questions in the code itself to get a better understanding.

I could reduce the number of methods or override a setter in the controller which manages changing `networkActivityIndicatorVisible` based on if it is a new value, but I'm not sure that provides any better readability.

@danthorpe mentioned in #335 that reducing the methods might make it harder to test. But considering the tests are just looking at starting/stopping operations and what is or isn't set on `networkActivityIndicator` I'm not sure I understand how it would make it harder to test. The public interface would be kept intact.

Again, any suggestions on how to improve this is appreciated and I can take a crack at making those changes.

--

As a side note it is a little unfortunate that `addCompletionBlockToTestOperation` doesn't actually add a completion block, it adds a `DidFinishObserver`. I wonder if changing the naming of that method / adding another method that does add a completion block would be helpful since we have the following a few times in these tests (which is _pretty close_ to `addCompletionBlockToTestOperation`):

```swift
let expectation = expectationWithDescription("Test: \(#function)")
operation.addCompletionBlock {
    let after = dispatch_time(DISPATCH_TIME_NOW, Int64(1.5 * Double(NSEC_PER_SEC)))
    dispatch_after(after, Queue.Main.queue) {
        expectation.fulfill()
    }
}
```

### DIFF
--- a/Tests/Core/NetworkObserverTests.swift
+++ b/Tests/Core/NetworkObserverTests.swift
@@ -80,4 +80,63 @@ class NetworkObserverTests: OperationTests {
             XCTAssertFalse(self.visibilityChanges[1])
         }
     }
+
+    func test__network_indicator_changes_once_when_multiple_operations_start() {
+
+        let operation1 = TestOperation(delay: 1)
+        operation1.addObserver(NetworkObserver(indicator: indicator))
+
+        let operation2 = TestOperation(delay: 1)
+        operation2.addObserver(NetworkObserver(indicator: indicator))
+
+        addCompletionBlockToTestOperation(operation1, withExpectation: expectationWithDescription("Test: \(#function)"))
+        addCompletionBlockToTestOperation(operation2, withExpectation: expectationWithDescription("Test: \(#function)"))
+
+        runOperations(operation1, operation2)
+
+        waitForExpectationsWithTimeout(3) { error in
+            XCTAssertTrue(operation1.didExecute)
+            XCTAssertTrue(operation1.finished)
+            XCTAssertTrue(operation2.didExecute)
+            XCTAssertTrue(operation2.finished)
+            XCTAssertTrue(self.visibilityChanges[0])
+            XCTAssertEqual(self.visibilityChanges.count, 1)
+        }
+    }
+
+    func test__network_indicator_hides_once_multiple_operations_end() {
+
+        let operation1 = TestOperation(delay: 1)
+        operation1.addObserver(NetworkObserver(indicator: indicator))
+
+        let operation2 = TestOperation(delay: 1)
+        operation2.addObserver(NetworkObserver(indicator: indicator))
+
+        let expectation1 = expectationWithDescription("Test: \(#function)")
+        operation1.addCompletionBlock {
+            let after = dispatch_time(DISPATCH_TIME_NOW, Int64(1.5 * Double(NSEC_PER_SEC)))
+            dispatch_after(after, Queue.Main.queue) {
+                expectation1.fulfill()
+            }
+        }
+        let expectation2 = expectationWithDescription("Test: \(#function)")
+        operation2.addCompletionBlock {
+            let after = dispatch_time(DISPATCH_TIME_NOW, Int64(1.5 * Double(NSEC_PER_SEC)))
+            dispatch_after(after, Queue.Main.queue) {
+                expectation2.fulfill()
+            }
+        }
+
+        runOperations(operation1, operation2)
+
+        waitForExpectationsWithTimeout(3) { error in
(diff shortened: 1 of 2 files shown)


Tests/Core/NetworkObserverTests.swift
```diff
@@ -80,4 +80,63 @@ class NetworkObserverTests: OperationTests {
...
+    func test__network_indicator_changes_once_when_multiple_operations_start() {
+
+        let operation1 = TestOperation(delay: 1)
+        operation1.addObserver(NetworkObserver(indicator: indicator))
+
+        let operation2 = TestOperation(delay: 1)
+        operation2.addObserver(NetworkObserver(indicator: indicator))
+
+        addCompletionBlockToTestOperation(operation1, withExpectation: expectationWithDescription("Test: \(#function)"))
+        addCompletionBlockToTestOperation(operation2, withExpectation: expectationWithDescription("Test: \(#function)"))
```
ryanjm: What is the purpose / need behind having these "completion" blocks (really `DidFinishObserver`)? I don't understand the timing behind when these expectations get fulfilled. The tests test for the operation having finished, but the `visbilityChanges` don't include the changes that would happen with the `NetworkObserver`.

It seems like ideally these tests would be able to properly tests for visibility changes starting _and_ stopping. And you wouldn't need two different tests to test that functionality.

danthorpe:
Hi @ryanjm - sorry I'm not got back to you on this. Just been crazy busy at work before being away for WWDC. Will get to it over the weekend for sure!

ryanjm:
No problem, I'm sure things are busy for you right now.

Tests/Core/NetworkObserverTests.swift
danthorpe: Hi @ryanjm - sorry for the delay getting to this....

The whole `addCompletionBlockToTestOperation...` stuff is just some helper methods in the `OperationTest` base class. And it helps setup the expectation fulfilment for the async testing. However, this has evolved over the course of the project (and I should really clean everything up and unify it all). Now, you can just do something like this:

```
func test__network_indicator_changes_once_when_multiple_operations_start() {

    let operation1 = TestOperation(delay: 1)
    operation1.addObserver(NetworkObserver(indicator: indicator))

    let operation2 = TestOperation(delay: 1)
    operation2.addObserver(NetworkObserver(indicator: indicator))

    waitForOperations(operation1, operation2)

    XCTAssertTrue(operation1.didExecute)
    XCTAssertTrue(operation1.finished)
    XCTAssertTrue(operation2.didExecute)
    XCTAssertTrue(operation2.finished)
    XCTAssertTrue(self.visibilityChanges[0])
    XCTAssertEqual(self.visibilityChanges.count, 1)
}

swiftlyfalling:
@ryanjm: Could you please take a look at #422, which was merged into the `swift/2.2` branch? An additional test exhibited these issues in a new way, and a fix was checked in to help ensure robustness and further eliminate flickering cases.

ryanjm:
All of that looks good to me. That additional test looks very robust and should handle a few additional edge cases.

PR #346, opened 2016-06-18
Swift/3.0

fcobia:
I fixed some problems that were preventing my code from running under Swift 3.0.

1) In Support.swift, I first modified the code for DispatchQueue.sync to remove an endless loop where it called itself. Then I ended up commenting out the code. It seems to work whether it is commented out or not.

2) In ThreadSafety.swift I added the .barrier flag. Otherwise, spawned operations never launch because the observer does not get added or at least there is a race condition.

### DIFF
--- a/Sources/Core/Shared/Support.swift
+++ b/Sources/Core/Shared/Support.swift
@@ -134,11 +134,7 @@ public extension DispatchQueue {
      - parameter block: a throwing block returning T.
      */
     func sync<T>(execute block: () throws -> T) rethrows -> T {
-        var result: T!
-        try sync {
-            result = try block()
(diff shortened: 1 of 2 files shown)


danthorpe:
@fcobia this is great! You are somewhat on the bleeding edge here! I've not finished editing the unit tests yet to get them to compile in Swift 3.0.

And yes, `Queue` is almost certainly going to disappear completely, it is in progress. At the moment, I'm still in SF so - not been able to really sit down and focus on this. Also, I think this is stuck waiting for CI because my agents are not correctly configured yet.

Regardless - will merge into the `swift/3.0` branch...

PR #351, opened 2016-06-28
Exposing the alert property as public for customization.

blg-andreasbraun:
This also fixes a bug where it wasn't possible to set a barButtonItem on an ActionSheet while presenting as a popover.


danthorpe:
@blg-andreasbraun @felix-dumit - thanks for resolving this! 👍 Agree with Felix - maintaining a full wrapper around `UIKit` is not really the goal here.

PR #353, opened 2016-07-02
Fix for OperationQueue.delegate crash (SR-192)

swiftlyfalling:
##### Sometimes visible as:
`EXC_BAD_ACCESS` in `swift_unknownWeakLoadStrong`
`EXC_BAD_ACCESS` in `OSAtomicDequeue`
`EXC_BAD_ACCESS` in `swift_unknownWeakDestroy` [in OperationQueue.deinit]

##### Cause:
Race condition involving the weak `delegate` property of `OperationQueue`, caused by a bug in Swift:
[**(SR-192):** Weak properties are not thread safe when reading](https://bugs.swift.org/browse/SR-192)

Should be fixed in Swift 3.0+ (see https://github.com/apple/swift/pull/1454), but is not fixed in Swift 2.2.1/2.3 (and unlikely to be).

##### Simple, quick repro case:

```swift
class TestDelegate: OperationQueueDelegate {
    func operationQueue(queue: OperationQueue, willAddOperation operation: NSOperation) { /* do nothing */ }
    func operationQueue(queue: OperationQueue, willFinishOperation operation: NSOperation, withErrors errors: [ErrorType]) { /* do nothing */ }
    func operationQueue(queue: OperationQueue, didFinishOperation operation: NSOperation, withErrors errors: [ErrorType]) { /* do nothing */ }
}

for i in 0..<1000000 {
    print(i)
    let testQueue = OperationQueue()
    testQueue.delegate = TestDelegate()
    dispatch_async(dispatch_get_global_queue(0, 0), {
        let _ = testQueue.delegate
    })
    dispatch_async(dispatch_get_global_queue(0, 0), {
        let _ = testQueue.delegate
    })
}
```

##### Reproducible in practice:
_**YES.**_ Will sometimes take millions of GroupOperations until it triggers, but the crash is sporadically reproducible on two Macs here (dozens of times thus far).

##### Solution:
This PR wraps the public `delegate` property with an NSLock. It also sets it to nil first-thing in `deinit` to safely deinitialize it.

It avoids all of this on Swift 3.0+, where it shouldn’t be needed.* 
_*I haven't yet had a chance to test this on Swift 3.0._

##### Notes:

Because the issue can be triggered by concurrent reads of a weak property, the Protector class (Read/Write lock) is not appropriate.

Also adds `test__SR_192_OperationQueue_delegate_weak_var_thread_safety` to `StressTest.swift`.

### DIFF
--- a/Sources/Core/Shared/OperationQueue.swift
+++ b/Sources/Core/Shared/OperationQueue.swift
@@ -49,12 +49,46 @@ is achieved via the overridden functionality of `addOperation`.
 */
 public class OperationQueue: NSOperationQueue {
 
+    #if swift(>=3.0)
+        // (SR-192 is fixed in Swift 3)
+    #else
+    deinit {
+        // Swift < 3 FIX:
+        // (SR-192): Weak properties are not thread safe when reading
+        // https://bugs.swift.org/browse/SR-192
+        //
+        // Cannot surround native deinitialization of _delegate with a lock,
+        // so avoid the issue by setting it to nil here.
+        delegate = nil
+    }
+    #endif
+
     /**
     The queue's delegate, helpful for reporting activity.
 
     - parameter delegate: a weak `OperationQueueDelegate?`
     */
+    #if swift(>=3.0)
     public weak var delegate: OperationQueueDelegate? = .None
+    #else
+    // Swift < 3 FIX:
+    // (SR-192): Weak properties are not thread safe when reading
+    // https://bugs.swift.org/browse/SR-192
+    //
+    // Surround access of delegate with a lock to avoid the issue.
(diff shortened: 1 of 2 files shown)


swiftlyfalling:
The new test should probably be run as part of the iOS tests as well. It looks like the Stress Tests are currently run on MacOS only.

danthorpe:
@swiftlyfalling - firstly - this is the most amazingly awesome pull request I've ever seen! 😬💚

Looking through all of it now...

Will quickly add though, that the stress tests are only on the OS X target as typically, development is done on the iOS target, but CI runs everything of course, so the stress tests get run but don't impact a quick feedback loop.

danthorpe:
@swiftlyfalling - merged!

PR #356, opened 2016-07-06
Updated to Xcode 8 beta 2

fcobia:
I updated the code to compile under Xcode 8 Beta 2. For the most part everything was just accepting the Xcode suggestions. However, I did change the Timer struct to be a class. Not 100% sure this is the best solution, but it at least works.

I have not tried the test yet. We just needed it to compile so we could keep working on our project. We have just started our project, but we do already have a few operations and they seems to be working without any problems.

### DIFF
--- a/Sources/Core/Shared/ComposedOperation.swift
+++ b/Sources/Core/Shared/ComposedOperation.swift
@@ -18,7 +18,7 @@ public class ComposedOperation<T: Operation>: Operation {
     }
 
     init(operation composed: T) {
-        self.target = composed as? Operation ?? GroupOperation(operations: [composed])
+        self.target = composed ?? GroupOperation(operations: [composed])
         self.operation = composed
         super.init()
         name = "Composed Operation"
--- a/Sources/Core/Shared/Condition.swift
+++ b/Sources/Core/Shared/Condition.swift
@@ -166,7 +166,7 @@ public class ComposedCondition<C: Condition>: Condition, AutomaticInjectionOpera
         super.init()
         mutuallyExclusive = condition.mutuallyExclusive
         name = condition.name
-        injectResultFromDependency(condition) { operation, dependency, _ in
+        let _ = injectResultFromDependency(condition) { operation, dependency, _ in
             operation.requirement = dependency.result
         }
     }
--- a/Sources/Core/Shared/OperationQueue.swift
+++ b/Sources/Core/Shared/OperationQueue.swift
@@ -197,7 +197,7 @@ public extension OperationQueue {
 
      - returns: The main queue
      */
-    public override class func main() -> OperationQueue {
+    public class func main() -> OperationQueue {
         return sharedMainQueue
     }
 }
--- a/Sources/Core/iOS/NetworkObserver.swift
+++ b/Sources/Core/iOS/NetworkObserver.swift
@@ -82,19 +82,19 @@ private class NetworkIndicatorController {
     // Public API
 
     func networkActivityDidStart() {
-        assert(Thread.isMainThread(), "Altering network activity indicator state can only be done on the main thread.")
+        assert(Thread.isMainThread, "Altering network activity indicator state can only be done on the main thread.")
         activityCount += 1
         updateIndicatorVisibility()
     }
 
     func networkActivityDidEnd() {
-        assert(Thread.isMainThread(), "Altering network activity indicator state can only be done on the main thread.")
+        assert(Thread.isMainThread, "Altering network activity indicator state can only be done on the main thread.")
         activityCount -= 1
         updateIndicatorVisibility()
     }
 }
 
-private struct Timer {
+private class Timer {
 
     private var isCancelled = false
 
@@ -107,7 +107,7 @@ private struct Timer {
         }
     }
 
-    mutating func cancel() {
+    func cancel() {
         isCancelled = true
     }
 }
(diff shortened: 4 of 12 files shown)


danthorpe:
@fcobia - great! Did you get the unit tests running locally?

I've not even downloaded Xcode 8b2 yet - doing it now on my build machine.

Sources/Core/Shared/ComposedOperation.swift
```diff
@@ -18,7 +18,7 @@ public class ComposedOperation<T: Operation>: Operation {
     }
 
     init(operation composed: T) {
-        self.target = composed as? Operation ?? GroupOperation(operations: [composed])
+        self.target = composed ?? GroupOperation(operations: [composed])
```
danthorpe: @fcobia - this is actually a mistake though - on my original work in the `swift/3.0` branch.

The type constraint should be `Foundation.Operation` and the conditional cast should be `Operations.Operation`.

Sources/Core/Shared/Condition.swift
```diff
@@ -166,7 +166,7 @@ public class ComposedCondition<C: Condition>: Condition, AutomaticInjectionOpera
         super.init()
         mutuallyExclusive = condition.mutuallyExclusive
         name = condition.name
-        injectResultFromDependency(condition) { operation, dependency, _ in
+        let _ = injectResultFromDependency(condition) { operation, dependency, _ in
```
danthorpe: Oh, is this now a warning in b2? Would probably be better to add the appropriate attribute to the method.

Sources/Core/Shared/OperationQueue.swift
```diff
@@ -197,7 +197,7 @@ public extension OperationQueue {
 
      - returns: The main queue
      */
-    public override class func main() -> OperationQueue {
+    public class func main() -> OperationQueue {
```
danthorpe: This might be incorrect? The return type should be `Operations.OperationQueue`. Not sure if that will need an override though.

Sources/Core/iOS/NetworkObserver.swift
```diff
@@ -82,19 +82,19 @@ private class NetworkIndicatorController {
     // Public API
 
     func networkActivityDidStart() {
-        assert(Thread.isMainThread(), "Altering network activity indicator state can only be done on the main thread.")
+        assert(Thread.isMainThread, "Altering network activity indicator state can only be done on the main thread.")
```
danthorpe: I will rebase these changes on-top-of latest `develop` as this has now been improved. We have `Queue.isMainQueue` which doesn't use `Thread` class property anymore.

Sources/Core/Shared/ComposedOperation.swift
fcobia: I thought that was what it was in my last pull request, but it gave me an error saying the cast always succeeds.

Sources/Core/Shared/Condition.swift
fcobia: Probably. It is probably easier to just add to the method itself saying the result is ignorable.

Sources/Core/Shared/OperationQueue.swift
fcobia: I was not sure what to do here. According to the documentation, your original line should be correct. However, the compiler claims that function does not exist in the superclass.

I figured if it has really been removed from the superclass, then only other operations code would call it and it is returning a variable made here anyway, so I thought this would be safe.

Sources/Core/Shared/ComposedOperation.swift
danthorpe: Yeah, sorry - wasn't clear, the `danthorpe:swift/3.0` branch has the error - it's why it's not really ready yet. Basically, this is because the name clash causes Xcode's migrator to produce incorrect, but compiling changes. I showed this to the Xcode/Swift team at WWDC.

I will focus on the Swift 3.0 branch this weekend. Basically I'm going to go through the changes an double check every `Operation` instance to verify if it should be `Foundation.Operation` or `Operations.Operation`.

Sources/Core/Shared/Condition.swift
danthorpe: Yeah 👍.

I know that attribute exists, I just can't recall what it is.... 🤔

Sources/Core/Shared/OperationQueue.swift
danthorpe: I see - so, need to check `Foundation.OperationQueue` to see if this has changed to a static property or just been removed...

Sources/Core/Shared/Condition.swift
blg-andreasbraun: It's @discardableResult

Sources/Core/Shared/OperationQueue.swift
blg-andreasbraun: It's 
```
public extension Operations.OperationQueue {

    private static let sharedMainQueue = MainQueue()

    public override class var main: Operations.OperationQueue {
        Return sharedMainQueue
    }
}
```

now, because main was changed to be a class var instead of a class func.

Sources/Core/iOS/NetworkObserver.swift
blg-andreasbraun: Should we actually bump this to beta 3 now that it's out?
Because there are some additional changes to be done.

jshier: I have a fork that has full beta 3 / Swift 3 compatibility, just trying to get a PR up. Changes are big enough, especially with the rebase off of `develop`, that it won't land easily.

danthorpe: Yep -  I will revert the current `swift/3.0` branch and re-create from `develop` and @jshier can PR to it.

danthorpe:
Sorry - didn't intentionally close this - I just removed the original `swift/3.0` branch to allow @jshier to submit a new PR.

PR #358, opened 2016-07-07
"Operation: Thread Safety"

swiftlyfalling:
This PR aggregates a series of changes made to improve Operation framework thread-safety and behavior.

Problems were identified when trying to cancel GroupOperations. Further testing and examination led to some substantial thread-safety improvements that allow a series of new Stress Tests (also included) to complete without error. (Previously, they would crash, trip asserts, etc.)

Besides thread-safety issues, several other issues were identified and fixed:
- **FIX:** GroupOperation.operations contains the initial operations twice after beginning to execute.
- **FIX:** Extraneous KVO Notifications were being sent for NSOperation keyPaths. KVO notifications are now explicit.
- **FIX:** GroupOperation now only finishes once its child operations have finished.
- **FIX:** GroupOperation now handles cancelling & finishing itself, ensuring that there are not orphaned operations on a queue that never starts if the GroupOperation is cancelled before it executes.
- **FIX:** RepeatedOperation.addNextOperation() did not check if it was cancelled before adding a new child operation. This could lead to an endless loop.
- **FIX:** Logging-related crashes and thread-safety issues.

Numerous new tests that previously exhibited issues, Stress Tests that previously crashed, and a new Operation KVONotification testing framework are included.

(The new Stress Tests also support a `batches` parameter, currently set to 1.)

Additional details on larger fixes below.

### DIFF
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -8,6 +8,8 @@
 
 import Foundation
 
+// swiftlint:disable file_length
+
 /**
 An `Operation` subclass which enables the grouping
 of other operations. Use `GroupOperation`s to associate
@@ -256,7 +258,7 @@ public class GroupOperation: Operation, OperationQueueDelegate {
         if operation !== finishingOperation {
             groupFinishLock.withCriticalScope {
                 assert(!isGroupFinishing, "Cannot add new operations to a group after the group has started to finish.")
-                
+
                 willAddChildOperationObservers.forEach { $0.groupOperation(self, willAddChildOperation: operation) }
 
                 canFinishOperation.addDependency(operation)
@@ -424,9 +426,9 @@ private extension GroupOperation {
      - a CanFinishOperation which manages handling GroupOperation internal state and has every child
        operation as a dependency
      - a finishingOperation, which has the CanFinishOperation as a dependency
-     
-     The purpose of this is to handle the possibility that GroupOperation.addOperation() or 
-     GroupOperation.queue.addOperation() are called right after all current child operations have 
+
+     The purpose of this is to handle the possibility that GroupOperation.addOperation() or
+     GroupOperation.queue.addOperation() are called right after all current child operations have
      completed (i.e. after the CanFinishOperation has been set to ready), but *prior* to being able
      to process that the GroupOperation is finishing (i.e. prior to the CanFinishOperation executing and
      acquiring the GroupOperation.groupFinishLock to set state).
@@ -434,7 +436,7 @@ private extension GroupOperation {
     private class CanFinishOperation: NSOperation {
         private weak var parent: GroupOperation?
         private var _finished = false
-        
+
         init(parentGroupOperation: GroupOperation) {
             self.parent = parentGroupOperation
             super.init()
@@ -470,7 +472,7 @@ private extension GroupOperation {
                         // GroupOperation's new CanFinishOperation.
 
                         let newCanFinishOp = GroupOperation.CanFinishOperation(parentGroupOperation: parent)
-                        
+
                         nonFinishedOperations.forEach { op in
                             newCanFinishOp.addDependency(op)
                         }
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -46,6 +46,7 @@ public class GroupOperation: Operation, OperationQueueDelegate {
     private var canFinishOperation: GroupOperation.CanFinishOperation!
     private var isGroupFinishing = false
     private let groupFinishLock = NSRecursiveLock()
+    private var isAddingOperationsGroup = dispatch_group_create()
 
     /// - returns: the OperationQueue the group runs operations on.
     public let queue = OperationQueue()
@@ -148,33 +149,13 @@ public class GroupOperation: Operation, OperationQueueDelegate {
 
         if additional.count > 0 {
 
-            let addedOperations = groupFinishLock.withCriticalScope { () -> Bool in
+            let shouldAddOperations = groupFinishLock.withCriticalScope { () -> Bool in
                 guard !isGroupFinishing else { return false }
-
-                var handledCancelled = false
-                if self.cancelled {
-                    additional.forEachOperation { $0.cancel() }
-                    handledCancelled = true
-                }
-                let logSeverity = self.log.severity
-                additional.forEachOperation { $0.log.severity = logSeverity }
-
-                self.queue.addOperations(additional)
-
-                if addToOperationsArray {
-                    self._operations.appendContentsOf(additional)
-                }
-
-                if !handledCancelled && self.cancelled {
-                    // It is possible that the cancellation happened before adding the
-                    // additional operations to the operations array.
-                    // Thus, ensure that all additional operations are cancelled.
-                    additional.forEachOperation { if !$0.cancelled { $0.cancel() } }
-                }
+                dispatch_group_enter(isAddingOperationsGroup)
                 return true
             }
 
-            guard addedOperations else {
+            guard shouldAddOperations else {
                 if !finishingOperation.finished {
                     assertionFailure("Cannot add new operations to a group after the group has started to finish.")
                 }
@@ -182,6 +163,31 @@ public class GroupOperation: Operation, OperationQueueDelegate {
                     assertionFailure("Cannot add new operations to a group after the group has completed.")
                 }
                 return
+            }
+
+            var handledCancelled = false
+            if self.cancelled {
+                additional.forEachOperation { $0.cancel() }
+                handledCancelled = true
+            }
+            let logSeverity = self.log.severity
+            additional.forEachOperation { $0.log.severity = logSeverity }
+
+            self.queue.addOperations(additional)
+
+            if addToOperationsArray {
+                self._operations.appendContentsOf(additional)
+            }
+
+            if !handledCancelled && self.cancelled {
+                // It is possible that the cancellation happened before adding the
+                // additional operations to the operations array.
+                // Thus, ensure that all additional operations are cancelled.
+                additional.forEachOperation { if !$0.cancelled { $0.cancel() } }
+            }
+
+            groupFinishLock.withCriticalScope {
+                dispatch_group_leave(isAddingOperationsGroup)
             }
         }
     }
@@ -256,12 +262,23 @@ public class GroupOperation: Operation, OperationQueueDelegate {
         assert(!finishingOperation.finished, "Cannot add new operations to a group after the group has completed.")
 
         if operation !== finishingOperation {
+            let shouldContinue = groupFinishLock.withCriticalScope { () -> Bool in
+                guard !isGroupFinishing else {
+                    assertionFailure("Cannot add new operations to a group after the group has started to finish.")
+                    return false
+                }
+                dispatch_group_enter(isAddingOperationsGroup)
+                return true
+            }
+            
+            guard shouldContinue else { return }
+
+            willAddChildOperationObservers.forEach { $0.groupOperation(self, willAddChildOperation: operation) }
+
+            canFinishOperation.addDependency(operation)
+            
             groupFinishLock.withCriticalScope {
-                assert(!isGroupFinishing, "Cannot add new operations to a group after the group has started to finish.")
-
-                willAddChildOperationObservers.forEach { $0.groupOperation(self, willAddChildOperation: operation) }
-
-                canFinishOperation.addDependency(operation)
+                dispatch_group_leave(isAddingOperationsGroup)
             }
         }
     }
@@ -302,13 +319,25 @@ public class GroupOperation: Operation, OperationQueueDelegate {
         // Ensure that produced operations are added to GroupOperation's
         // internal operations array (and cancelled if appropriate)
 
-        groupFinishLock.withCriticalScope {
+        let shouldContinue = groupFinishLock.withCriticalScope { () -> Bool in
             assert(!finishingOperation.finished, "Cannot produce new operations within a group after the group has completed.")
-            assert(!isGroupFinishing, "Cannot produce new operations within a group after the group has started to finish.")
-            _operations.append(operation)
-            if cancelled {
-                if !operation.cancelled { operation.cancel() }
+            guard !isGroupFinishing else {
+                assertionFailure("Cannot produce new operations within a group after the group has started to finish.")
+                return false
             }
+            dispatch_group_enter(isAddingOperationsGroup)
+            return true
+        }
+
+        guard shouldContinue else { return }
+
+        _operations.append(operation)
+        if cancelled {
+            if !operation.cancelled { operation.cancel() }
+        }
+
+        groupFinishLock.withCriticalScope {
+            dispatch_group_leave(isAddingOperationsGroup)
         }
     }
 }
@@ -455,7 +484,19 @@ private extension GroupOperation {
                 //
                 // Handle an edge case caused by concurrent calls to GroupOperation.addOperations()
 
-                parent.groupFinishLock.withCriticalScope {
+                let isWaiting = parent.groupFinishLock.withCriticalScope { () -> Bool in
+
+                    // Is anything currently adding operations?
+                    guard dispatch_group_wait(parent.isAddingOperationsGroup, DISPATCH_TIME_NOW) == 0 else {
+                        // Operations are actively being added to the group
+                        // Wait for this to complete before proceeding.
+                        dispatch_group_notify(parent.isAddingOperationsGroup,
+                            dispatch_get_global_queue(NSQualityOfServiceToDispatchQOS(self.qualityOfService), 0),
+                            { [unowned self] in
+                            self.execute()
+                        })
+                        return true
+                    }
 
                     // Check whether new operations were added prior to the lock
                     // by checking for child operations that are not finished.
@@ -486,12 +527,23 @@ private extension GroupOperation {
                         // Ensure that no new operations can be added.
                         parent.isGroupFinishing = true
                     }
+                    return false
                 }
+
+                guard !isWaiting else { return }
             }
 
             willChangeValueForKey("isFinished")
             _finished = true
             didChangeValueForKey("isFinished")
+        }
+        private func NSQualityOfServiceToDispatchQOS(qualityOfService: NSQualityOfService) -> qos_class_t {
+            let mapping = [NSQualityOfService.Background: QOS_CLASS_BACKGROUND,
+                           NSQualityOfService.Default: QOS_CLASS_DEFAULT,
+                           NSQualityOfService.UserInitiated: QOS_CLASS_USER_INITIATED,
+                           NSQualityOfService.UserInteractive: QOS_CLASS_USER_INTERACTIVE,
+                           NSQualityOfService.Utility: QOS_CLASS_UTILITY]
+            return mapping[qualityOfService] ?? QOS_CLASS_DEFAULT
         }
         override var executing: Bool {
             return false
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -93,7 +93,7 @@ public class GroupOperation: Operation, OperationQueueDelegate {
      starting the queue, and adding the finishing operation.
     */
     public override func execute() {
-        addOperations(operations.filter { !self.queue.operations.contains($0) })
+        _addOperations(operations.filter { !self.queue.operations.contains($0) }, addToOperationsArray: false)
         queue.addOperation(finishingOperation)
         queue.suspended = false
     }
@@ -122,10 +122,16 @@ public class GroupOperation: Operation, OperationQueueDelegate {
      - parameter operations: an array of `NSOperation` instances.
      */
     public func addOperations(additional: [NSOperation]) {
+        _addOperations(additional, addToOperationsArray: true)
+    }
+
+    private func _addOperations(additional: [NSOperation], addToOperationsArray: Bool = true) {
         if additional.count > 0 {
             additional.forEachOperation { $0.log.severity = log.severity }
             queue.addOperations(additional)
-            operations.appendContentsOf(additional)
+            if addToOperationsArray {
+                operations.appendContentsOf(additional)
+            }
         }
     }
 
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -63,7 +63,9 @@ public class GroupOperation: Operation, OperationQueueDelegate {
     */
     public init(operations ops: [NSOperation]) {
         operations = ops
-        super.init()
+        // GroupOperation handles calling finish() on cancellation once all of its children have cancelled and finished
+        // and its finishingOperation has finished.
+        super.init(disableAutomaticFinishing: true) // Override default Operation finishing behavior
         name = "Group Operation"
         queue.suspended = true
         queue.delegate = self
@@ -126,11 +128,18 @@ public class GroupOperation: Operation, OperationQueueDelegate {
     }
 
     private func _addOperations(additional: [NSOperation], addToOperationsArray: Bool = true) {
-        if additional.count > 0 {
-            additional.forEachOperation { $0.log.severity = log.severity }
-            queue.addOperations(additional)
-            if addToOperationsArray {
-                operations.appendContentsOf(additional)
+        stateLock.withCriticalScope {
+            if additional.count > 0 {
+                assert(!finishingOperation.executing, "Cannot add new operations to a group after the group has started to finish.")
+                assert(!finishingOperation.finished, "Cannot add new operations to a group after the group has completed.")
+                if cancelled {
+                    additional.forEachOperation { $0.cancel() }
+                }
+                additional.forEachOperation { $0.log.severity = log.severity }
+                queue.addOperations(additional)
+                if addToOperationsArray {
+                    operations.appendContentsOf(additional)
+                }
             }
         }
     }
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -113,7 +113,7 @@ public class GroupOperation: Operation, OperationQueueDelegate {
     */
     public override func execute() {
         _addOperations(operations.filter { !self.queue.operations.contains($0) }, addToOperationsArray: false)
-        queue._addCanFinishOperation(canFinishOperation, finishingOperation: finishingOperation)
+        _addCanFinishOperation(canFinishOperation)
         queue.addOperation(finishingOperation)
         queue.suspended = false
     }
@@ -465,12 +465,27 @@ private extension GroupOperation {
     private class CanFinishOperation: NSOperation {
         private weak var parent: GroupOperation?
         private var _finished = false
+        private var _executing = false
 
         init(parentGroupOperation: GroupOperation) {
             self.parent = parentGroupOperation
             super.init()
         }
         override func start() {
+
+            // Override NSOperation.start() because this operation may have to
+            // finish asynchronously (if it has to register to be notified when
+            // operations are no longer being added concurrently).
+            //
+            // Since we override start(), it is important to send NSOperation
+            // isExecuting / isFinished KVO notifications.
+            //
+            // (Otherwise, the operation may not be released, there may be
+            // problems with dependencies, with the queue's handling of
+            // maxConcurrentOperationCount, etc.)
+
+            executing = true
+
             main()
         }
         override func main() {
@@ -490,19 +505,20 @@ private extension GroupOperation {
                     guard dispatch_group_wait(parent.isAddingOperationsGroup, DISPATCH_TIME_NOW) == 0 else {
                         // Operations are actively being added to the group
                         // Wait for this to complete before proceeding.
-                        dispatch_group_notify(parent.isAddingOperationsGroup,
-                            dispatch_get_global_queue(NSQualityOfServiceToDispatchQOS(self.qualityOfService), 0),
-                            { [unowned self] in
-                            self.execute()
-                        })
+                        //
+                        // Register to dispatch a new call to execute() in the future, after the
+                        // wait completes (i.e. after concurrent calls to GroupOperation.addOperations()
+                        // have completed), and return from this call to execute() without finishing
+                        // the operation.
+                        dispatch_group_notify(parent.isAddingOperationsGroup, Queue(qos: qualityOfService).queue, execute)
                         return true
                     }
 
                     // Check whether new operations were added prior to the lock
                     // by checking for child operations that are not finished.
 
-                    let nonFinishedOperations = parent.operations.filter({ !$0.finished })
-                    if !nonFinishedOperations.isEmpty {
+                    let activeOperations = parent.operations.filter({ !$0.finished })
+                    if !activeOperations.isEmpty {
 
                         // Child operations were added after this CanFinishOperation became
                         // ready, but before it executed or before the lock could be acquired.
@@ -514,13 +530,13 @@ private extension GroupOperation {
 
                         let newCanFinishOp = GroupOperation.CanFinishOperation(parentGroupOperation: parent)
 
-                        nonFinishedOperations.forEach { op in
+                        activeOperations.forEach { op in
                             newCanFinishOp.addDependency(op)
                         }
 
                         parent.canFinishOperation = newCanFinishOp
 
-                        parent.queue._addCanFinishOperation(newCanFinishOp, finishingOperation: parent.finishingOperation)
+                        parent._addCanFinishOperation(newCanFinishOp)
                     }
                     else {
                         // There are no additional operations to handle.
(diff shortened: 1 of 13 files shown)


swiftlyfalling:
### FIX: GroupOperation.operations contains the initial operations twice after beginning to execute.
#### Cause: 
`GroupOperation.execute()` calls `GroupOperation.addOperations(operations)`, which then re-adds the operations back to the operations array.

#### Solution: 
Added a private **`_addOperations()`** method to `GroupOperation` with an additional flag to override this behavior - used in `GroupOperation.execute()`. 

`addOperations()` now calls this private implementation method.

#### Test: 
```swift
func test__group_operation__initial_operations_only_added_once_to_operations_array() in GroupOperationTests.swift

**Comment by swiftlyfalling** [comment]
### FIX: Operation.cancel(), Operation.finish() thread-safety.

#### Discussion:
`Operation.cancel()` and `Operation.finish()` operate on and manipulate Operation state, but are not thread-safe.

See: [comments in Issue #352](https://github.com/danthorpe/Operations/issues/352#issuecomment-230068446)

#### Solution:
**Utilize the stateLock.**, but ensure that the lock is *not* held while notifying observers or sending KVO notifications. 

(See the numerous commits.)

#### Tests:
```swift
func test__block_operation_cancel() in StressTests.swift
func test__group_operation_cancel() in StressTests.swift

**Comment by swiftlyfalling** [comment]
### FIX: Operation.main() thread-safety.

#### Discussion:
A race condition is possible in Operation.main() with the following order of operations:

1. The Operation checks `!cancelled` prior to setting `state = .Executing`, and `cancelled == false`.
2. Another thread cancels the operation, and this call to `cancel()` completes in full, **not** calling `finish()` automatically because `Operation.executing` still `== false`.
3. Back in Operation.main(), Operation sets the state to `.Executing`.
4. Operation.main() calls `execute()`.
5. The subclass implementation of `execute()` checks cancelled, and returns without finishing if the Operation is cancelled (which it now is).

In this order of operations, the Operation is never finished.


#### Solution:
**Use of the stateLock.**

#### Tests:
Randomly detectable using:
```swift
func test__block_operation_cancel() in StressTests.swift

**Comment by swiftlyfalling** [comment]
### FIX: RepeatedOperation.cancel() can result in thread-starvation and endless loop.

#### Cause: 
`RepeatedOperation.addNextOperation()` does not check `cancelled` status.

#### Solution:
**Add `guard !cancelled` check at top of `RepeatedOperation.addNextOperation()`**.

#### Test:
```swift
func test__repeated_operation_cancel() in StressTests.swift //(which is also effectively a stress-test for cancelling RepeatedOperations and other thread-safety issues)
```

swiftlyfalling:
### FIX: Crashes exacerbated by non-strict KVO ordering (and spurious notifications?) on mandatory NSOperation keyPaths.

#### Discussion: 
It seems like race conditions may be introduced (or exacerbated) if an NSOperation subclass does not properly send mandatory KVO notifications (for example, for `isFinished`) at the exact right times. A lot of stress testing shows random crashes that go away when enforcing strict ordering and required KVO compliance.

(In #175, it was discovered that overriding .isReady can result in very rare issues.)

> Revision: Some of this may be due to other thread-safety issues, like with the Logger, but converging on NSOperation behavior seems appropriate nonetheless.

#### Solution:

Besides other change sets that enforce thread-safety,
**All KVO notifications for critical NSOperation key paths (`isExecuting`, `isFinished`, `isCancelled`) are now sent explicitly, at the right time**. 

The final `didChangeValueForKey(isFinished)` is sent at the very end of `finish()` (_after_ didFinishObservers are called).

This reduces spurious notifications for unintended key paths, which were happening any time the state changed internally. 
(For example, when internal state transitions from .Initialized -> .Pending, notifications were being generated for `isExecuting` and `isFinished`.)

It converges upon the KVO notification behavior (and frequency) of NSOperation.

It also seems to reduce random crashes of the Stress Tests over many billions of runs.
> This may just be due to sending `isFinished` only once, after the didFinishObservers are executed, but converging on the behavior of NSOperation seems reasonable regardless.

#### Tests:

A test suite was added to verify that KVO notifications are being sent by Operation on the NSOperation keyPaths at the right times and in an appropriate ordering.

```swift
func test__nsoperation_kvo__groupoperation_cancel_from_initialized() in `KVONotificationTests.swift`
func test__nsoperation_kvo__groupoperation_cancelled_to_completion() in `KVONotificationTests.swift`
func test__nsoperation_kvo__groupoperation_child_cancel_from_initialized() in `KVONotificationTests.swift`
func test__nsoperation_kvo__groupoperation_execute_to_completion() in `KVONotificationTests.swift`

func test__nsoperation_kvo__operation_cancel_from_initialized() in `KVONotificationTests.swift`
func test__nsoperation_kvo__operation_cancelled_to_completion() in `KVONotificationTests.swift`
func test__nsoperation_kvo__operation_execute_to_completion() in `KVONotificationTests.swift`
func test__nsoperation_kvo__operation_execute_with_dependencies_to_completion() in `KVONotificationTests.swift`
func test__nsoperation_kvo__operation_state_transition_from_initializing_to_pending() in `KVONotificationTests.swift`
func test__nsoperation_kvo__operation_state_transition_to_executing() in `KVONotificationTests.swift`
func test__nsoperation_kvo__operation_state_transition_to_executing_via_queue() in `KVONotificationTests.swift`
```

swiftlyfalling:
### FIX: GroupOperations can finish() before the .finishingOperation has finished.

#### Symptoms:
A number of race conditions and crashes (easily triggered by cancelling GroupOperations), leaked child Operations, and more.

#### Cause: 
`GroupOperation` relied on the default behavior of `Operation`, which can call `finish()` immediately in response to cancelling in:

1. the `.cancel()` method

2. the `.start()` method

3. the `.main()` method

#### Solution:

**Added support in `Operation` to `disableAutomaticFinishing`**. 
Available in a new initializer: `init(disableAutomaticFinishing: Bool)`

When **`disableAutomaticFinishing`** is set, `Operation` will never call `.finish()` itself, and the subclass is solely responsible for triggering `.finish()` when appropriate (for example, when handling `.cancel()`).

> ℹ️ **NOTES:** 
> - The default behavior of Operation remains the same. 
> - Automatic finishing can only be disabled.
> - Once disabled, it cannot be re-enabled (as this would make little sense and introduces problems). 
> Example: A subclass of GroupOperation, which disables Operation's automatic finishing, cannot re-enable it.

`GroupOperation` **now calls `super.init(disableAutomaticFinishing: true)`**.

This makes the state behavior of `GroupOperation` in response to cancelling much easier to reason about, and much safer. (GroupOperation will now *only* `.finish()` when the `finishingOperation` finishes.)

#### Tests:

```swift
func test__operation_with_disableAutomaticFinishing_doesnt_finish_automatically_when_cancelled() in BasicTests.swift
func test__operation_with_disableAutomaticFinishing_cancelled_before_running_doesnt_finish_automatically_when_started() in BasicTests.swift
func test__operation_with_disableAutomaticFinishing_manual_cancel_and_finish_on_willexecute_does_not_result_in_invalid_state_transition_finished_to_executing() in BasicTests.swift

func test__group_operation__does_not_finish_before_child_operations_have_finished() in GroupOperations.swift
func test__group_operation__does_not_finish_before_child_groupoperations_are_finished() in GroupOperations.swift
func test__group_operation_does_not_finish_before_child_produced_operations_are_finished() in GroupOperations.swift
func test__group_operation__execute_is_called_when_cancelled_before_running() in GroupOperationTests.swift
func test__group_operation_cancellation__queue_is_empty_when_finished() in GroupOperationTests.swift

func test__group_operation__does_not_finish_before_child_operations_are_finished() in StressTests.swift
```

swiftlyfalling:
### FIX: GroupOperation.addOperations thread safety().
#### Cause: 
If `GroupOperation.addOperations()` is called while the operation is being cancelled, it is possible for an operation to be added that does not reflect the Group state (i.e. that is not cancelled, and will never be cancelled).

#### Solution: 
**Thread-safety changes to`GroupOperation._addOperations()`.**

Additionally, a check is done for `.cancelled` status and the new child operations are cancelled if the GroupOperation is already cancelled (but not yet finished).

#### Tests:
```swift
func test__group_operation_cancel_and_add_operation() in StressTests.swift
```

swiftlyfalling:
### FIX: Operation.log (and class _Logger, etc) is not thread-safe.

#### Symptoms:
Crashes in objc_release, etc. With Zombies and Address Sanitizer enabled, the following is produced:
```swift
*** -[CFString release]: message sent to deallocated instance 0x604006632910
```
Running memory history shows (abbreviated):
```swift
Memory allocated at:
    frame #0: <...> libclang_rt.asan_osx_dynamic.dylib`__sanitizer_mz_malloc + 173
    frame #1: <...> libsystem_malloc.dylib`malloc_zone_malloc + 70
    frame #2: <...> CoreFoundation`_CFRuntimeCreateInstance + 300
    frame #3: <...> CoreFoundation`__CFStringCreateImmutableFunnel3 + 2671
    frame #4: <...> CoreFoundation`CFStringCreateWithBytes + 58
    frame #5: <...> Foundation`-[NSPlaceholderString initWithBytes:length:encoding:] + 124
    frame #6: <...> Foundation`+[NSString stringWithUTF8String:] + 135
    **frame #7: <...> Foundation`-[__NSOperationInternal _name] + 41
    frame #8: <...> Operations`NSOperation.operationName.getter(self=<unavailable>) + 44 at Logging.swift:302
    frame #9: <...> Operations`@objc NSOperation.operationName.getter + 33 at Logging.swift:0
    frame #10: <...> Operations`Operation.log.getter(self=<unavailable>) + 109 at Operation.swift:193**
    frame #11: <...> Operations`Operation.(self=<unavailable>, newState=<unavailable>).setter).(closure #1) + 1102 at Operation.swift:414
...
```

#### Cause:

Operation.log(), the _Logger class and LogManager class are not thread-safe.

#### Discussion:

Two immediate problems are:
- `_Logger.operationName` is not thread-safe.
- `Operation._log` is a lazy var [and lazy vars are not thread safe](https://bugs.swift.org/browse/SR-1042).

_Logger.operationName is set on the _Logger object every time that Operation.log is accessed.

A crash can be triggered in many cases - for example, with a stress test of Operations with MutualExclusionConditions (because the ExclusivityManager dispatches, in a separate thread, ExclusivityManager._removeOperation(), which calls `operation.log.verbose("<<< \(category)”)`).

This is the cause behind real-world intermittent crashes in built-in Operations that use Mutual Exclusivity, like `ReverseGeocodeOperation`.

Additionally, even if a crash doesn’t result, it has the potential to return the “wrong” data for operationName, depending on timing (and when/if the Operation’s name is changed).

The issue with `Operation._log` being a lazy var is also substantial. From the Swift book:

> If a property marked with the lazy modifier is accessed by multiple threads simultaneously and the property has not yet been initialized, there is no guarantee that the property will be initialized only once.

Simple Repro Case for `lazy var`-related Crash:

```swift
func test__operations_log_trample() {
    let expectation = expectationWithDescription("Test: \(#function)")
    
    dispatch_async(Queue.Initiated.queue) {
        let group = dispatch_group_create()
        for i in 0..<1000000 {
            let operation = Operation()
            operation.name = "Test Operation \(i)"
            dispatch_group_async(group, dispatch_get_global_queue(0, 0), {
                operation.log.verbose("Test")
            })
            dispatch_group_async(group, dispatch_get_global_queue(0, 0), {
                operation.log.verbose("Test")
            })
        }
        dispatch_group_wait(group,DISPATCH_TIME_FOREVER)
        expectation.fulfill()
    }
    
    waitForExpectationsWithTimeout(60, handler: nil)
}
```

#### Initial Solution:

- `Operation._log` must not be a lazy var.
- `Operation.log` must be thread-safe.

> Initial fix involves capturing a copy of the logging context (settings + operationName) at the time of access of `Operation.log`, returning that captured context as a LoggerType, and surrounding access of .log with a Read/Write lock.

- LogManager properties must be thread-safe.

danthorpe:
@swiftlyfalling - wow, just wow.

I'm reading through it all today. Looks fantastic - really happy to see many improvements here, and the stress tests to validate the fixes. Truly fantastic work - stuff like this is why I love working on open source projects!

Will squash & merge later today I expect, and then will cherry pick into the release branch - I think it makes sense to put all these improvements into v3

Sources/Core/Shared/Logging.swift
```diff
@@ -284,11 +310,50 @@ public class LogManager: LogManagerType {
...
-    var enabled: Bool = true
-    var severity: LogSeverity = .Warning
-    var logger: LoggerBlockType = { message, severity, file, function, line in
-        print("\(LogManager.metadataForFile(file, function: function, line: line))\(message)")
+    var enabled: Bool {
+        get { return _enabled.read { $0 } }
+        set {
+            _enabled.write { (value) in
+                value = newValue
+            }
```
danthorpe: I'm thinking that it would probably make sense to add subscript support to `Protector` to avoid having to this stuff - quite a bit of boiler plate required when using `Protector` like this...

Sources/Core/Shared/Operation.swift
```diff
@@ -79,27 +79,24 @@ public class Operation: NSOperation {
...
-        return ["State"]
-    }
-
-    class func keyPathsForValuesAffectingIsCancelled() -> Set<NSObject> {
-        return ["Cancelled"]
+    private enum NSOperationKeyPaths: String {
+        case Cancelled = "isCancelled"
+        case Executing = "isExecuting"
+        case Finished = "isFinished"
     }
```
danthorpe: Just thinking a little bit down the road, with Swift 3.0. It would probably be better to add this to `Foundation.NSOperation` instead. Because soon "NSOperation" is not going to be a thing.

```swift
private extension Foundation.NSOperation {
    enum KeyPaths: String {
        case Cancelled = "isCancelled"
        case Executing = "isExecuting"
        case Finished = "isFinished"
    }
}
```

Sources/Core/Shared/Operation.swift
```diff
@@ -204,14 +185,60 @@ public class Operation: NSOperation {
...
+        (b) when willExecuteObservers log errors
+
+     To ensure that an Operation subclass does not finish until the
+     subclass calls finish():
+     call `super.init(disableAutomaticFinishing: true)` in the init.
+
+     IMPORTANT: If disableAutomaticFinishing == TRUE, the subclass is
+     responsible for calling finish() in *ALL* cases, including when the
+     operation is cancelled.
+
```
danthorpe: Just thinking that here it might be wise to include how a subclass should do this.

Earlier, while developing the `WillCancelObserver` and `DidCancelObserver` it was suggested that `Operation` make its `cancel()` override final. The intention being that subclasses would use these two observers to manage their internal state.

Sources/Core/Shared/Operation.swift
```diff
@@ -204,14 +185,60 @@ public class Operation: NSOperation {
...
+    */
+    public init(disableAutomaticFinishing: Bool) {
+        self._disableAutomaticFinishing = disableAutomaticFinishing
+        super.init()
+    }
+
+    private var disableAutomaticFinishing: Bool {
+        return _disableAutomaticFinishing
+    }
```
danthorpe: Rather than use a `var` and then have a private mutable backing store, and a private reader... how about using a constant, and then override `init()` to set it to false?

Sources/Core/Shared/Operation.swift
```diff
@@ -303,10 +334,14 @@ public class Operation: NSOperation {
      - parameter errors: an `[ErrorType]` defaults to empty array.
      */
     public func cancelWithErrors(errors: [ErrorType] = []) {
-        if !errors.isEmpty {
-            log.warning("Did cancel with errors: \(errors).")
+        stateLock.withCriticalScope {
+            assert(state != .Finishing, "Operation cannot be cancelled after while it is finishing.")
```
danthorpe: There is some confusing english here.

> Operation cannot be cancelled while it is finishing

Sources/Core/Shared/Operation.swift
```diff
@@ -303,10 +334,14 @@ public class Operation: NSOperation {
...
+            assert(state != .Finished, "Operation cannot be cancelled after it has already finished.")
```
danthorpe: If we have thread-safety in `Operation` around cancelling and finishing, are these asserts really necessary now? Their use puts the burden onto the framework consumer to then check that an operation is not getting cancelled at the same time that its finishing - which I guess seems redundant if the base class can work around it. No? Thoughts?

Sources/Core/Shared/Operation.swift
```diff
@@ -327,13 +362,43 @@ public class Operation: NSOperation {
...
+        let willFinish = stateLock.withCriticalScope { () -> Bool in
+            let willFinish = executing && !disableAutomaticFinishing && !_isHandlingFinish
+            if willFinish {
+                _isHandlingFinish = true
             }
+            return willFinish
+        }
+        if willFinish {
+            _finish([], fromCancel: true)
         }
```
danthorpe: Very nice!

Sources/Core/Shared/Operation.swift
```diff
@@ -569,33 +684,69 @@ public extension Operation {
...
+            log.warning("Will finish with \(errors.count) errors.")
+        }
 
-            willFinishObservers.forEach { $0.willFinishOperation(self, errors: self._internalErrors) }
+        operationWillFinish(errors)
+        willChangeValueForKey(NSOperationKeyPaths.Finished.rawValue)
+        willFinishObservers.forEach { $0.willFinishOperation(self, errors: errors) }
 
-            state = .Finished
+        state = .Finished
```
danthorpe: Does this not need to be in the state lock?

```swift
stateLock.withCriticalScope {
    state = .Finished
}
```

Sources/Core/Shared/Operation.swift
```diff
@@ -569,33 +684,69 @@ public extension Operation {
...
-            }
-            else {
-                log.warning("Did finish with errors: \(errors).")
-            }
+        if errors.isEmpty {
+            log.verbose("Did finish with no errors.")
         }
+        else {
+            log.warning("Did finish with errors: \(errors).")
+        }
```
danthorpe: I guess should probably clean this up too: 

```swift
let message = errors.isEmpty ? "errors: \(errors)" : "no errors"
```

Sources/Core/Shared/OperationQueue.swift
```diff
@@ -102,6 +112,9 @@ public class OperationQueue: NSOperationQueue {
 
             /// Add an observer so that any produced operations are added to the queue
             operation.addObserver(ProducedOperationObserver { [weak self] op, produced in
+                if let q = self {
+                    q.delegate?.operationQueue(q, willProduceOperation: produced)
+                }
                 self?.addOperation(produced)
```
danthorpe: Might as well just add this line into the `if let q` too. 🙂

Sources/Core/Shared/RepeatedOperation.swift
```diff
@@ -299,6 +299,7 @@ public class RepeatedOperation<T where T: NSOperation>: GroupOperation {
      can prevent another operation from being added.
     */
     public func addNextOperation(@autoclosure shouldAddNext: () -> Bool = true) -> Bool {
+        guard !cancelled else { return false }
         guard shouldAddNext(), let payload = next() else { return false }
```
danthorpe: I think maybe it's nicer to add the `!cancelled` to the guard that is already there?

```swift
guard !cancelled && shouldAddNext(), let payload = next() else { return false }
```

Sources/Core/Shared/Operation.swift
swiftlyfalling: Leftover from a prior implementation detail that was quickly reconsidered. (It doesn't make sense to allow the value to change after init().) Fixing. 😊

Sources/Core/Shared/Operation.swift
swiftlyfalling: Whoopsies. Fixing.

Sources/Core/Shared/Operation.swift
swiftlyfalling: Was going back and forth over whether this was something that made sense to alert the framework consumer about. In the case of `cancel()`/`cancelWithErrors()`, I agree with your suggestion - it should be something you can safely call at any time, from any thread, without it asserting.
Fixing.

Sources/Core/Shared/Operation.swift
swiftlyfalling: Not needed because of a side-effect of the setter of Operation.`state` (which acquires the lock), but I agree that it's much better to be explicit about it (and not doing so is an oversight). Fixing.

Sources/Core/Shared/Operation.swift
swiftlyfalling: Nice! Fixing.

Sources/Core/Shared/OperationQueue.swift
swiftlyfalling: Yup. Fixing.

Sources/Core/Shared/RepeatedOperation.swift
swiftlyfalling: Sure. Fixing.

Sources/Core/Shared/Operation.swift
swiftlyfalling: Moved the KeyPaths stuff out to an extension of NSOperation. Makes sense to me.

Sources/Core/Shared/Operation.swift
swiftlyfalling: I actually completely agree with making `Operation.cancel()` final (especially at this point, with all the thread-safety stuff). Was going to wait to submit that as a separate PR.

swiftlyfalling: Added some suggestions on how to accomplish handling cancellation.

Sources/Core/Shared/Operation.swift
```diff
@@ -227,16 +222,23 @@ public class Operation: NSOperation {
...
+     and/or checking periodically during execute with something like:
+
+     ```swift
+     guard !cancelled else {
+        // do any necessary clean-up
+        finish()    // always call finish if automatic finishing is disabled
+        return
+     }
+     ```
+
```
danthorpe: This is great!

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -122,10 +142,53 @@ public class GroupOperation: Operation, OperationQueueDelegate {
...
+                // It is possible that the cancellation happened before adding the
+                // additional operations to the operations array.
+                // Thus, ensure that all additional operations are cancelled.
+                additional.forEachOperation { if !$0.cancelled { $0.cancel() } }
+            }
+
+            groupFinishLock.withCriticalScope {
+                dispatch_group_leave(isAddingOperationsGroup)
+            }
         }
```
danthorpe: In the method above - there are quite a few usages of `self.` to access properties. In the codebase, this is only really don't inside closures or initialisers. While it does seem a little silly to discuss style in a 2 year old language, this is consistent throughout, and it helps to reduce visual noise and highlight captures. 🙂

swiftlyfalling: Not silly - an earlier commit used a lock (and, thus, closure) around all of that code before switching to finer-grained locks. Tweaking shortly.

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -185,22 +248,38 @@ public class GroupOperation: Operation, OperationQueueDelegate {
 
     /**
      The group operation acts as its own queue's delegate. When an operation is added to the queue,
-     assuming that the finishing operation has not started (or finished), and the operation is not
-     the finishing operation itself, then we add the operation as a dependency to the finishing
-     operation.
+     assuming that the finishing operation has not started (or finished), the operation is not
+     the finishing operation itself, and the CanFinishOperation has not yet started,
+     then we add the operation as a dependency to the CanFinishOperation.
```
danthorpe: This is an odd issue regarding the docs. `CanFinishOperation` is a private class which the framework consumer has no idea about. Is there a way to describe the behaviour without mentioning this class?

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -185,22 +248,38 @@ public class GroupOperation: Operation, OperationQueueDelegate {
...
 
      The purpose of this is to keep the finishing operation as the last child operation that executes
-     when there are no more operations in the group.
+     when there are no more operations in the group, and to safely handle GroupOperation state changes.
      */
     public func operationQueue(queue: OperationQueue, willAddOperation operation: NSOperation) {
+        guard queue === self.queue else { return }
```
danthorpe: Nice!

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -185,22 +248,38 @@ public class GroupOperation: Operation, OperationQueueDelegate {
...
+            guard shouldContinue else { return }
 
             willAddChildOperationObservers.forEach { $0.groupOperation(self, willAddChildOperation: operation) }
 
-            finishingOperation.addDependency(operation)
+            canFinishOperation.addDependency(operation)
+            
+            groupFinishLock.withCriticalScope {
+                dispatch_group_leave(isAddingOperationsGroup)
+            }
```
danthorpe: Hmm, this is interesting, is this because `dispatch_group_enter/leave` is not thread safe? I maybe need to refresh myself on the docs....

danthorpe: Discussed with @swiftlyfalling - in essence this is because `dispatch_group_enter/leave`'s locking mechanism is noticeably slower than `NSRecursiveLock`.

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -225,12 +305,41 @@ public class GroupOperation: Operation, OperationQueueDelegate {
...
     public func operationQueue(queue: OperationQueue, didFinishOperation operation: NSOperation, withErrors errors: [ErrorType]) {
+        guard queue === self.queue else { return }
 
         if operation === finishingOperation {
             finish(fatalErrors)
             queue.suspended = true
         }
     }
+
+    public func operationQueue(queue: OperationQueue, willProduceOperation operation: NSOperation) {
```
danthorpe: I really like this addition. We previously add some (poor in hindsight) code in `OperationQueue` to watch out for this. But this is much nicer.

Sources/Core/Shared/GroupOperation.swift
swiftlyfalling: Hmm.

Arguably, the finishing operation is also private and an implementation detail of GroupOperation.

How about something like:

> The group operation acts as its own queue's delegate. When an operation is added to the queue, assuming that the GroupOperation is not yet finishing or finished, then we add the operation as a dependency to an internal operation.

> The purpose of this is to keep the internal operation as a "barrier" child operation that executes when there are no more operations in the group, safely handling the transition GroupOperation state.

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -225,12 +305,41 @@ public class GroupOperation: Operation, OperationQueueDelegate {
...
+            assert(!finishingOperation.finished, "Cannot produce new operations within a group after the group has completed.")
+            guard !isGroupFinishing else {
+                assertionFailure("Cannot produce new operations within a group after the group has started to finish.")
+                return false
+            }
+            dispatch_group_enter(isAddingOperationsGroup)
+            return true
+        }
+
+        guard shouldContinue else { return }
```
danthorpe: I can see this is a common pattern. Returning a bool from a critical scope with a success block, then `guard` on the returned value & exit. As opposed to:

```swift
guard let check = lock.withCriticalScope { () -> Bool in
    guard checkSomethingSafely else { return false }
    doSomethingElse()
    return true
} else { return }
```

I'm just wondering if we can extend `NS(Recursive)Lock` to accept a boolean which we wrap in an `@autoclosure` to do this for us?

Something like:

```swift
extension NSRecursiveLock {
    func guardWithCriticalScope(@autoclosure check: () -> Bool, assertionMessage: String? = .None, success: @noescape (() -> Void)? = .None) -> Bool {
        return withCriticalScope {
            guard check() else {
                if let message = assertionMessage {
                    assertionFailure(message)
                }
                return false
            }
            success?()
            return true
        }
    }
}
```

Usage would then be something like:

```swift
guard let shouldContinue = lock.guardWithCriticalScope(!isGroupFinishing, success: { dispatch_group_enter(isAddingOperationsGroup) }) else { return }
```

Still means you have a trailing closure with you might then want to ` else { }` from... which I'm not 100% sure about.

Just an idea. This same pattern is used here a couple of times and in `Operation`.

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -225,12 +305,41 @@ public class GroupOperation: Operation, OperationQueueDelegate {
...
+
+        _operations.append(operation)
+        if cancelled {
+            if !operation.cancelled { operation.cancel() }
+        }
```
danthorpe: Can we re-write this as:

```swift
if cancelled && !operation.cancelled {
    operation.cancel()
}
```

It took me a little while to try to figure out if this was a trailing closure or not.

swiftlyfalling: Done.

Sources/Core/Shared/GroupOperation.swift
swiftlyfalling: Done.

Sources/Core/Shared/Operation.swift
danthorpe: Okay - lets make `cancel()` final in 3.1.

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -339,3 +448,117 @@ public struct WillAddChildObserver: GroupOperationWillAddChildObserver {
...
+     - a finishingOperation, which has the CanFinishOperation as a dependency
+
+     The purpose of this is to handle the possibility that GroupOperation.addOperation() or
+     GroupOperation.queue.addOperation() are called right after all current child operations have
+     completed (i.e. after the CanFinishOperation has been set to ready), but *prior* to being able
+     to process that the GroupOperation is finishing (i.e. prior to the CanFinishOperation executing and
+     acquiring the GroupOperation.groupFinishLock to set state).
+     */
+    private class CanFinishOperation: NSOperation {
+        private weak var parent: GroupOperation?
```
danthorpe: Just thinking, would this not be better as `unowned` rather than weak. This would avoid having to unwrap it all the time.

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -339,3 +448,117 @@ public struct WillAddChildObserver: GroupOperationWillAddChildObserver {
...
+        private var _finished = false
+
+        init(parentGroupOperation: GroupOperation) {
+            self.parent = parentGroupOperation
+            super.init()
+        }
+        override func start() {
+            main()
+        }
```
danthorpe: Are you sure you want to override `start()` without calling `super.start()` - I would just not over-ride here at all.

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -339,3 +448,117 @@ public struct WillAddChildObserver: GroupOperationWillAddChildObserver {
...
+
+                    // Is anything currently adding operations?
+                    guard dispatch_group_wait(parent.isAddingOperationsGroup, DISPATCH_TIME_NOW) == 0 else {
+                        // Operations are actively being added to the group
+                        // Wait for this to complete before proceeding.
+                        dispatch_group_notify(parent.isAddingOperationsGroup,
+                            dispatch_get_global_queue(NSQualityOfServiceToDispatchQOS(self.qualityOfService), 0),
+                            { [unowned self] in
+                            self.execute()
+                        })
```
danthorpe: I think it would be great to add an initialiser to `Queue` here which takes a `qos_class` and/or `NSQualityOfService`. See Queue.swift.

```swift
    /**
     Initialize a Queue with a given NSQualityOfService.
     
     - parameter qos: a NSQualityOfService value
     - returns: a Queue with an equivalent quality of service
    */
    public init(qos: NSQualityOfService) {
        switch qos {
        case .Background:
            self = .Background
        case .Default:
            self = .Default
        case .UserInitiated:
            self = .Initiated
        case .UserInteractive:
            self = .Interactive
        case .Utility:
            self = .Utility
        }
    }

    /**
     Initialize a Queue with a given GCD quality of service class.

     - parameter qos: a qos_class_t value
     - returns: a Queue with an equivalent quality of service
     */
    public init(qos: qos_class_t) {
        switch qos {
        case qos_class_main():
            self = .Main
        case QOS_CLASS_BACKGROUND:
            self = .Background
        case QOS_CLASS_USER_INITIATED:
            self = .Initiated
        case QOS_CLASS_USER_INTERACTIVE:
            self = .Interactive
        case QOS_CLASS_UTILITY:
            self = .Utility
        default:
            self = .Default
        }
    }
```

If you add that into Queue.swift, then your code above can become:

```swift
dispatch_group_notify(parent.isAddingOperationsGroup, Queue(qos: qualityOfService).queue, { [unowned self] in
    self.execute()
})
```

Also, is it not possible to just pass `execute` directly?

```swift
dispatch_group_notify(parent.isAddingOperationsGroup, Queue(qos: qualityOfService).queue, execute)
```

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -339,3 +448,117 @@ public struct WillAddChildObserver: GroupOperationWillAddChildObserver {
...
+            didChangeValueForKey("isFinished")
+        }
+        private func NSQualityOfServiceToDispatchQOS(qualityOfService: NSQualityOfService) -> qos_class_t {
+            let mapping = [NSQualityOfService.Background: QOS_CLASS_BACKGROUND,
+                           NSQualityOfService.Default: QOS_CLASS_DEFAULT,
+                           NSQualityOfService.UserInitiated: QOS_CLASS_USER_INITIATED,
+                           NSQualityOfService.UserInteractive: QOS_CLASS_USER_INTERACTIVE,
+                           NSQualityOfService.Utility: QOS_CLASS_UTILITY]
+            return mapping[qualityOfService] ?? QOS_CLASS_DEFAULT
+        }
```
danthorpe: This can go now, see comments above.

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -339,3 +448,117 @@ public struct WillAddChildObserver: GroupOperationWillAddChildObserver {
...
+                            self.execute()
+                        })
+                        return true
+                    }
+
+                    // Check whether new operations were added prior to the lock
+                    // by checking for child operations that are not finished.
+
+                    let nonFinishedOperations = parent.operations.filter({ !$0.finished })
+                    if !nonFinishedOperations.isEmpty {
```
danthorpe: I think here, from a readability point of view - it's best to try to express things to avoid double negatives. It just requires a little bit less mental jumping when reading through code. Suggest we rename this constant to `activeOperations`. This was, `if !activeOperation.isEmpty { } or `if activeOperations.count > 0 { }` is clearer what the logic is expressing.

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -339,3 +448,117 @@ public struct WillAddChildObserver: GroupOperationWillAddChildObserver {
...
+        private func NSQualityOfServiceToDispatchQOS(qualityOfService: NSQualityOfService) -> qos_class_t {
+            let mapping = [NSQualityOfService.Background: QOS_CLASS_BACKGROUND,
+                           NSQualityOfService.Default: QOS_CLASS_DEFAULT,
+                           NSQualityOfService.UserInitiated: QOS_CLASS_USER_INITIATED,
+                           NSQualityOfService.UserInteractive: QOS_CLASS_USER_INTERACTIVE,
+                           NSQualityOfService.Utility: QOS_CLASS_UTILITY]
+            return mapping[qualityOfService] ?? QOS_CLASS_DEFAULT
+        }
+        override var executing: Bool {
+            return false
```
danthorpe: How does this work?

Sources/Core/Shared/GroupOperation.swift
```diff
@@ -339,3 +448,117 @@ public struct WillAddChildObserver: GroupOperationWillAddChildObserver {
...
+            return _finished
+        }
+    }
+}
+
+private extension OperationQueue {
+    private func _addCanFinishOperation(canFinishOperation: GroupOperation.CanFinishOperation, finishingOperation: NSOperation) {
+        // Do not add observers (not needed - CanFinishOperation is an implementation detail of GroupOperation)
+        // Do not add conditions (CanFinishOperation has none)
+        finishingOperation.addDependency(canFinishOperation)
```
danthorpe: Hmm, maybe it would be better for the `GroupOperation` or `CanFinishOperation` to manage it's dependencies, rather than the queue?

danthorpe:
@swiftlyfalling - I've gone through everything, pretty much. Just a last few comments in the `GroupOperation` for you to give your thoughts on, then we're good to merge. 🚀 

Didn't have much to comment on all the tests. Generally look really good - happy to see that you've followed and extended the testing harness that is already in place 👍.

Once this lands, I will update CI to run the stress tests on iOS too to see how it performs.

Sources/Core/Shared/GroupOperation.swift
swiftlyfalling: Looks good - making changes.

Sources/Core/Shared/GroupOperation.swift
swiftlyfalling: Agreed - making the change.

danthorpe:
Looks great @swiftlyfalling! Just waiting for CI. Is there anything else?

swiftlyfalling:
That should be it for this PR. 😄

danthorpe:
Alright - thanks @swiftlyfalling for all this work - gonna squash & merge it now to develop. Will cherry pick  it into the v3 RC too.

PR #359, opened 2016-07-09
Make Operation.cancel() final

swiftlyfalling:
Per discussion in:
https://github.com/danthorpe/Operations/pull/293
https://github.com/danthorpe/Operations/pull/358#discussion_r70167086

(For a post-3.0 release.)


danthorpe:
@swiftlyfalling - great, I'm actually going to cherry pick this too. Makes sense to group any breaking changes into the same major version.

danthorpe:
Just realised - we should probably also make `cancelWithErrors` etc also final.

PR #361, opened 2016-07-12
Make GroupOperation's internal queue non-public, and expose appropriate properties on GroupOperation

swiftlyfalling:
- Make `GroupOperation.queue` internal, instead of public. (Internal so it can be tested.)
- Expose appropriate run-time editable properties of the GroupOperation’s queue as properties on GroupOperation.
> - public var maxConcurrentOperationCount: Int
> - public var suspended: Bool
> - public var qualityOfService: NSQualityOfService

- Add tests.

Per discussion in: https://github.com/danthorpe/Operations/pull/293

-----

#### Discussion:

Per discussion in:
https://github.com/danthorpe/Operations/pull/293

`GroupOperation.queue` probably shouldn’t be public, but there are run-time editable properties of the queue that ought to be configurable by framework consumers. Thus, this PR exposes those properties directly on GroupOperation.

The following properties of NSOperationQueue seem sensible to expose:

```swift
public var maxConcurrentOperationCount: Int
public var suspended: Bool
public var qualityOfService: NSQualityOfService
```

The following methods of NSOperationQueue do **not** seem sensible to expose:

```swift
public func cancelAllOperations()
```
**Why:** Use GroupOperation.cancel() instead.

```swift
public func waitUntilAllOperationsAreFinished()
```
**Why:** Use a DidFinishObserver on GroupOperation instead (and waiting on the queue itself is probably not what anyone should do, as the GroupOperation’s queue is an implementation detail of GroupOperation).

### DIFF
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -47,9 +47,9 @@ public class GroupOperation: Operation, OperationQueueDelegate {
     private var isGroupFinishing = false
     private let groupFinishLock = NSRecursiveLock()
     private var isAddingOperationsGroup = dispatch_group_create()
-
-    /// - returns: the OperationQueue the group runs operations on.
-    public let queue = OperationQueue()
+    private var groupSuspendLock = NSLock()
+    private var isGroupSuspended = false
+    internal let queue = OperationQueue()   // internal for testing
 
     /// - returns: the operations which have been added to the queue
     public private(set) var operations: [NSOperation] {
@@ -69,6 +69,77 @@ public class GroupOperation: Operation, OperationQueueDelegate {
             let (nsops, ops) = operations.splitNSOperationsAndOperations
             nsops.forEach { $0.setQualityOfServiceFromUserIntent(userIntent) }
             ops.forEach { $0.userIntent = userIntent }
+        }
+    }
+
+    /**
+     The maximum number of child operations that can execute at the same time.
+
+     The value in this property affects only the operations that the current GroupOperation has
+     executing at the same time. Other operation queues and GroupOperations can also execute
+     their maximum number of operations in parallel.
+
+     Reducing the number of concurrent operations does not affect any operations that are
+     currently executing.
+
+     Specifying the value NSOperationQueueDefaultMaxConcurrentOperationCount (which is recommended)
+     causes the system to set the maximum number of operations based on system conditions.
+
+     The default value of this property is NSOperationQueueDefaultMaxConcurrentOperationCount.
+    */
+    public final var maxConcurrentOperationCount: Int {
+        get {
+            return queue.maxConcurrentOperationCount
+        }
+        set {
+            queue.maxConcurrentOperationCount = newValue
+        }
+    }
+
+    /**
+     A Boolean value indicating whether the Group is actively scheduling operations for execution.
+
+     When the value of this property is false, the GroupOperation actively starts child operations
+     that are ready to execute once the GroupOperation has been executed.
+
+     Setting this property to true prevents the GroupOperation from starting any child operations,
+     but already executing child operations continue to execute.
+
+     You may continue to add operations to a GroupOperation that is suspended but those operations
+     are not scheduled for execution until you change this property to false.
+
+     The default value of this property is false.
+    */
+    public final var suspended: Bool {
+        get {
+            return groupSuspendLock.withCriticalScope { isGroupSuspended }
+        }
+        set {
+            groupSuspendLock.withCriticalScope {
+                isGroupSuspended = newValue
+                queue.suspended = newValue
+            }
+        }
+    }
+
+    /**
+     The default service level to apply to the GroupOperation and its child operations.
+
+     This property specifies the service level applied to the GroupOperation itself, and to
+     operation objects added to the GroupOperation.
+
+     If the added operation object has an explicit service level set, that value is used instead.
+
+     For more, see the NSOperation and NSOperationQueue documentation for `qualityOfService`.
+    */
+    @available(OSX 10.10, iOS 8.0, tvOS 8.0, watchOS 2.0, *)
+    public final override var qualityOfService: NSQualityOfService {
+        get {
+            return queue.qualityOfService
+        }
+        set {
+            super.qualityOfService = newValue
+            queue.qualityOfService = newValue
         }
     }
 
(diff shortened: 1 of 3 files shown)


swiftlyfalling:
Would it make sense to add some `@available` markers to GroupOperation.queue, and maybe have it return a stub that implements `@available` warnings for queue properties/functions? Something like:

```swift
@available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation's methods and properties, like: addOperation(), maxConcurrentOperationCount, suspended, qualityOfService.")
public var queue: GroupOperationQueueIsUnavailableStub { return GroupOperationQueueIsUnavailableStub() }
```

```swift
public class GroupOperationQueueIsUnavailableStub {

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation.operations instead.")
    public var operations: [NSOperation] { return [] }

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation.maxConcurrentOperationCount instead.")
    public final var maxConcurrentOperationCount: Int { return -1 }

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation.suspended instead.")
    public final var suspended: Bool { return false }

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation.qualityOfService instead.")
    public final var qualityOfService: NSQualityOfService { return NSQualityOfService.Default }

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Considering subclassing GroupOperation and implementing overrides for functions like willFinishOperation.")
    public weak var delegate: OperationQueueDelegate? = .None

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation.addOperation() instead.")
    func addOperation(operation: NSOperation) { }

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation.addOperations() instead.")
    func addOperations(ops: [NSOperation], waitUntilFinished wait: Bool) { }

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation.addOperations() instead.")
    public func addOperationWithBlock(block: () -> Void) { }

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation.addOperations() instead.")
    func addOperations<S where S: SequenceType, S.Generator.Element: NSOperation>(ops: S) { }

    @available(*, unavailable, message="GroupOperation's queue is an internal implementation detail. Use GroupOperation.addOperation() instead.")
    func addOperations(ops: NSOperation...) { }
}
```

Is there any easy way to do an automatic fix-it from `GroupOperation.queue.suspended` to `GroupOperation.suspended`? Thoughts?

jshier:
One capability I'd still like to have is the ability to customize the queue's `underlyingQueue` in some way. This enables the use of GCD APIs inside the `GroupOperation` on the same queue as the other tasks.

swiftlyfalling:
@jshier: When do you set the `underlyingQueue`? In a GroupOperation subclass's `init()`?

(According to the `NSOperationQueue` documentation, underlyingQueue `should only be set if there are no operations in the queue`, otherwise it will raise an exception. Since GroupOperation may  - and does - add additional operations to its internal queue to handle finishing and such, it seems to me that the only safe place to set underlyingQueue would be init. Thoughts?)

danthorpe:
@swiftlyfalling yep completely agree on your points above.

@jshier Perhaps we can add an optional argument to `GroupOperation` for the underlying queue, with a default to `.None`?

But I think the changes are great - so just waiting for some thoughts on the `underlyingQueue`.

jshier:
Yes, I think adding it as an initializer option is the only way to go. And it should likely be an `OperationQueue` parameter, not just a `dispatch_queue_t`.

danthorpe:
@jshier @swiftlyfalling yes, so something like:

```swift
init(queue: OperationQueue = OperationQueue(), operations: [NSOperation]) {
    // etc
}
```

danthorpe:
So, after further discussion in Slack, we're going to change this to accepting the underlying queue:

```swift
init(queue: dispatch_queue_t = Queue.Default.queue, operations: [NSOperation]) {
    // etc
}
```

The primary reason for this, is that if the queue is provided to the group from outside, where it could also be modified, for the group to work correctly, it must be set as the OperationQueue's delegate, which in turn could not be guaranteed.

danthorpe:
@swiftlyfalling regarding the availability markers.. if we expose the safe APIs through the `GroupOperation` we can make the `queue` property internal.

What we could also do, is provide a protocol, similar to your stub class here, for `queue` which is public, and then deprecate the entire API with Fix-It renames for the ones which are available, and messages for those that are not.

Essentially, we replace the implementation of `queue` out from underneath, but maintain its API. At the same time we use the change to force people to receive warnings for safe but moved interfaces, and refactor for unsafe and no longer available APIs.

danthorpe:
@swiftlyfalling shall I finish this off?

swiftlyfalling:
@danthorpe: Had a scheduling change with another project - haven't had a chance yet to make the remaining changes here. Please feel free!

swiftlyfalling:
One quick thought:

It might be best to default the new underlying queue parameter to nil. i.e.
```swift
init(queue: dispatch_queue_t? = nil, operations: [NSOperation]) {
    // etc
}
```

This matches the current behavior, which utilizes the default for [NSOperationQueue](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/NSOperationQueue_class/#//apple_ref/occ/instp/NSOperationQueue/underlyingQueue):

> ```swift 
unowned(unsafe) var underlyingQueue: dispatch_queue_t?
```
> **The default value of this property is `nil`.** You can set the value of this property to an existing dispatch queue to have enqueued operations interspersed with blocks submitted to that dispatch queue.
> 
> The value of this property should only be set if there are no operations in the queue; setting the value of this property when `operationCount` is not equal to `0` raises an `NSInvalidArgumentException`. The value of this property must not be the value returned by `dispatch_get_main_queue`. The quality-of-service level set for the underlying dispatch queue overrides any value set for the operation queue's `qualityOfService` property.

danthorpe:
@swiftlyfalling - yeah, that's exactly what I was going to do, except call it `underlyingQueue` then just set this on the queue's property.

I think it'll be better to let you do this, and then keep all the changes into this single PR - easier to follow the history/progress.

swiftlyfalling:
@danthorpe: Not a problem. I should have a chance to get back to this next week.

PR #362, opened 2016-07-12
Support composing simple types for Result injection

danthorpe:
Adopting _Operations_ might be considered a risky proposition - it's a "framework" with all the associated issues that come with adding 3rd party dependencies.

One way which might make this less burdensome would be if there was an `Operation` type which composed simple functionality type for Result Injection.

### DIFF
--- a/Sources/Core/Shared/GroupOperation.swift
+++ b/Sources/Core/Shared/GroupOperation.swift
@@ -263,13 +263,13 @@ public class GroupOperation: Operation, OperationQueueDelegate {
                 dispatch_group_enter(isAddingOperationsGroup)
                 return true
             }
-            
+
             guard shouldContinue else { return }
 
             willAddChildOperationObservers.forEach { $0.groupOperation(self, willAddChildOperation: operation) }
 
             canFinishOperation.addDependency(operation)
-            
+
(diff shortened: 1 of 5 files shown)


danthorpe:
The intended use case for this would be something like this:

```swift
class MyClassWhichTransformsState {
    func execute() {
        // Details are not important here, but it must support "execution"
    }
    func cancel() {
        // and cancellation  
    }
}

extension MyClassWhichTransformsState: Executor {
    var requirement: Bar {
        get { /* Totally up to framework */ }
        set { /* consumer how this works */ }
    }
    var result: Foo {
       return foo // Totally up to framework consumer how this works
    }
}
```

Given the above, this type which does not subclass `Operation`, can be used _as_ an operation like this:

```swift
let thing = MyClassWhichTransformsState()
let queue = OperationQueue()
let operation = Execute(thing)
queue.addOperation(operation)
```

And now  the `operation` can have dependencies, conditions, observers etc etc.

In some situations, the _executor_ might not have any requirements, in which case, set the type to Void like this:

```swift
extension MyClassWhichTransformsState: Executor {
    var requirement: Void = Void() // no requirements
    var result: Foo {
       return foo // Totally up to framework consumer how this works
    }
}
```

PR #365, opened 2016-07-15
Fix a log

DeFrenZ:
Fixed an operation finish log


Sources/Core/Shared/Operation.swift
```diff
@@ -741,7 +741,7 @@ public extension Operation {
         operationDidFinish(errors)
         didFinishObservers.forEach { $0.didFinishOperation(self, errors: errors) }
 
-        let message = errors.isEmpty ? "errors: \(errors)" : "no errors"
+        let message = !errors.isEmpty ? "errors: \(errors)" : "no errors"
```
danthorpe: Whoops! Thanks @DeFrenZ

PR #394, opened 2016-07-31
Deferred mutual exclusivity

danthorpe:
Given an operation which has a `Condition` which is mutually exclusive, currently the operation would be gain this exclusion lock when it is added to the queue.

This doesn't seem like it would be ideal, because it is very easy to create a deadlock where an operation is mutually exclusive with a child or dependent operation.  The child cannot start until the parent finishes, and the parent is waiting for the child to finish before it starts.

### DIFF
--- a/Sources/Core/Shared/ExclusivityManager.swift
+++ b/Sources/Core/Shared/ExclusivityManager.swift
@@ -24,20 +24,9 @@ internal class ExclusivityManager {
         return dispatch_sync(queue) { self._addOperation(operation, category: category) }
     }
 
-    func removeOperation(operation: Operation, category: String) {
-        dispatch_async(queue) {
-            self._removeOperation(operation, category: category)
-        }
-    }
-
     private func _addOperation(operation: Operation, category: String) -> NSOperation? {
-        operation.log.verbose(">>> \(category)")
 
-        operation.addObserver(DidFinishObserver { [unowned self] op, _ in
-            self.removeOperation(op, category: category)
-        })
-
-        var operationsWithThisCategory = operations[category] ?? []
+        let operationsWithThisCategory = operations[category] ?? []
 
         let previous = operationsWithThisCategory.last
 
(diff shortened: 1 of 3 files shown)


danthorpe:
@swiftlyfalling what do you think about this? This does actually work in the sense that now an operation "grabs" the mutually exclusive token only just before it runs, instead of when it gets added to the queue. I'm just trying to think through what the consequences of that are if you add operations which are all mutually exclusive on the same category to the queue.

swiftlyfalling:
@danthorpe: These changes no longer prevent mutually exclusive operations from executing concurrently.

A new test case, that fails with the new code but succeeds with the prior code:
(This takes advantage of the ConcurrencyRegistrar and RegistersWhenRunningOperation that were added with the recent GroupOperation changes.)
```swift
class MutualExclusiveTests2: OperationTests {
    
    private class ConcurrencyRegistrar {
        private let _sharedRunningOperations = Protector([NSOperation]())
        private let _maxConcurrentOperationsDetectedCount = Protector(Int(0))
        var maxConcurrentOperationsDetectedCount: Int {
            get {
                return _maxConcurrentOperationsDetectedCount.read { $0 }
            }
        }
        private func recordOperationsCount(currentOperationsCount: Int) {
            _maxConcurrentOperationsDetectedCount.write { (inout ward: Int) in
                if currentOperationsCount > ward {
                    ward = currentOperationsCount
                }
            }
        }
        func registerRunning(op: NSOperation) {
            _sharedRunningOperations.write { (inout ward: [NSOperation]) in
                ward.append(op)
                self.recordOperationsCount(ward.count)
            }
        }
        func deregisterRunning(op: NSOperation) {
            _sharedRunningOperations.write { (inout ward: [NSOperation]) in
                if let foundOperation = ward.indexOf(op) {
                    ward.removeAtIndex(foundOperation)
                }
            }
        }
        func atLeastOneOperationIsRunning(otherThan exception: NSOperation? = nil) -> Bool {
            return _sharedRunningOperations.read { ward -> Bool in
                for op in ward {
                    if op !== exception {
                        return true
                    }
                }
                return false
            }
        }
    }
    private class RegistersWhenRunningOperation: Operation {
        private weak var concurrencyRegistrar: ConcurrencyRegistrar?
        private let microsecondsToSleep: useconds_t
        init(microsecondsToSleep: useconds_t, registrar: ConcurrencyRegistrar) {
            self.concurrencyRegistrar = registrar
            self.microsecondsToSleep = microsecondsToSleep
            super.init()
        }
        override func execute() {
            concurrencyRegistrar?.registerRunning(self)
            usleep(microsecondsToSleep)
            concurrencyRegistrar?.deregisterRunning(self)
            finish()
        }
    }
    
    func test__mutually_exclusive_operation_are_run_exclusively() {
        let numOperations = 3
        let childDelayMicroseconds: useconds_t = 500000 // 0.5 seconds
        let concurrency = ConcurrencyRegistrar()
        let operations: [RegistersWhenRunningOperation] = (0..<numOperations).map { _ -> RegistersWhenRunningOperation in
            let op = RegistersWhenRunningOperation(microsecondsToSleep: childDelayMicroseconds, registrar: concurrency)
            let condition = MutuallyExclusive<RegistersWhenRunningOperation>()
            op.addCondition(condition)
            return op
        }
        queue.maxConcurrentOperationCount = numOperations

        let startTime = CFAbsoluteTimeGetCurrent()
        waitForOperations(operations[0], operations[1], operations[2]) // modify if numOperations is changed
        let endTime = CFAbsoluteTimeGetCurrent()
        let duration = Double(endTime) - Double(startTime)
        
        XCTAssertEqual(concurrency.maxConcurrentOperationsDetectedCount, 1)
        for i in operations.enumerate() {
            XCTAssertTrue(i.element.finished, "Child operation [\(i.index)] did not finish")
        }
        XCTAssertGreaterThanOrEqual(duration, Double(useconds_t(operations.count) * childDelayMicroseconds) / 1000000.0)
    }
}
```

Sources/Core/Shared/ExclusivityManager.swift
```diff
@@ -24,41 +24,48 @@ internal class ExclusivityManager {
...
-        operation.addObserver(DidFinishObserver { [unowned self] op, _ in
-            self.removeOperation(op, category: category)
-        })
-
-        var operationsWithThisCategory = operations[category] ?? []
+        let operationsWithThisCategory = operations[category] ?? []
 
         let previous = operationsWithThisCategory.last
 
         if let previous = previous {
```
swiftlyfalling: Because `operationsWithThisCategory` is only modified when an operation is executed, if I create a bunch of operations, add MutuallyExclusive conditions to them, and then add them all to a queue, none of them will be dependent on each other with the current code (i.e. `operationsWithThisCategory.last` will always be `nil`).

danthorpe:
Thanks @swiftlyfalling - that's pretty much what I expected.

PR #395, opened 2016-08-01
CloudKitOperation: ability to add configuration to the next (retry) operation

swiftlyfalling:
~~This PR adds two new public methods:~~

#### ~~**CloudKitOperation.setPrepareForRetryHandler()**:~~
~~Sets a handler block which receives a closure that can be used to add configuration blocks to the retry operation.~~

~~This will be useful for some of the new (iOS 10 / macOS 10.12) CloudKit operations that use pipelining, like [CKFetchRecordZoneChangesOperation](https://developer.apple.com/reference/cloudkit/ckfetchrecordzonechangesoperation) (where the client code may want to update the previousServerChangeToken for a zone in the event of prior partial success that was persisted).~~

~~Hypothetical pseudo-code example (doesn't work yet because support for the new CloudKit operations isn't yet included):~~
```swift
// note: this example is outdated - do not use
// 
// operation.setPrepareForRetryHandler { addConfigureBlock in
//     addConfigureBlock { nextOperation in
//         // only retry with zones that haven't completed
//         // and ensure that the per-zone previousServerChangeToken is the last-processed one
//         let incompleteZones = zones.filter { (recordZoneID) -> Bool in
//             return !(zoneCompleted[recordZoneID] ?? false)
//         }
//         nextOperation.recordZoneIDs = incompleteZones
// 
 //        var optionsForIncompleteZones: [CKRecordZoneID: CKFetchRecordZoneChangesOptions] = [:]
//         for zone in incompleteZones {
//             let options = CKFetchRecordZoneChangesOptions()
//             options.previousServerChangeToken = currentChangeTokens[zone]
//             optionsForIncompleteZones[zone] = options
//         }
//         nextOperation.optionsByRecordZoneID = optionsForIncompleteZones
//     }
// }
```

#### ~~**BatchedCloudKitOperation.setPrepareForNextOperationHandler()**:~~

~~Sets a handler block which receives a closure that can be used to add configuration blocks to the next CloudKitOperation<T> (i.e. when the current CloudKitOperation returns with moreComing == true)~~

~~For example, if the "next" operation needs to use the updated serverChangeToken:~~
```swift
// batchedOperation.setPrepareForNextOperationHandler { addConfigureBlock in
//     addConfigureBlock { nextOperation in
//         nextOperation.previousServerChangeToken = currentServerChangeToken // <-- (retrieved/persisted elsewhere)
//     }
// }
```

### DIFF
--- a/Sources/Features/Shared/CloudKit/CloudKitOperation.swift
+++ b/Sources/Features/Shared/CloudKit/CloudKitOperation.swift
@@ -27,15 +27,15 @@ public class OPRCKOperation<T where T: NSOperation, T: CKOperationType>: Compose
 public class CloudKitRecovery<T where T: NSOperation, T: CKOperationType, T: AssociatedErrorType, T.Error: CloudKitErrorType> {
     public typealias V = OPRCKOperation<T>
 
-    public typealias ErrorResponse = (delay: Delay?, configure: V -> Void)
+    public typealias ConfigureOperationBlock = V -> Void
+    public typealias ErrorResponse = (delay: Delay?, configure: ConfigureOperationBlock)
     public typealias Handler = (operation: T, error: T.Error, log: LoggerType, suggested: ErrorResponse) -> ErrorResponse?
-    public typealias PrepareForRetryHandler = (addConfigureBlock: (V -> Void) -> Void) -> Void
 
     typealias Payload = RepeatedPayload<V>
 
     var defaultHandlers: [CKErrorCode: Handler]
     var customHandlers: [CKErrorCode: Handler]
-    var prepareForRetryHandler: PrepareForRetryHandler?
+    private var finallyConfigureRetryOperationBlock: ConfigureOperationBlock?
 
     init() {
         defaultHandlers = [:]
@@ -55,14 +55,14 @@ public class CloudKitRecovery<T where T: NSOperation, T: CKOperationType, T: Ass
             return .None
         }
 
-        if let prepareForRetryHandler = prepareForRetryHandler {
-            prepareForRetryHandler(addConfigureBlock: { (block) in
-                let config = response.configure
-                response.configure = { operation in
-                    config(operation)
-                    block(operation)
-                }
-            })
+        // The error handler responded with a suggested delay and configuration block.
+        // If a finallyConfigureRetryOperationBlock is specified, add it to the configuration block.
+        if let finallyConfigureRetryOperationBlock = finallyConfigureRetryOperationBlock {
+            let config = response.configure
+            response.configure = { operation in
+                config(operation)
+                finallyConfigureRetryOperationBlock(operation)
+            }
         }
 
         return response
@@ -107,8 +107,8 @@ public class CloudKitRecovery<T where T: NSOperation, T: CKOperationType, T: Ass
         customHandlers.updateValue(handler, forKey: code)
     }
 
-    func setPrepareForRetryHandler(handler: PrepareForRetryHandler?) {
-        prepareForRetryHandler = handler
+    func setFinallyConfigureRetryOperationBlock(block: ConfigureOperationBlock?) {
+        finallyConfigureRetryOperationBlock = block
     }
 
     internal func cloudKitErrorsFromInfo(info: RetryFailureInfo<OPRCKOperation<T>>) -> (code: CKErrorCode, error: T.Error)? {
@@ -271,8 +271,29 @@ public final class CloudKitOperation<T where T: NSOperation, T: CKOperationType,
         recovery.customHandlers = handlers
     }
 
-    public func setPrepareForRetryHandler(handler: (addConfigureBlock: (OPRCKOperation<T> -> Void) -> Void) -> Void) {
-        recovery.setPrepareForRetryHandler(handler)
+    // When an error occurs, CloudKitOperation executes the appropriate error handler (as long as the completion block is set).
+    // (By default, certain errors are automatically handled with a retry attempt, such as common network errors.)
+    //
+    // If the error handler specifies that the operation should retry, it also specifies a configuration block for the operation.
+    // After the configuration block returned by the error handler configures the operation, the finallyConfigureRetryOperationBlock
+    // will be passed the new ("retry") operation so it can further modify its properties.
+    //
+    // For example:
+    //      - CKFetchDatabaseChangesOperation, for updating:
+    //          - `previousServerChangeToken`
+    //              - with the last changeToken received by the changeTokenUpdatedBlock prior to the error
+    //                (assuming you have persisted the information received prior to the last changeToken update)
+    //
+    //      - CKFetchRecordZoneChangesOperation, for updating:
+    //          - `recordZoneIDs`
+    //              - to remove zones that were completely fetched prior to the error
+    //                (assuming you have persisted the fetched data)
+    //          - `optionsByRecordZoneID`
+    //              - to update the previousServerChangeToken for zones that were partially fetched prior to the error
+    //                (assuming you have persisted the successfully-fetched data)
+    //
+    public func setFinallyConfigureRetryOperationBlock(block: (retryOperation: OPRCKOperation<T>) -> Void) {
+        recovery.setFinallyConfigureRetryOperationBlock(block)
     }
 }
 
@@ -280,7 +301,7 @@ public final class CloudKitOperation<T where T: NSOperation, T: CKOperationType,
 
 class CloudKitOperationGenerator<T where T: NSOperation, T: CKOperationType, T: AssociatedErrorType, T.Error: CloudKitErrorType>: GeneratorType {
 
-    typealias PrepareForNextOperationHandler = (addConfigureBlock: (CloudKitOperation<T> -> Void) -> Void) -> Void
+    typealias ConfigureBlock = CloudKitOperation<T> -> Void
 
     let recovery: CloudKitRecovery<T>
 
@@ -288,8 +309,8 @@ class CloudKitOperationGenerator<T where T: NSOperation, T: CKOperationType, T: 
     var generator: AnyGenerator<T>
     var more: Bool = true
 
-    var prepareForNextOperationHandler: PrepareForNextOperationHandler?
-    var lastOperationConfigure: (CloudKitOperation<T> -> Void) = { _ in }
+    var configureNextOperationBlock: ConfigureBlock?
+    var lastOperationConfigure: ConfigureBlock = { _ in }
 
     init<G where G: GeneratorType, G.Element == T>(timeout: NSTimeInterval? = 300, generator: G) {
         self.timeout = timeout
@@ -304,8 +325,8 @@ class CloudKitOperationGenerator<T where T: NSOperation, T: CKOperationType, T: 
         return operation
     }
 
-    func setPrepareForNextOperationHandler(handler: PrepareForNextOperationHandler?) {
-        prepareForNextOperationHandler = handler
+    func setConfigureNextOperationBlock(block: ConfigureBlock?) {
+        configureNextOperationBlock = block
     }
 }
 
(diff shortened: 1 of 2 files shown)


danthorpe:
Hi @swiftlyfalling - not had a chance to look though this yet - unfortunately, won't get to it until tomorrow now...

Sources/Features/Shared/CloudKit/CloudKitOperation.swift
```diff
@@ -255,18 +270,27 @@ public final class CloudKitOperation<T where T: NSOperation, T: CKOperationType,
...
+    public func setPrepareForRetryHandler(handler: (addConfigureBlock: (OPRCKOperation<T> -> Void) -> Void) -> Void) {
+        recovery.setPrepareForRetryHandler(handler)
+    }
 }
 
 // MARK: - BatchedCloudKitOperation
 
 class CloudKitOperationGenerator<T where T: NSOperation, T: CKOperationType, T: AssociatedErrorType, T.Error: CloudKitErrorType>: GeneratorType {
 
+    typealias PrepareForNextOperationHandler = (addConfigureBlock: (CloudKitOperation<T> -> Void) -> Void) -> Void
```
danthorpe: Can we get some more typealiases on this...

```swift
typealias ConfigureBlock = CloudKitOperation<T> -> Void
typealias AddConfigureBlock = ConfigureBlock -> Void
typealias PrepareForNextOperationHandler = (addConfigureBlock: AddConfigureBlock) -> Void
```

Sources/Features/Shared/CloudKit/CloudKitOperation.swift
```diff
@@ -297,24 +328,54 @@ public class BatchedCloudKitOperation<T where T: NSOperation, T: CKBatchedOperat
     init<G where G: GeneratorType, G.Element == T>(timeout: NSTimeInterval? = 300, generator gen: G, enableBatchProcessing enable: Bool = true) {
 
         enableBatchProcessing = enable
-        generator = CloudKitOperationGenerator(timeout: timeout, generator: gen)
+
+        // Creates a CloudKitOperationGenerator object
+        let _generator = CloudKitOperationGenerator(timeout: timeout, generator: gen)
+
+        // Creates a Configuration Handler for next operations using the CloudKitOperationGenerator
+        let newConfigurationHandler: ConfigurationHandler = { [weak _generator] (nextOperation) -> PayLoad.ConfigureBlock? in
```
danthorpe: I'd just capture this as `[unowned _generator]` and not worry about the necessary unwrapping:

```swift
guard let prepareForNextOperationHandler = _generator.prepareForNextOperationHandler else { return .None }
```

Sources/Features/Shared/CloudKit/CloudKitOperation.swift
```diff
@@ -297,24 +328,54 @@ public class BatchedCloudKitOperation<T where T: NSOperation, T: CKBatchedOperat
...
         let tuple = TupleGenerator(primary: generator, secondary: delay)
-        let mapped = MapGenerator(tuple) { RepeatedPayload(delay: $0.0, operation: $0.1, configure: .None) }
+        let mapped = MapGenerator(tuple) { RepeatedPayload(delay: $0.0, operation: $0.1, configure: newConfigurationHandler($0.1)) }
         super.init(generator: AnyGenerator(mapped))
     }
 
     public override func willFinishOperation(operation: NSOperation) {
         if let cloudKitOperation = operation as? CloudKitOperation<T> {
             generator.more = enableBatchProcessing && cloudKitOperation.current.moreComing
+            generator.lastOperationConfigure = self.configure
```
danthorpe: Can we drop the `self.` please?

Tests/Features/CloudKitOperationTests.swift
```diff
@@ -1440,6 +1441,45 @@ class CloudKitOperationDiscoverAllContractsTests: CKTests {
         XCTAssertEqual(errorHandlers.count, 1)
         XCTAssertNotNil(errorHandlers[.InternalError])
     }
+
+    func test__set_prepare_for_retry_handler__with_error_which_retries_with_default_retry_handler() {
+        var userInfosByAddress: [String: TestDiscoverUserInfosOperation.DiscoveredUserInfo]? = .None
+        var userInfosByRecordID: [TestDiscoverUserInfosOperation.RecordID: TestDiscoverUserInfosOperation.DiscoveredUserInfo]? = .None
+
+        var shouldError = true
+        let operation: CloudKitOperation<TestDiscoverUserInfosOperation> =
```
danthorpe: FYI - pretty sure that you don't need to specify the type here. Swift will figure it out from the return type.

danthorpe:
@swiftlyfalling - looks good. I made a few comments, nothing too serious. Also, would be great if you could pull `development` into the branch.

Also, and I know this was my suggestion... but just so that we're clear... the `prepareForRetryOperation` handler, is a closure which receives a closure which receives a configure closure which receives the operation....

While this is the most flexible - as it allows the framework consumer to do whatever they want in that closure, I'm am wondering if it's a bit complicated. As in, you have to call this block, with another closure, which will receive the operation.

We could, I think, make it so that the `prepareForRetryOperation` handler receives the operation, and the framework consumer can set it up there. Internally, we would then optionally execute this block at the correct time.

To re-iterate I'm not looking to change it necessarily, I'm just re-visiting the decision to double check what makes sense for the consumer. What do you think?

(Otherwise - the code & tests look great 🙌)

swiftlyfalling:
Will make the suggested code tweaks later today.

Regarding the structure of this addition:

If the goal is simply to permit changing the next operation's parameters, the "simpler" method you proposed could work. (Although it would seem to require more code changes to other things, like Repeated/RetryOperation.)

But this method could also be extended to support other functionality.

For example:
What if a consumer wants to override the default retry behavior, regardless of the cause of the retry (i.e. regardless of the specific CKError) based on some other state?

Instead of them having to set an error handler for *every single error code that needs to potentially retry* (and thus duplicate the built-in retry support), they can just set this handler, once. (It would obviously need to be changed to also provide the suggested delay & configure, and return a new one.)

Tests/Features/CloudKitOperationTests.swift
swiftlyfalling: For whatever reason, the Swift 2.2 compiler does not - it errors. (Not sure about Swift 3.)

swiftlyfalling:
This PR adds two new public methods:
```diff
class CloudKitOperation {
+    func setFinallyConfigureRetryOperationBlock()
}
```
```diff
class BatchedCloudKitOperation {
+    func setConfigureNextOperationBlock()
}
```

## CloudKitOperation:

### **CloudKitOperation. setFinallyConfigureRetryOperationBlock()**:

Sets a block that can be used to configure retry operations, after the error handler configures them.

This will be useful for some of the new (iOS 10 / macOS 10.12) CloudKit operations that use pipelining, like [CKFetchRecordZoneChangesOperation](https://developer.apple.com/reference/cloudkit/ckfetchrecordzonechangesoperation) (where the client code may want to update the previousServerChangeToken for a zone in the event of prior partial success that was persisted).

Hypothetical pseudo-code example (doesn't work yet because support for the new CloudKit operations isn't yet included):
```swift
var recordsChanged: [CKRecord] = []
var recordIDsDeleted: [CKRecordID] = []

let zone1 = CKRecordZoneID(zoneName: "zone1", ownerName: CKCurrentUserDefaultName)
let zone2 = CKRecordZoneID(zoneName: "zone2", ownerName: CKCurrentUserDefaultName)
var currentChangeTokens: [CKRecordZoneID: CKServerChangeToken] = [zone1: nil, zone2: nil] // fresh/first download of changes
var zoneCompleted: [CKRecordZoneID: Bool] = [zone1: false, zone2: false]

let zone1Options = CKFetchRecordZoneChangesOptions()
let zone2Options = CKFetchRecordZoneChangesOptions()
zone1Options.previousServerChangeToken = currentChangeTokens[zone1]
zone2Options.previousServerChangeToken = currentChangeTokens[zone2]
let zones = [zone1, zone2]

let operation: CloudKitOperation<CKFetchRecordZoneChangesOperation> = CloudKitOperation { CKFetchRecordZoneChangesOperation() }

operation.recordZoneIDs = zones
operation.optionsByRecordZoneID = [zone1: zone1Options, zone2: zone2Options]

operation.recordChangedBlock = { record in
    recordsChanged.append(record)
}
operation.recordWithIDWasDeletedBlock = { recordID in
    recordIDsDeleted.append(recordID)
}
operation.recordZoneChangeTokensUpdatedBlock = { (recordZoneID, serverChangeToken, clientChangeTokenData) in
    currentChangeTokens.updateValue(serverChangeToken, forKey: recordZoneID)
}
operation.recordZoneFetchCompletionBlock = { (recordZoneID, serverChangeToken, clientChangeTokenData, moreComing, recordZoneError) in
    currentChangeTokens.updateValue(serverChangeToken, forKey: recordZoneID)
    zoneCompleted.updateValue(true, forKey: recordZoneID)
    // handle error if needed here, but shouldn't be required - error should bubble up to an operationError in fetchRecordZoneChangesCompletionBlock
    // moreComing is always false unless fetchAllChanges is set to false (fetchAllChanges default is true)
}
operation.fetchRecordZoneChangesCompletionBlock = {
    // if reached here, success - handle the fact that the fetch of all changes is complete
}

operation.setFinallyConfigureRetryOperationBlock { nextOperation in
    // only retry with zones that haven't completed
    // and ensure that the per-zone previousServerChangeToken is the last-processed one
    let incompleteZones = zones.filter { (recordZoneID) -> Bool in
        return !(zoneCompleted[recordZoneID] ?? false)
    }
    nextOperation.recordZoneIDs = incompleteZones

    var optionsForIncompleteZones: [CKRecordZoneID: CKFetchRecordZoneChangesOptions] = [:]
    for zone in incompleteZones {
        let options = CKFetchRecordZoneChangesOptions()
        options.previousServerChangeToken = currentChangeTokens[zone]
        optionsForIncompleteZones[zone] = options
    }
    nextOperation.optionsByRecordZoneID = optionsForIncompleteZones
}
```

## BatchedCloudKitOperation:

### **BatchedCloudKitOperation.setConfigureNextOperationBlock()**:

Set a block that can configure the next internal CloudKitOperation (i.e. if moreComing == true)
The operation passed-into the block will be the next operation, **already configured like the previous operation**. Use this block to make any additional required changes.

For example, if the "next" operation needs to use the updated serverChangeToken:
```swift
batchedOperation.setConfigureNextOperationBlock { nextOperation in
    nextOperation.previousServerChangeToken = currentServerChangeToken // <-- (retrieved/persisted elsewhere)
}
```

swiftlyfalling:
I also added some documentation for BatchedCloudKitOperation.

swiftlyfalling:
For clarity on the order of operations (in case anyone is curious about how CloudKitOperation works):

When a **CloudKitOperation** experiences an error*:

1. If a custom error handler for the specific error code was set by the client (i.e. by calling `setErrorHandlerForCode` on the CloudKitOperation), that custom error handler will be called.
2. Else, if a default error handler for the specific error code is set (in `class CloudKitRecovery`), that default error handler will be called.
3. **If no error handler is set for the specific error code, the CloudKitOperation finishes with the error.**
4. If an error handler is called, it is provided with:
 - the current underlying `CKOperation` that errored
 - the `error`
 - a `log` variable (for the purpose of logging diagnostic information)
 - a `suggested` suggestion of (delay & configure block);
   - the `suggested.configure` configure block is the last CKOperation's configure block
5. The error handler returns an optional tuple of (delay & configure block) as the `response`.
6. **If the error handler returns `nil`, the CloudKitOperation finishes with the error.**
7. **If the error handler returns a (delay & configure block)**, it indicates that the CKOperation **should be retried after the specified delay, and configured with the specified configure block.**
8. **After** the new (retry) operation is configured **by the configuration block returned by the error handler**, the retry operation is passed to the **`finallyConfigureRetryOperationBlock`** (if one was set by a call to `CloudKitOperation.setFinallyConfigureRetryOperationBlock`), which has a chance to further configure it.

> \* Error handling/recovery requires setting the CKOperation-specific completion block on the CloudKitOperation. See the CloudKitOperation documentation.

Sources/Features/Shared/CloudKit/CloudKitOperation.swift
```diff
@@ -279,10 +324,106 @@ class CloudKitOperationGenerator<T where T: NSOperation, T: CKOperationType, T:
...
+ If you need to handle the completion of the BatchedCloudKitOperation itself, use a Will/DidFinishObserver.
+ 
+ ### Error Handling
+ 
+ See the documentation for CloudKitOperation Error Handling.
+ 
+ BatchedCloudKitOperation provides a convenience setErrorHandlerForCode method, which sets the error
+ handling for every CloudKitOperation in the batch.
+ 
+ */
```
danthorpe: This is awesome.

PR #405, opened 2016-08-17
Add base types to ProcedureKit

danthorpe:
The goal here is to slowly copy/rename/Swift 3 sanity check files from _Operations_ to _ProcedureKit_.

## `Procedure`, `ProcedureQueue`
- [x] Basic state transitions
- [x] Execution
- [x] Cancellation with errors
- [ ] Finishing with errors

## `ProcedureKitTestCase`
- [x] `TestProcedure`
- [x] `QueueTestDelegate`
- [x] `run(procedure: Procedure)` API
- [x] `wait(for: Procedure)` API


Sources/ProcedureObserver.swift
```diff
@@ -0,0 +1,148 @@
...
+}
+
+public extension ProcedureObserver {
+
+    /**
+     Default implementation does nothing.
+
+     - parameter procedure: the observed `Procedure`.
+     */
+    func didAttachTo(procedure: Procedure) { }
```
jshier: I think a more Swift 3 name would be `didAttach(to procedure: Procedure)`, so the use site would look like `didAttach(to: procedure)`. Also, for the other protocol names, it might be more fitting for them to drop the `procedure` external parameter label. So `willExecute(procedure: Procedure)` would be come `willExecute(_ procedure: Procedure)`, as that would allow usage like `willExecute(procedure)`.

danthorpe: Yeah - you're right, will revise it. Thanks @jshier.

Sources/ProcedureObserver.swift
```diff
@@ -0,0 +1,148 @@
...
+     - parameter procedure: the observed `Procedure`.
+     */
+    func didCancel(procedure: Procedure)
+}
+
+/**
+ Types which conform to this protocol, can be attached to `Procedure` subclasses. They 
+ will receive a callback when the Procedure produces another operation.
+ */
+public protocol DidProduceOperationProcedureObserver: ProcedureObserver {
```
swiftlyfalling: @danthorpe: Shouldn't this be `DidProduceProcedureObserver`?

swiftlyfalling: This is a bit of a mouthful. Thoughts on something like `DidProduceOperationObserver` instead?

danthorpe: Yeah, I thought about that. Maybe.

But here, the procedure has just produced an (NS)Operation. Not necessarily a procedure. So thought it sounded misleading.

danthorpe: And I liked how they were all ActionProcedureObserver

danthorpe: I think also, the thing here is that the block based type which conforms to this protocol should be called `DidProduceOperationObserver` - as this is what will most likely be used.

I'm gonna add the block types now, and we'll see how it looks.

swiftlyfalling: Ah, yeah. Agreed that the more-commonly-used piece (the block-based observer) should take the `DidProduceOperationObserver` name.

Issue #416, opened 2016-08-24
Access to deinited object


pomozoff:
Here is a [screenshot](https://yadi.sk/i/inf9gH9fuSZ9B) of the crash.
**Self** is already destroyed but it marked as **unowned** in the block and code crashes trying to access **self** in the line #545.

Should the **self** will be marked as **weak**?

danthorpe:
Yeah, it can be weak - the evaluator is owned by the operation, so, `unowned` should be okay - but there may be some odd async edge cases on de-init I suppose. You are probably correct that `weak` and then unwrapping is safer.

I will fix this in the `swift/2.2` branch.

pomozoff:
Thanks.

PR #417, opened 2016-08-24
Fix crash when accessing destroyed self object #416

pomozoff:
This fix works for me, but I understand there may be more complicated issue and it should be investigated.
Unfortunately I've a little time to explore whole project to learn how it works.
I'll try to make a test case to reveal this issue.


danthorpe:
@pomozoff - can you pull `swift/2.2` into your branch please?

danthorpe:
@pomozoff - thanks for this! It'll go into 3.1.2 later this week I expect :)

pomozoff:
Sorry for delay, looks like it works fine now - no crashes yet.

Issue #418, opened 2016-08-24
URLSessionOperation completion handler executes before Task is finished.


S0MMS:
There seems to be an issue with URLSessionOperation.  If the NSURLSessionTask that the URLSessionOperation is initialized with has a completion handler **AND** the URLSessionOperation itself has a completion handler, the completion handler for the URLSessionOperation can execute **BEFORE** the task completion handler.  The following code produces the issue...

    func testURLSessionTaskOperation() {
        let url = NSURL(string: "https://www.google.com")
        let request = NSURLRequest(URL: url!)
        
        let task : NSURLSessionTask = NSURLSession.sharedSession().dataTaskWithRequest(request, completionHandler: {(data: NSData?, response: NSURLResponse?, error: NSError?) in
            for i in 0..<100000 {
                var foo:String = "foo \(i)"
                foo = foo.stringByAppendingString(foo)
            }
            print("Task done")
        })
        
        let sessionTaskOperation = URLSessionTaskOperation.init(task: task);
        sessionTaskOperation.completionBlock = {
            print("Operation done")
        }
        
        let queue = OperationQueue()
        queue.addOperations(sessionTaskOperation);
        
        sleep(10)
    }

Running this test will yield output of... 
'Operation done' 
'Task done'

The expected result should be (in my opinion)...
'Task done'
'Operation done'

danthorpe:
Thanks @S0MMS - will look into this.

Unfortunately, the `completionBlock` is `NSOperation` is very unreliable - especially when it comes to deterministic behaviour.

Can you retry your test using a `DidFinishObserver` instead:

```swift
let sessionTaskOperation = URLSessionTaskOperation.init(task: task);
sessionTaskOperation.addObserver(DidFinishObserver { op, errors in
    print("Operation done")
})
```

Thanks!

S0MMS:
Dan,

Thanks for getting back - I hear you on the completionBlock being unreliable.  I changed my test to the following...

    func testURLSessionTaskOperation() {
        let url = NSURL(string: "https://www.google.com")
        let request = NSURLRequest(URL: url!)
        
        let task : NSURLSessionTask = NSURLSession.sharedSession().dataTaskWithRequest(request, completionHandler: {(data: NSData?, response: NSURLResponse?, error: NSError?) in
            // do some stuff
            for i in 0..<100000 {
                var foo:String = "foo \(i)"
                foo = foo.stringByAppendingString(foo)
            }
            print("Task done")
        })
[...]

Issue #340, opened 2016-06-09
Operation which depends on the network operation


razalur:
Hello!

I have one more error with operations usage
So I have an operation which download data and then another operation parse this data.
But sometimes I have a crash on parsing data cause the data to parse is nil. It seems that sometimes parsing operation starts earlier than the downloading operation finishes. Is it a possibility to find out the reason of such crash? Or may be you can tell me why so by looking at my code?

```
class GetAndSaveAuthorOperation: GroupOperation {
    init(guid: String, context: NSManagedObjectContext) {
        let url = NSURL(string: GET_AUTHOR_SERVLET_URL)!
        let authorRequest = NSMutableURLRequest(URL: url)
        authorRequest.HTTPMethod = "POST"
        authorRequest.HTTPBody = String("\(POST_PARAM_GUID)=\(guid)").dataUsingEncoding(NSUTF8StringEncoding)
        
        let downloadAuthor = DownloadDataOperation(request: authorRequest)
        
        let parseAuthor = ParseAuthorOperation(mainContext: context)
        parseAuthor.injectResultFromDependency(downloadAuthor)
        
        super.init(operations: [parseAuthor, downloadAuthor])
    }
}

class DownloadDataOperation: GroupOperation, ResultOperationType {
    let request: NSMutableURLRequest
    private(set) var result: NSData?

    init(request: NSMutableURLRequest) {
        self.request = request
  
        super.init(operations: [])
        
        let task = NSURLSession.sharedSession().dataTaskWithRequest(request) { data, response, error in
            if let error = error {
                print("Can't download! Cause ...")
            }
            else {
                self.result = data
            }
        }

        let taskOperation = URLSessionTaskOperation(task: task)
        self.addOperation(taskOperation)
    }
}

class ParseAuthorOperation: Operation, AutomaticInjectionOperationType {
    var requirement: NSData?

    let mainContext: NSManagedObjectContext
    var privateContext: NSManagedObjectContext!
   
    init(mainContext: NSManagedObjectContext) {
        self.mainContext = mainContext
        
        super.init()
    }
    
    override func execute() {
        guard !cancelled else { return }
        
        self.privateContext = NSManagedObjectContext(concurrencyType: .PrivateQueueConcurrencyType)
        self.privateContext.parentContext = self.mainContext
        
        do {
            /// Here I have a crash sometimes, cause requirement is nil
[...]

Issue #320, opened 2016-05-18
NSURLSessionTask state property KVO may be inadvisable


itsthejb:
Hi Dan,

So looks like we may have a relevant issue here, although AFAIK it's not public yet on our code (FYI we previously used background `NSURLSession`s only, but might be forced to move to mostly foreground due to some odd queue behaviour when using background).

In any case, the issue itself is discussed at length wrt AFNetworking at https://github.com/AFNetworking/AFNetworking/issues/1477 and https://devforums.apple.com/thread/210597. The tl;dr seems to be that using KVO on `NSURLSessionTask`'s `state` property in `URLSessionTaskOperation` is problematic.

There are various work-arounds suggested on the thread. I'll play with a few and get back to you.

Cheers

itsthejb:
And here's an example stack track:

```
  thread #1: tid = 0x6dd5, 0x248446ac libsystem_kernel.dylib`pread + 20, queue = 'NSPersistentStoreCoordinator 0x14ef6020'
    frame #0: 0x248446ac libsystem_kernel.dylib`pread + 20
    frame #1: 0x25057066 libsqlite3.dylib`___lldb_unnamed_function356$$libsqlite3.dylib + 206
    frame #2: 0x24ff3b9c libsqlite3.dylib`___lldb_unnamed_function25$$libsqlite3.dylib + 24
    frame #3: 0x2500c724 libsqlite3.dylib`___lldb_unnamed_function77$$libsqlite3.dylib + 120
    frame #4: 0x2500babe libsqlite3.dylib`___lldb_unnamed_function72$$libsqlite3.dylib + 718
    frame #5: 0x25027892 libsqlite3.dylib`___lldb_unnamed_function116$$libsqlite3.dylib + 402
    frame #6: 0x2502df52 libsqlite3.dylib`___lldb_unnamed_function155$$libsqlite3.dylib + 242
    frame #7: 0x2501cac4 libsqlite3.dylib`___lldb_unnamed_function114$$libsqlite3.dylib + 1428
    frame #8: 0x2501ba50 libsqlite3.dylib`sqlite3_step + 472
    frame #9: 0x266bacee CoreData`_execute + 114
    frame #10: 0x266baa40 CoreData`-[NSSQLiteConnection execute] + 1804
    frame #11: 0x267c1222 CoreData`newFetchedRowsForFetchPlan_MT + 1038
[...]

PR #426, opened 2016-08-26
URLSessionTaskOperation rework proposal

swiftlyfalling:
**Note: This PR is unfinished, and is a work-in-progress.**

There are several issues surrounding the current structure of URLSessionTaskOperation.

- #418 URLSessionOperation completion handler executes before Task is finished.
- #320 NSURLSessionTask state property KVO may be inadvisable
- #340 Operation which depends on the network operation
etc.

Most of these issues are caused by the lack of consistent ordering and timing surrounding the completion of the URLSessionOperation and the completion of its NSURLSessionTask.

The culprit? KVO to determine when the NSURLSessionTask has completed.

This proposal restructures URLSessionTaskOperation to be constructed with static factory methods that create the task and set a completionHandler with predictable ordering and behavior. (i.e. You now construct the URLSessionTaskOperation and the NSURLSessionTask at the same time.)

This means that code like:

```swift
let task = session.dataTaskWithURL(url) { (data, response, error) in
    // do something with the data
}
let operation = URLSessionTaskOperation(task: task)
```

becomes:

```swift
let operation = URLSessionTaskOperation.dataTask(fromSession: session, withURL: url) { (data, response, error) in
    // do something with the data
}
// to access the task directly, use operation.task
```

More importantly, it means that we can ensure that the URLSessionTaskOperation itself finishes **after** the user's completionHandler has returned.

-----

Needs factory methods for the following methods of NSURLSession:

Adding Data Tasks to a Session
- [ ] dataTaskWithURL(_:)
- [x] dataTaskWithURL(_:completionHandler:)
- [ ] dataTaskWithRequest(_:)
- [x] dataTaskWithRequest(_:completionHandler:)

Adding Download Tasks to a Session
- [ ] downloadTaskWithURL(_:)
- [x] downloadTaskWithURL(_:completionHandler:)
- [ ] downloadTaskWithRequest(_:)
- [x] downloadTaskWithRequest(_:completionHandler:)
- [ ] downloadTaskWithResumeData(_:)
- [ ] downloadTaskWithResumeData(_:completionHandler:)

Adding Upload Tasks to a Session
- [ ] uploadTaskWithRequest(_:fromData:)
- [ ] uploadTaskWithRequest(_:fromData:completionHandler:)
- [ ] uploadTaskWithRequest(_:fromFile:)
- [ ] uploadTaskWithRequest(_:fromFile:completionHandler:)
- [ ] uploadTaskWithStreamedRequest(_:)

-----

This is complicated by the methods without completion handlers.

It may be possible to safely use KVO for tasks created without completion handlers since there is no user completion handler to worry about (timing-wise), and the documentation for NSURLSessionTask state says:

> Completed
> The task has completed (without being canceled), and the **task's delegate receives no further callbacks**.

Thus, KVO may be safe in this case. (For these tasks, the consumer relies on receiving all data as delegate callbacks.)

### DIFF
--- a/Sources/Features/Shared/URLSessionTaskOperation.swift
+++ b/Sources/Features/Shared/URLSessionTaskOperation.swift
@@ -18,15 +18,24 @@ setup, so that conditions or observers can be attached.
 
 */
 public class URLSessionTaskOperation: Operation {
-
-    enum KeyPath: String {
-        case State = "state"
-    }
-
     public private(set) var task: NSURLSessionTask!
 
+    private override init() {
+        super.init()
+    }
+
+    // Deprecated init
+    @available(*, unavailable, message="URLSessionTaskOperation has been refactored to utilize factory methods that create the task and the operation.")
+    public init(task: NSURLSessionTask) {
+        super.init()
+    }
+
+    // MARK: - Factory Methods
+
+    // MARK: -- Data Tasks
+
     public static func dataTask(fromSession session: NSURLSession, withURL url: NSURL, completionHandler: (NSData?, NSURLResponse?, NSError?) -> Void) -> URLSessionTaskOperation {
-        let op = URLSessionTaskOperation()
+        let op = URLSessionTaskOperationWithCompletionHandler()
         let task = session.dataTaskWithURL(url) { (data, response, error) in
             completionHandler(data, response, error)
             op.finish(error)
@@ -40,7 +49,7 @@ public class URLSessionTaskOperation: Operation {
     }
 
     public static func dataTask(fromSession session: NSURLSession, withRequest request: NSURLRequest, completionHandler: (NSData?, NSURLResponse?, NSError?) -> Void) -> URLSessionTaskOperation {
-        let op = URLSessionTaskOperation()
+        let op = URLSessionTaskOperationWithCompletionHandler()
         let task = session.dataTaskWithRequest(request) { (data, response, error) in
             completionHandler(data, response, error)
             op.finish(error)
@@ -52,9 +61,35 @@ public class URLSessionTaskOperation: Operation {
         })
         return op
     }
+    
+    // without completion handler
+
+    public static func dataTask(fromSession session: NSURLSession, withURL url: NSURL) -> URLSessionTaskOperation {
+        let op = URLSessionTaskOperationWithoutCompletionHandler(session: session)
+        let task = session.dataTaskWithURL(url)
+        assert(task.state == .Suspended, "NSURLSessionTask must be suspended, not \(task.state)")
+        op.task = task
+        op.addObserver(DidCancelObserver { _ in
+            task.cancel()
+        })
+        return op
+    }
+
+    public static func dataTask(fromSession session: NSURLSession, withRequest request: NSURLRequest) -> URLSessionTaskOperation {
+        let op = URLSessionTaskOperationWithoutCompletionHandler(session: session)
+        let task = session.dataTaskWithRequest(request)
+        assert(task.state == .Suspended, "NSURLSessionTask must be suspended, not \(task.state)")
+        op.task = task
+        op.addObserver(DidCancelObserver { _ in
+            task.cancel()
+        })
+        return op
+    }
+
+    // MARK: -- Download Tasks
 
     public static func downloadTask(fromSession session: NSURLSession, withURL url: NSURL, completionHandler: (NSURL?, NSURLResponse?, NSError?) -> Void) -> URLSessionTaskOperation {
-        let op = URLSessionTaskOperation()
+        let op = URLSessionTaskOperationWithCompletionHandler()
         let task = session.downloadTaskWithURL(url) { (location, response, error) in
             completionHandler(location, response, error)
             op.finish(error)
@@ -68,7 +103,7 @@ public class URLSessionTaskOperation: Operation {
     }
 
     public static func downloadTask(fromSession session: NSURLSession, withRequest request: NSURLRequest, completionHandler: (NSURL?, NSURLResponse?, NSError?) -> Void) -> URLSessionTaskOperation {
-        let op = URLSessionTaskOperation()
+        let op = URLSessionTaskOperationWithCompletionHandler()
         let task = session.downloadTaskWithRequest(request) { (location, response, error) in
             completionHandler(location, response, error)
             op.finish(error)
@@ -81,23 +116,88 @@ public class URLSessionTaskOperation: Operation {
         return op
     }
 
-    private override init () {
-        super.init()
-    }
+    // without completion handler
 
-    @available(*, unavailable, message="URLSessionTaskOperation has been refactored to utilize factory methods that create the task and the operation.")
-    public init(task: NSURLSessionTask) {
+    public static func downloadTask(fromSession session: NSURLSession, withURL url: NSURL) -> URLSessionTaskOperation {
+        let op = URLSessionTaskOperationWithoutCompletionHandler(session: session)
+        let task = session.downloadTaskWithURL(url)
         assert(task.state == .Suspended, "NSURLSessionTask must be suspended, not \(task.state)")
-        self.task = task
-        super.init()
-        addObserver(DidCancelObserver { _ in
+        op.task = task
+        op.addObserver(DidCancelObserver { _ in
             task.cancel()
         })
+        return op
     }
 
-    public override func execute() {
+    public static func downloadTask(fromSession session: NSURLSession, withRequest request: NSURLRequest) -> URLSessionTaskOperation {
+        let op = URLSessionTaskOperationWithoutCompletionHandler(session: session)
+        let task = session.downloadTaskWithRequest(request)
+        assert(task.state == .Suspended, "NSURLSessionTask must be suspended, not \(task.state)")
+        op.task = task
+        op.addObserver(DidCancelObserver { _ in
+            task.cancel()
+        })
+        return op
+    }
+}
+
+// For URLSessionTasks with a completion handler, we can properly handle finishing the operation 
+// by utilizing the completion handler. So this is pretty simple:
+internal class URLSessionTaskOperationWithCompletionHandler: URLSessionTaskOperation {
+    override func execute() {
         assert(task.state == .Suspended, "NSURLSessionTask resumed outside of \(self)")
         task.resume()
     }
 }
 
+// For URLSessionTasks *without* a completion handler, we must: 
+// 1.) Use KVO to determine when the task is finished
+// 2.) Dispatch the call to `finish()` to the NSURLSession's delegateQueue
+internal class URLSessionTaskOperationWithoutCompletionHandler: URLSessionTaskOperation {
+
+    enum KeyPath: String {
+        case State = "state"
+    }
+
+    private var removedObserved = false
+    private let lock = NSLock()
+    private var sessionDelegateQueue: NSOperationQueue
+    
+    init(session: NSURLSession) {
+        self.sessionDelegateQueue = session.delegateQueue
+        super.init()
+    }
+
+    internal override func execute() {
+        assert(task.state == .Suspended, "NSURLSessionTask resumed outside of \(self)")
+        task.addObserver(self, forKeyPath: KeyPath.State.rawValue, options: [], context: &URLSessionTaskOperationKVOContext)
+        task.resume()
+    }
+
+    internal override func observeValueForKeyPath(keyPath: String?, ofObject object: AnyObject?, change: [String : AnyObject]?, context: UnsafeMutablePointer<Void>) {
+        guard context == &URLSessionTaskOperationKVOContext else { return }
+
+        lock.withCriticalScope {
+            if object === task && keyPath == KeyPath.State.rawValue && !removedObserved {
+
+                if case .Completed = task.state {
+                    sessionDelegateQueue.addOperation(NSBlockOperation { [unowned self] in
+                        self.finish(self.task.error)
+                    })
+                }
+
+                switch task.state {
+                case .Completed, .Canceling:
+                    task.removeObserver(self, forKeyPath: KeyPath.State.rawValue)
+                    removedObserved = true
+                default:
+                    break
+                }
+            }
+        }
+    }
+}
+
+// swiftlint:disable variable_name
+private var URLSessionTaskOperationKVOContext = 0
+// swiftlint:enable variable_name
--- a/Sources/Features/Shared/URLSessionTaskOperation.swift
+++ b/Sources/Features/Shared/URLSessionTaskOperation.swift
@@ -23,11 +23,69 @@ public class URLSessionTaskOperation: Operation {
         case State = "state"
     }
 
-    public let task: NSURLSessionTask
+    public private(set) var task: NSURLSessionTask!
 
-    private var removedObserved = false
-    private let lock = NSLock()
+    public static func dataTask(fromSession session: NSURLSession, withURL url: NSURL, completionHandler: (NSData?, NSURLResponse?, NSError?) -> Void) -> URLSessionTaskOperation {
+        let op = URLSessionTaskOperation()
+        let task = session.dataTaskWithURL(url) { (data, response, error) in
+            completionHandler(data, response, error)
+            op.finish(error)
+        }
+        assert(task.state == .Suspended, "NSURLSessionTask must be suspended, not \(task.state)")
+        op.task = task
+        op.addObserver(DidCancelObserver { _ in
+            task.cancel()
+        })
+        return op
+    }
 
+    public static func dataTask(fromSession session: NSURLSession, withRequest request: NSURLRequest, completionHandler: (NSData?, NSURLResponse?, NSError?) -> Void) -> URLSessionTaskOperation {
+        let op = URLSessionTaskOperation()
+        let task = session.dataTaskWithRequest(request) { (data, response, error) in
+            completionHandler(data, response, error)
+            op.finish(error)
+        }
+        assert(task.state == .Suspended, "NSURLSessionTask must be suspended, not \(task.state)")
+        op.task = task
+        op.addObserver(DidCancelObserver { _ in
+            task.cancel()
+        })
+        return op
+    }
+
+    public static func downloadTask(fromSession session: NSURLSession, withURL url: NSURL, completionHandler: (NSURL?, NSURLResponse?, NSError?) -> Void) -> URLSessionTaskOperation {
+        let op = URLSessionTaskOperation()
+        let task = session.downloadTaskWithURL(url) { (location, response, error) in
+            completionHandler(location, response, error)
+            op.finish(error)
+        }
+        assert(task.state == .Suspended, "NSURLSessionTask must be suspended, not \(task.state)")
+        op.task = task
+        op.addObserver(DidCancelObserver { _ in
+            task.cancel()
+        })
+        return op
+    }
+
+    public static func downloadTask(fromSession session: NSURLSession, withRequest request: NSURLRequest, completionHandler: (NSURL?, NSURLResponse?, NSError?) -> Void) -> URLSessionTaskOperation {
+        let op = URLSessionTaskOperation()
+        let task = session.downloadTaskWithRequest(request) { (location, response, error) in
+            completionHandler(location, response, error)
+            op.finish(error)
+        }
+        assert(task.state == .Suspended, "NSURLSessionTask must be suspended, not \(task.state)")
+        op.task = task
+        op.addObserver(DidCancelObserver { _ in
+            task.cancel()
+        })
+        return op
+    }
+
+    private override init () {
+        super.init()
+    }
+
+    @available(*, unavailable, message="URLSessionTaskOperation has been refactored to utilize factory methods that create the task and the operation.")
     public init(task: NSURLSessionTask) {
         assert(task.state == .Suspended, "NSURLSessionTask must be suspended, not \(task.state)")
         self.task = task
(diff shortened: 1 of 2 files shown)


danthorpe:
Overall, I quite like the static method approach. :+1:

However, I'm thinking that there must be a better way to reduce the amount of code via protocols, generics and retroactive modelling.

Couple of code smells:
1. the _request_ version is basically the same as the _url_ version, 
2. the _data_ version is basically the same as the _download_ version.

So, something like:

```swift
public protocol SessionTaskProtocol {
    associatedtype Response 
}

public extension NSURLSessionDataTask: SessionTaskProtocol {
    typealias Response = (NSData?, NSURLResponse?, NSError?)
}

public extension NSURLSessionDownloadTask: SessionTaskProtocol {
    typealias Response = (NSData?, NSURLResponse?, NSError?)
}

public protocol URLSessionProtocol {
    func task<SessionTask: SessionTaskProtocol>(withRequest request: NSURLRequest, completion: SessionTask.Response -> Void) -> SessionTask
}

public extension URLSessionProtocol {
    func task<SessionTask: SessionTaskProtocol>(withURL url: NSURL, completion: SessionTask.Response -> Void) -> SessionTask {
        return task<SessionTask>(withRequest: NSURLRequest(URL: url), completion: completion)
    }
}

public extension NSURLSession: URLSessionProtocol {
    func task<SessionTask: SessionTaskProtocol>(withRequest request: NSURLRequest, completion: SessionTask.Response -> Void) -> SessionTask {
          return session.dataTaskWithRequest(request, completion: completion)
    }

    func task<SessionTask: SessionTaskProtocol where SessionTask: NSURLSessionDownloadTask>(withRequest request: NSURLRequest, completion: SessionTask.Response -> Void) -> SessionTask {
          return session.downloadTaskWithRequest(request, completion: completion)
    }
}

public class NSURLSessionTaskOperation {
    public static func task<SessionTask: SessionTaskProtocol where SessionTask: NSURLSessionTask>(fromSession session: NSURLSession, withURL url: NSURL, completionHandler: (NSData?, NSURLResponse?, NSError?) -> Void) -> SessionTask {

    let op = URLSessionTaskOperation()
    let task: SessionTask = session.task(withRequest: request) { response in
        completion(response)
        op.finish()
    }

    // etc etc

    }
}
```

Main idea behind this is to use the return type to indicate what API on NSURLSession to use. Then use a protocol extension so that we get the URL version "for free".

Having said all that - I don't want to re-write Alamofire....

danthorpe:
Also, I'm thinking that this is a pretty significant and breaking change - so we won't release it as part of v3.1.* - and I wasn't planning on doing a v3.2 - which means that we can put a pin in it and pick it up with v4 and write it in Swift 3?

swiftlyfalling:
Is calling `dataTaskWithRequest(NSURLRequest(URL: url))` equivalent to what NSURLSession does when you call `NSURLSession.dataTaskWithURL(url)`?

That'd be my first question.

swiftlyfalling:
For future reference:

- Handling NSURLSessionTasks with completion handlers is simple. KVO is not necessary (if we're creating the tasks ourselves with static factory methods).
- Handling NSURLSessionTasks that lack completion handlers seems like it will still require KVO.
    - Testing seems to indicate that dispatching the call to `self.finish()` on the NSURLSession's `delegateQueue` ensures that it occurs after the final NSURLSessionTaskDelegate callback occurs.

malhal:
Here is a better redesign (inspired by CloudKit's API):

1. make the completion handler a property rather than passed in to the constructor. If you want to continue to support data task and download task in the same operation then you need two.
2. always create the task with a completion block and remove KVO.
3. set the created task to an ivar (bug in latest commit does let task).
4. within the task's completion handler save the data in an ivar and call finish (for data task) or save location (for download task).
5. within the operationDidFinish method check the class of the task ivar and call the respective completion handler with the data ivar, task.response, and task.error (for data task) or the location one (for download task).
6. remove the session param from all the constructors and make it a property, which returns shared session if not set.

This allows you to remove all the KVO and you only need the constructors for URL and request. I would suggest just calling the request constructor from the URL one with request with URL rather than duplicating the code.

The final use ends up looking like this:

```
URLSessionTaskOperation* op = [[URLSessionTaskOperation alloc] initWithURLRequest:self.request];
// op.session = myCustomSession; //optional
    [op setRequestCompletionBlock:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
        NSLog(@"complete");
    }];
    [queue addOperation:op];
```

hope you like it :)

danthorpe:
@malhal - thanks for the input, all good points! 😀 Following the `CKOperation` strategy makes a lot of sense.

malhal:
Thanks @danthorpe, whoever designed `CKOperation` deserves a medal it is absolute quality! 

So I was intrigued by the separate classes design I mentioned, so decided to investigate it myself and you can see what I came up with here:

[MHFURLSessionTaskOperation.h](https://github.com/malhal/MHFoundation/blob/master/MHFoundation/MHFURLSessionTaskOperation.h
)

[MHFURLSessionTaskOperation.m](https://github.com/malhal/MHFoundation/blob/master/MHFoundation/MHFURLSessionTaskOperation.m)

By the way, the shouldRun and thread-safe callback queue technique (especially the custom completion handler capability) were all reversed from `CKOperation` which give some incredible power and stability.

swiftlyfalling:
@malhal, @danthorpe: 

The behavior of `NSURLSessionTasks` created with completion handlers and without completion handlers differs significantly.

Tasks created **with** a completion handler bypass calls to several `NSURLSession` delegate methods related to response and data delivery.

Therefore, **it is not safe to hide use of the completion handler as an implementation detail of the Operation subclass, nor use it for "non-completionHandler" cases**.

In your example, if someone were to do:

    URLSessionDataTaskOperation* op = [[URLSessionDataTaskOperation alloc] initWithURLRequest:self.request];
    // op.session = myCustomSession; //optional
    // do not set the completionHandler, because the response is handled through delegate callbacks
    [queue addOperation:op];

They would not receive all of the expected `NSURLSession` delegate callbacks, **despite having deliberately not provided a completionHandler** (_which is the only method to differentiate between these behaviors when using `NSURLSession`_).

For tasks created with a completionHandler, as I think we all agree, the solution is simple and does not require KVO - the class just has to have ultimate responsibility for actually initializing the task (which lets it insert its own call to `finish` in the `completionHandler`).

But silently using a completionHandler does not solve the other case. **It breaks expected behavior.**

Now, I'm fine with not supporting the "no-`completionHandler`" case in the replacement for `URLSessionTaskOperation`. But then we should be explicit, and **require** the `completionHandler` parameter in the initializer. (Not as an optional property.)

The user can then always pass an empty `completionHandler` block, but this makes the contract clear: **you are receiving the behavior of an NSURLSessionTask created with a completionHandler**.

malhal:
@swiftlyfalling that is really interesting, I did wonder why anyone would want to create a request with a url in a fire-and-forget fashion. This certainly gives a lot to think about!

danthorpe:
@malhal @swiftlyfalling and we can't even do anything fancy like conform to `NSURLSessionDelegate` and intercept those callbacks (before forwarding them to the original delegate) because the framework consumer can set the delegate on the session after creating the operation.

All of these reasons (and more) is why support for `NSURLSession` APIs inside _Operations_ has always been minimal, and I expect that most framework consumers should be implementing their own `Operation` subclasses.

malhal:
So maybe that's a good reason for only supporting the tasks that take a completion handler?

Since I don't use swift I'm not sure about making it a compulsory init param tho, with operations I prefer always having optional params and properties, so things can be set depending on the result of a previous operation.

swiftlyfalling:
@malhal: Ah yes, I meant to say "when the object is initialized". If using convenience (factory) methods, the completionHandler would also be required (if built-in support is limited to the completionHandler-provided case).

In regards to compile-time enforcement vs. run-time checks:

While we certainly could have an operation that takes a `completionHandler` property, and then check at operation execution time whether it's set (and error if it isn't), that's not how a lot of the built-in operation "wrapper" classes work (as of Operations/ProcedureKit 3.x). In fact, the current `URLSessionTaskOperation` requires a `NSURLSessionTask` in its initializer.

Other functionality in this framework makes that a non-issue for cases like your example, where the completionHandler for an NSURLSessionTask is dependent on the result from a previous operation.

malhal:
@swiftlyfalling so it does, I misread the commit diff because I thought this pull request fixed this. I think the the redesign should take the construction of the task inside this operation.

swiftlyfalling:
@malhal: I think that may be a larger point of discussion that affects many more parts of the framework (since lots of things currently do *not* work that way).

But, again, Operations/ProcedureKit has functionality that makes that a non-issue.

One simple example (and probably not even the best way to achieve this): If you need to create the task depending on the result of a prior operation, make a GroupOperation subclass that has the job of receiving the result (for example, via result injection), and then creates a task & child URLSessionTaskOperation in: `override func execute()`. Things can depend on the GroupOperation, if needed.

malhal:
@swiftlyfalling fair enough about the larger point of discussion.

It is a bit of a disappointment using operations in the way you describe is it is sort of negating what I feel is their best feature, in that you create them all initially and put them in a queue (which could be a group operation, which I understand is actually a serial operation queue) and then full control from then on is handed over to the operation queue. It sounds to me what you are describing is no different from normal (pre-using-operatinos) way of say starting something and then in the completion block doing something else.

Anyway someone can work around this issue just by passing in a dummy param to the constructor and then re-setting it if required, that is if the param isn't readonly!

swiftlyfalling:
@malhal: Quick point of clarification: GroupOperation is, by default, not a serial operation queue. It's a logical grouping of child operations. If you want it to function as a serial operation queue, you'd have to set the GroupOperation's `maxConcurrentOperationCount` to 1.

malhal:
@swiftlyfalling apologies for the slip up, I should have said an operation-queue operation.

It drives me nuts that a serial queue can be implemented either max operation count 1 or add the op with a dependency on the last op in the queue, its the kind of thing that you cannot envisage before-hand which is going to work out best.

swiftlyfalling:
@malhal, here's a more verbose pseudo-code version of my example above:

```swift
class DownloadOperation: GroupOperation, InjectionOperationType {
    var priorResult: Any? = .None
// ...
    override func execute() {
        guard let priorResult = priorResult else { print("Missing result from prior operation"; return }
        let task = session.downloadTaskWithURL(url) { (location, response, error) in
            // does something with the result from a prior operation, for example:
            print(priorResult)
        }
        addOperation(URLSessionTaskOperation(task: task))
        super.execute()
    }
/// ...
}
```

```swift
let firstOperation = FirstOperation() // assumes that this does something and then sets properties we'd like to use in later operations

let downloadOperation = DownloadOperation()
downloadOperation.injectResultFromDependency(firstOperation) { op, dep, errors in
    // TODO: check dependency errors, if you'd like
    op.priorResult = dep.result // inject results from the firstOperation (dep) into the downloadOperation (op)
}

let postDownloadOperation = PostDownloadOperation() // some operation that does something after the download operation is complete
postDownloadOperation.addDependency(downloadOperation)

let queue = OperationQueue()
queue.addOperations([firstOperation, downloadOperation, postDownloadOperation])

```

As you can see, the operations would indeed all be created initially, and then put in a queue. An implementation detail of the DownloadOperation is that it creates the task after the `priorResult` has been injected.

The power of Observers, ResultInjection and GroupOperations in Operations/ProcedureKit is fantastic.

malhal:
@swiftlyfalling thanks for the example, apologies if I'm being naive but I'm just not seeing the need for that though, given you can simply do this:

```
let firstOperation = FirstOperation() 
let downloadOperation = DownloadOperation()

firstOperation.completion = (){
     downloadOperation.somethingItNeeds = firstOperation.result
}

let queue = OperationQueue()
queue.addOperations([firstOperation, downloadOperation])
```

swiftlyfalling:
@malhal: On the topic of serial queue execution:

If you care about the order in which the operations execute (i.e. FIFO), the only safe method is to use `addDependency` on the "prior" operation.

An NSOperationQueue with a `maxConcurrentOperationCount` of 1 only ensures that 1 operation is executing at a time. It does not ensure FIFO ordering. 

Some example cases where FIFO ordering may not occur:
- If some of the operations on the queue have dependencies
- If the `queuePriority` property is set differently on different operations
- If an Operation subclass overrides isReady
- If your Operations use Conditions

malhal:
@swiftlyfalling thanks for that explanation, so addDependency is def a good idea! In my serial queue operation one of the reasons I moved from using a final block operation to [operationCount KVO](https://github.com/malhal/MHFoundation/blob/master/MHFoundation/MHFQueueOperation.m) is because I wanted to make the internal operationQueue publicly accessible but I couldn't guarantee operations wouldn't be added directly, by-passing the group's addOperation method which enforced the dependency. I noticed PerformKit creatively solves that by using the queue delegate to monitor for new operations and adding the dependency there, however that still has a problem that final block operation will not run if the queue gets cancelled. I feel it gives many benefits to end with a "group was cancelled" exception, which by the way, all the CKOperations do ;-)

swiftlyfalling:
@malhal:

> @swiftlyfalling thanks for the example, apologies if I'm being naive but I'm just not seeing the need for that though, given you can simply do this:
> 
> ```
> let firstOperation = FirstOperation() 
> let downloadOperation = DownloadOperation()
> 
> firstOperation.completion = (){
>      downloadOperation.somethingItNeeds = firstOperation.result
> }
> 
> let queue = OperationQueue()
> queue.addOperations([firstOperation, downloadOperation])
> ```
> 

A couple quick points:

1. NSOperation completionBlocks are non-deterministic.
Your completion block is not guaranteed to execute prior to DownloadOperation executing.

2. You could, of course, use one of the deterministic Observers in the ProcedureKit framework, like `WillFinishObserver`. That's what Result Injection does internally, so I decided to use Result Injection for the example.

The rest of this gets back to the point of making requirements a contract:

The `URLSessionTaskOperation` (currently) requires a task to be initialized. This enforces, by contract, the proper behavior: that it needs a task. There is no risk of someone not catching that requirement until later in testing (or after they've shipped their code, if they don't test it well) - they won't be able to write code that creates a `URLSessionTaskOperation` without already having what they need to make the NSURLSessionTask.

You had raised the question of how to handle the need to wait to create the NSURLSessionTask until after a prior operation had succeeded (because its completion block required some data from the prior operation). My example was to demonstrate how this is easily doable with the current framework, while still having the `URLSessionTaskOperation` take its requirements in its initializer.

i.e. The way things currently work, requirements are checked at development/compile-time (instead of throwing errors at runtime), and it is still easy to delay making things until you have the pieces to do so.

malhal:
@swiftlyfalling alright so I suppose CKOperation's design has enforced this deterministic behaviour, which it can do given you can only add CKOperations to the queue (ensuring they always error cancelled), and by using a callback queue to ensure the ordering of completion blocks is before the next operation starting. And it sounds like this is all possible in PerformKit just with a different implementation, all sounds good to me! Thanks for brining me up to speed, I'll need to process all of this to decide if my thoughts on the redesign are still valid which I doubt they are.

swiftlyfalling:
@malhal: Yup, the CloudKit operation-specific completion blocks (example: `fetchRecordsCompletionBlock` for `CKFetchRecordsOperation`) have execution timing guarantees. 

Example (for CKFetchRecordsOperation):
> The blocks you assign to process the fetched records are executed serially on an internal queue managed by the fetch records operation. Your blocks must be capable of executing on a background thread, so any tasks that require access to the main thread must be redirected accordingly.

However, the only thing that is said about the underlying NSOperation `completionBlock` property of CKOperation is:

> If you assign a completion block to the `completionBlock` property of the operation object, the completion block is called after the operation executes and returns its results to you. You can use a completion block to perform housekeeping chores related to the operation, but **do not use it to process the results of the operation itself**. Any completion block you specify should be prepared to handle the failure of the operation to complete its task, whether due to an error or an explicit cancellation.

This is because the same issue applies: `completionBlock` is only guaranteed to execute at some point "after the operation executes and returns its results to you". 
ex. Some point after the `fetchRecordsCompletionBlock` is called and, according to the NSOperation documentation, after the value in the `finished` property of the operation changes to `true`.

Notably, this does not provide a guarantee of executing _before_ anything else. (Or non-concurrently with anything else.)

malhal:
@swiftlyfalling it is great that they clarify that but to be honest I doubt anyone would use it given it has no params and thus no way to process the result. It maybe crossed their mind to throw an unavailable exception if it was set.

danthorpe:
Woah - just catching up... 😁 will add my two cents shortly (if appropriate)!

danthorpe:
Woah - just catching up... 😁 will add my two cents shortly (if appropriate)!

danthorpe:
Woah - just catching up... 😁 will add my two cents shortly (if appropriate)!

danthorpe:
Woah - just catching up... 😁 will add my two cents shortly (if appropriate)!

danthorpe:
@malhal - if you want to contribute your thoughts, happy to look at a separate project? Or just fork, create a branch from `swift/2.2` and submit a PR back to `swift/2.2` 👍

danthorpe:
@malhal a little bit of background on some of the designs/architectures here... _Operations_ (and _ProcedureKit_) offer many ways to Get Stuff Done. I think the Result Injection APIs are probably the nicest and most functional. It means that you can absolutely do this:

```swift
let fetch = Fetch(request: request)
let parse = Parse().requireResultFrom(dependency: fetch)
let store = Store().injectResultFrom(dependency: parse)
queue.add(operations: fetch, parse, store)
```

This is _ProcedureKit_ / v4 style code (currently on `development` branch). And it's very much the style of: set up the operations individually and let the framework inject the requirements as needed. Let the queue do its job: scheduling. This is nice. One draw back of this, is that it will more than likely mean that you need to be able to have default values, or explicitly unwrapped optionals as public settable properties - which is either, not always possible or "a Swift code smell".

So... we may prefer to inject requirements at construction, which, means, unless they are literals, create the operations later. For this: we have groups. 

But ultimately, it's all just dependencies, just the same. And it's really just a case of adopting the most appropriate style.

I think `CKOperation` with its completion blocks probably has lots of implementation details to get the scheduling of that correct. Effectively each one of those is like a `GroupOperation` managing its own internal queue and doing a bunch of stuff to ensure that blocks are schedules (inside `BlockOperation`s) correctly.

Which leads us full circle.... I suggest that people develop their operations and SRP async tasks, do a single thing, and focus on `Requirement` -> `Result` i.e. everything is just mapping. Make it all work in the global queue.

But then... move all of it into the constructor of a `GroupOperation` subclass, and just add the operations to the group's queue. I think this is basically how `CKOperation` works.

For `NSURLSession` because we want the framework consumer to provide the session, request, and callbacks - there isn't really a straightforward way to do this in a general way which provides compile time safety and will work for everyone - at that is largely because the session has a single delegate for all requests - so, it's not separable. And I don't want to end up recreating Alamofire....

malhal:
@danthorpe Would you be interested in moving towards the deterministic behaviour we mentioned in ProcedureKit v5? That being: completion handler ordering with respect to operation order (via internal callback dispatch queue), optional init params with runtime checking, and guaranteed cancel errors. Personally I think it would be an absolutely amazing framework! I thought I saw some people having some threading issues in some of the classes and this kind of solid framework should sort all those out, but you'll know better than me how much of a problem that is.

swiftlyfalling:
@malhal: A couple quick notes:

- ProcedureKit's observers are guaranteed to run before the operation finishes, and before any dependent operations begin. (Will/DidFinishObserver are your friends, and can completely replace completionBlocks.)

- ProcedureKit's Groups can be cancelled, and would seem to fit the behavior you desire. Add operations to a Group, cancel the group, the child operations are cancelled. Will/DidCancelObservers and Will/DidFinishObservers are called on every child. Errors are aggregated or, if you choose, you can cancel with an error (or otherwise customize the error behavior).

danthorpe:
@malhal - are you talking about v5 of `NSURLSessionTaskOperation` or the base class in general?

I think @swiftlyfalling's contributions over the pass few months have fixed many, if not all, threading issues. Now, in v3.3 / v4 the core functionality is deterministic and pretty reliable (see the stress tests). And just to add to what @swiftlyfalling has said above, I think the block based observers in _ProcedureKit_ are pretty good too, you can do this:

```swift
// ProcedureKit v4 on development branch
procedure.addDidFinishBlockObserver { procedure, errors in 
    // etc
}
```

And now, the first argument to these block based observers are correctly typed, which means no longer needing to do this:

```swift
// Operations v3.3 on master branch
myOperations.addObserver(DidFinishBlockObserver { [weak myObserver] operation, errors in 
    // Unwrap the observed operation & check that it is equal to the received one
    guard let myOperation = myOperation where myOperation == operation else { return }

    // etc
})
```

Not sure what will happen in v5. 🤔 Certainly we share the same goals here: _an absolutely amazing framework_* and the balance of compile time vs runtime safety is an evolving situation. 

* where by _amazing_ I mean correct, reliable, performant, flexible, documented and maintained. 😁

danthorpe:
@malhal - are you talking about v5 of `NSURLSessionTaskOperation` or the base class in general?

I think @swiftlyfalling's contributions over the pass few months have fixed many, if not all, threading issues. Now, in v3.3 / v4 the core functionality is deterministic and pretty reliable (see the stress tests). And just to add to what @swiftlyfalling has said above, I think the block based observers in _ProcedureKit_ are pretty good too, you can do this:

```swift
// ProcedureKit v4 on development branch
procedure.addDidFinishBlockObserver { procedure, errors in 
    // etc
}
```

And now, the first argument to these block based observers are correctly typed, which means no longer needing to do this:

```swift
// Operations v3.3 on master branch
myOperations.addObserver(DidFinishBlockObserver { [weak myObserver] operation, errors in 
    // Unwrap the observed operation & check that it is equal to the received one
    guard let myOperation = myOperation where myOperation == operation else { return }

    // etc
})
```

Not sure what will happen in v5. 🤔 Certainly we share the same goals here: _an absolutely amazing framework_* and the balance of compile time vs runtime safety is an evolving situation. 

* where by _amazing_ I mean correct, reliable, performant, flexible, documented and maintained. 😁

danthorpe:
So, I've been working on some stuff which uses _ProcedureKit_ v4 stuff, and in particular, a bit of simple networking.

I was just revisiting this issue to see what the discussion was about, and whether there were any conclusions. A key point here is the phrase from @swiftlyfalling:

> Now, I'm fine with not supporting the "no-completionHandler" case in the replacement for URLSessionTaskOperation. But then we should be explicit, and require the completionHandler parameter in the initializer. (Not as an optional property.)

Which, I fully agree with - coming from a pragmatic standpoint. As this allows us to provide something which 90% of the time (I want to download some `Data`) is what is needed.

To that end see: #525

PR #442, opened 2016-08-29
Add Stress Tests Phase 1

danthorpe:


Sources/Testing/ProcedureKitTestCase.swift
```diff
@@ -76,4 +116,38 @@ open class ProcedureKitTestCase: XCTestCase {
...
+                    block(batch, iteration, overallDispatchGroup)
+                }
+
+                let batchFinishTime = CFAbsoluteTimeGetCurrent()
+                let batchDuration = batchFinishTime - batchStartTime
+                print("\(stressTestName), finished batch: \(batch), in \(batchDuration) seconds")
+            }
+        }
+
+        overallDispatchGroup.notify(queue: .main) {
```
swiftlyfalling: The way that most of the current StressTests work is to utilize a dispatch group for each batch, and wait for each batch to complete. (i.e. Within the autoreleasepool.)

This waits after all batches.

From a timing standpoint, I found it more useful to have this at the batch level because the XCTest framework already provides a time for the completion of the entire test case.

Additionally, waiting at the end of each batch helps ensure that the autoreleasepool cleans up after all the finished operations in each batch (which helps keep memory usage bounded and predictable - and, thus, makes it easy to run as many batches as you want, bounded only by how long you feel like running the StressTests).

danthorpe: Yeah, I did have it like this in an previous commit, I'm just going to rethink how it will work. Currently we have one `XCTestCase` subclass, which then has many tests inside it.

I think what I'm going to change is to make a `StressTestCase` class (subclass of `ProcedureKitTestCase`) and then provide APIs on here to perform functions `beforeBatches`, `beforeIterations`, `afterIterations` and `afterBatches`, so that each stress test will be a single instance of `StressTestCase` with a single test in it, which calls the `stress` function.

Issue #453, opened 2016-09-07
macOS 10.10 deployment target


JamieWhite:
In your installation documentation you say you support 10.10+ but the deployment target is set for 10.11. Is there any reason for this?

JamieWhite:
Additionally I'd PR with this fix but I've no idea which branch this belongs to and how you're managing all your branches for Swift versions.

danthorpe:
Hi @JamieWhite - yeah, that's probably a mistake... where are you looking at this?

`development` -> Swift 3, _ProcedureKit_ v4 development
`swift/2.2` -> Swift 2.2, _Operations_ v3.* work (feature branches should be `swift/2_2/OPR-123_my_feature)
`swift/2.3` -> Swift 2.3, _Operations_ v3.* rebased off `swift/2.2`

So, if you have spotted this in v3.2.0, or the `swift/2.2` branch - a PR would be welcome 🙌

```
git remote add upstream git@github.com:ProcedureKit/ProcedureKit.git
git fetch upstream
git checkout -b swift/2_2/OPR-453_macos_deployment_target upstream/swift/2.2
```

swiftlyfalling:
@JamieWhite: Please see the current `swift/2.2` branch, which should have this fixed for the OSX target.

JamieWhite:
Marvellous. Thanks.

PR #454, opened 2016-09-08
FIX: macOS deployment target should be 10.10 (Swift 2.2) (#453)

swiftlyfalling:
Was incorrectly set to 10.11. Credit to @JamieWhite, issue #453.


swiftlyfalling:
Had another spurious test failure caused by:

> [04:22:29]: ▸ ✗ test__logger_outputs_number_of_received_errors, XCTAssertTrue failed - log message: ["Test Logging Operation: will execute.", "Test Logging Operation: will finish with error(s): [Operations.ConditionError.BlockConditionFailed].", "Test Logging Operation: did finish with error(s): [Operations.ConditionError.BlockConditionFailed]."]


But retrying succeeded.

(Those LoggingObserver tests are not thread-safe, and they occasionally trip such issues.)

PR #455, opened 2016-09-08
Split CalendarCapbability out of Standard for all due to iOS10 strictness

itsthejb:
iOS10 is stricter about this now, so if the code is linked (seems Swift doesn't do dead code stripping yet?) you get the following iTC warning:

```
This app attempts to access privacy-sensitive data without a usage description. The app's Info.plist must contain an NSCalendarsUsageDescription key with a string value explaining to the user how the app uses this data.
```

Can be fixed by inserting a nonsense `NSCalendarsUsageDescription`, but would tend to suggest the `CalendarCapability` should just be optional now.

Thanks!


danthorpe:
Hi @itsthejb - are you trying to submit using Xcode 8 GM, iOS 10 and Swift 2.3? If so, can you rename your branch `swift/2_3/remove-calendar-dependency`

Also, I'd say, it's probably best that you make the necessary changes on a fork at the moment to decouple yourself from any risk here.

danthorpe:
@itsthejb - I'll take care of this for you I think.

itsthejb:
You can take care of it?

danthorpe:
@itsthejb - yep, sorting it out as #456.

itsthejb:
Ok, then I'll close this

PR #464, opened 2016-09-14
Fix compilation with swift legacy version.

yageek:
It was impossible to compile the project after settings the build
settings `SWIFT_VERSION=2.3`.

There was one ambiguity on the `dispatch_sync` function calls. I simply
added an intermediate struct to be able to call it explicitely.

There was one conflict with the Objective-C selector `atomic` when we
wanted to make `CKModifyRecordsOperation` adopting the
`CKModifyRecordsOperationType` protocol.
I renamed it as `atomicRecord`.


danthorpe:
Hi @yageek - sorry, what is your intent with this PR?

danthorpe:
Okay - so, these changes are already on the `swift/2.3` branch. If you need to submit now with Xcode 8, you can use that branch.

It will be released (probably today) as v3.4.0

PR #533, opened 2016-10-19
Add space in the repo for presentations 

danthorpe:
There have been a couple of presentations given about _Operations_ and _ProcedureKit_. I think it would be great if they could be added to the repo. Partly as an element of documentation, but also for guidance and maybe for others to re-present them.


danthorpe:
@jshier - Just seen the SwiftSummit email 😬 

I'm sure you're already done - but just in case these help you figure out the direction I went in...

jshier:
@danthorpe Check your Slack, I'm starting right now. 💨 😄

Issue #176, opened 2016-02-04
More convenient operation dependency sequence creation


jshier:
I've found myself rebuilding my `Operations` based networking layer recently and have gotten rather sick of the `declare operation -> add conditions/observer -> declare second operation -> add conditions/observers -> add first operation dependency to second operation -> etc.` code I write rather frequently. Plus, I'm rather a fan of some of the asynchronous sequence code I've see used by the various promise libraries or the [Async library](https://github.com/duemunk/Async). So I created the following extensions as a first take on such a feature:

``` swift
extension NSOperation {    
    func then(operationClosure: () -> NSOperation) -> [NSOperation] {
        let operation = operationClosure()
        operation.addDependency(self)

        return [self, operation]
    }
}

extension CollectionType where Generator.Element == NSOperation, Index: BidirectionalIndexType {
    func then(operationClosure: () -> Generator.Element) -> [Generator.Element] {
        let operation = operationClosure()
        var allOperations = Array(self)

        if let lastOperation = allOperations.last {
            operation.addDependency(lastOperation)
        }

        allOperations.append(operation)

        return allOperations
    }
}
```

Which can be used like:

``` swift
static func sequenceOfOperations(completionHandler: () -> ()) {
    sharedManager.addOperations(InterruptibleNetworkOperation<HTTPBinResponse>(request: APIRouter.GET)
        .then { InterruptibleNetworkOperation<HTTPBinResponse>(request: APIRouter.GET) }
        .then { InterruptibleNetworkOperation<HTTPBinResponse>(request: APIRouter.GET) }
        .then { BlockOperation(mainQueueBlock: completionHandler) }, waitUntilFinished: false)
}
```

I can see this extended with something like `thenIf(predicate:)` which would attach a condition to the operation, or perhaps a call to add an observer.

So my question is, would this be valuable for the library? What immediate improvements or ideas do you see?

danthorpe:
This is really interesting, and quite serendipitous - see my [tweets from yesterday](https://twitter.com/danthorpe/status/694986693796085760). 

[...]

PR #536, opened 2016-10-20
Adds .then { } API

danthorpe:
See the discussion in #176.


jshier:
🎆 👍 

Perhaps an `enqueue(on: ProcedureQueue = ProcedureQueue())` for a `Sequence` of `Operation`? It would allow:
```swift
procedure.then { procedure2 }.then { procedure3 }.enqueue()
```
Though that could perhaps be for later.

PR #542, opened 2016-10-23
CloudKitProcedure

danthorpe:


StephenDaedalus:
Having been quite happily using the CloudKit support in the Swift 2.3 branch, I only recently thought about moving to Swift 3.0 when I saw announcements of 4.0 Beta 5. 

Unfortunately I didn't read the notes closely enough to know that CloudKit support was not fully there. 

I have no idea if integrating with OPR-542_cloud_kit_procedure is advisable at this point. 

Should I just continue with Swift 2.3 for the time being?

danthorpe:
I've added #582 for adding _batch modify_ support, which does exist in _Operations_, but is certainly a contentious piece of functionality, given the way that CloudKit processes items for modification.

Issue #543, opened 2016-10-24
ExclusivityManager synchronization bug


myexec:
I'm using the framework in my project and noticed a bug in the ExclusivityManager class (or may I be wrong). The queues returned from the DispatchQueue+ProcedureKit extension are all concurrent, in particular DispatchQueue.initiated returns concurrent queue. Therefore 'add' and 'remove' functions are not synchronized. I tested this adding 2 mutually exclusive operations at once. The result was they were not mutually exclusive. I fixed this like so:

```swift
func add(procedure: Procedure, category: String) -> Operation? {
        return lock.withCriticalScope{
            _add(procedure: procedure, category: category)
        }
    }

    func remove(procedure: Procedure, category: String) {
        lock.withCriticalScope{
            _remove(procedure: procedure, category: category)
        }
    }
```

Awesome framework btw 👍

danthorpe:
Hi @myexec  - yep that is definitely a bug! Great catch! Will fix asap.

PR #545, opened 2016-10-25
Fix ExclusivityManager synchronization bug (#543)

swiftlyfalling:
The ExclusivityManager was using `DispatchQueue.initiated`, which returns a global queue that is a **concurrent** queue, however its internals are dependent upon **serial** execution of add/remove.
Fixed to create and use a serial queue.

(Reported by @myexec in #543.)

### DIFF
--- a/Sources/MutualExclusion.swift
+++ b/Sources/MutualExclusion.swift
@@ -27,7 +27,7 @@ public class ExclusivityManager {
 
     static let sharedInstance = ExclusivityManager()
 
-    fileprivate let queue = DispatchQueue.initiated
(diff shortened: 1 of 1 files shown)


swiftlyfalling:
StressTests failed because of the issue that is fixed in PR #544.

danthorpe:
Thanks @swiftlyfalling - simple fix is often the best. Can discuss later how to minimise unintended misuse of the `DispatchQueue` extension. Possibly get rid of it, or maybe wrap the statics inside a `Global` struct?

Something like:

```swift
extension DispatchQueue {
    public struct Global {
        public static var `default`: DispatchQueue {
            return DispatchQueue.global(qos: DispatchQoS.QoSClass.default)
        }
        private init() { }
   }
}
```

PR #522, opened 2016-10-27
CloudKit programming model question

danthorpe:
My Swift 3 app is heavily using CloudKit. Based on beta 3 I tried to migrate the CloudKit fetches at app launch to ProcedureKit. I wrapped each of those fetches in a BlockOperation and run them as part of a group procedure. However due to the asynchronous nature of CloudKit those fetches are kicked of to CloudKit and return immediately triggering the group procedure finished observer. Problem is that the fetched results delivered by CloudKit arrive after that. I was thinking that I would need to add something like a lock to the group procedure that the results block would then remove. Only after all block operations are executed and all locks are removed the finish observer would run. 

I just don't understand yet how I would build such a lock with ProcedureKit? Or should i rather wait for CloudKitProcedure in beta 4? And if so, how would I then approach this problem?

### DIFF
--- a/Sources/Testing/ProcedureKitTestCase.swift
+++ b/Sources/Testing/ProcedureKitTestCase.swift
@@ -45,6 +45,10 @@ open class ProcedureKitTestCase: XCTestCase {
     }
 
     public func wait(for procedures: Procedure..., withTimeout timeout: TimeInterval = 3, withExpectationDescription expectationDescription: String = #function, handler: XCWaitCompletionHandler? = nil) {
+        wait(forAll: procedures, withTimeout: timeout, withExpectationDescription: expectationDescription, handler: handler)
+    }
+
+    public func wait(forAll procedures: [Procedure], withTimeout timeout: TimeInterval = 3, withExpectationDescription expectationDescription: String = #function, handler: XCWaitCompletionHandler? = nil) {
         for (i, procedure) in procedures.enumerated() {
             addCompletionBlockTo(procedure: procedure, withExpectationDescription: "\(i), \(expectationDescription)")
         }


danthorpe:
Whoops - wrong issue.

marwey:
@danthorpe Just wondering when Beta 5 will be released? Really depend on CloudKitProcedure in order to get started with ProcedureKit. 

Thanks for the great work so far! ;-)

danthorpe:
@marwey - yeah, sorry about that - I was hoping `CloudKitProcedure` would make it in time. But there are still some test cases to write, and the error handling model to do properly. So, #542 which is almost complete puts in place the majority of the functionality for CloudKit, but there will be another ticket to complete the error handling stuff.

In terms of timelines - I'm going to work on this mostly full-time today & tomorrow, so I hope it will all get done in `development` before Thursday - so you could begin using development. But then I'm going on holiday for a week, and I don't plan on doing Beta 5 I've finished off some other bits and pieces.

I would say that working of `development` is enough to get started integrating into an app. So, keep and eye out for what gets merged later in the week.

marwey:
@danthorpe Thanks for the heads-up! As long as the programming model has a certain level of stability I am very much willing to get started with development. Even if I have to make some changes later on I would be totally fine with this approach. 

I am actually more worried about understanding the programming model and heading into the right direction. So I will probably using the next week to play around with development, learn about  the programming model and put together a list with really dump questions for you to answer after your vacation. ;-)

danthorpe:
@marwey - sounds great - that's certainly something which will be worked on throughout November. I'd say that _ProcedureKit_ is almost dev complete, but it's not going to be "4.0.0" _final_ until the header docs are complete * published, and the wiki/programming guide is finished.

PR #554, opened 2016-10-27
Add didExecuteObserver

danthorpe:
As reported by @marek-rogosz-sky.

### DIFF
--- a/Tests/BlockProcedureTests.swift
+++ b/Tests/BlockProcedureTests.swift
@@ -30,4 +30,14 @@ class BlockProcedureTests: ProcedureKitTestCase {
         wait(for: block)
         XCTAssertProcedureFinishedWithErrors(block, count: 1)
     }
+
+    func test__block_did_execute_observer() {
+        let block = BlockProcedure { /* does nothing */ }
+        var didExecuteBlockObserver = false
+        block.addDidExecuteBlockObserver { procedure in
+            didExecuteBlockObserver = true
+        }
+        wait(for: block)
+        XCTAssertTrue(didExecuteBlockObserver)
+    }
 }
--- a/Sources/AnyObserver.swift
+++ b/Sources/AnyObserver.swift
@@ -7,6 +7,7 @@
 class AnyObserverBox_<Procedure: ProcedureProtocol>: ProcedureObserver {
     func didAttach(to procedure: Procedure) { _abstractMethod() }
     func will(execute procedure: Procedure) { _abstractMethod() }
+    func did(execute procedure: Procedure) { _abstractMethod() }
     func will(cancel procedure: Procedure, withErrors: [Error]) { _abstractMethod() }
     func did(cancel procedure: Procedure, withErrors errors: [Error]) { _abstractMethod() }
     func procedure(_ procedure: Procedure, didProduce newOperation: Operation) { _abstractMethod() }
@@ -27,6 +28,10 @@ class AnyObserverBox<Base: ProcedureObserver>: AnyObserverBox_<Base.Procedure> {
 
     override func will(execute procedure: Base.Procedure) {
         base.will(execute: procedure)
+    }
+
+    override func did(execute procedure: Base.Procedure) {
+        base.did(execute: procedure)
     }
 
     override func will(cancel procedure: Base.Procedure, withErrors errors: [Error]) {
@@ -65,6 +70,10 @@ public struct AnyObserver<Procedure: ProcedureProtocol>: ProcedureObserver {
 
     public func will(execute procedure: Procedure) {
         box.will(execute: procedure)
+    }
+
+    public func did(execute procedure: Procedure) {
+        box.did(execute: procedure)
     }
 
     public func will(cancel procedure: Procedure, withErrors errors: [Error]) {
--- a/Sources/BlockObservers.swift
+++ b/Sources/BlockObservers.swift
@@ -142,7 +142,7 @@ public struct DidExecuteObserver<Procedure: ProcedureProtocol>: ProcedureObserve
     /**
      Initialize the observer with a block.
 
-     - parameter willExecute: the `Block`
+     - parameter didExecute: the `Block`
      - returns: an observer.
      */
     public init(didExecute: @escaping Observer<Procedure>.DidExecute) {
--- a/Sources/BlockObservers.swift
+++ b/Sources/BlockObservers.swift
@@ -133,6 +133,16 @@ public struct WillExecuteObserver<Procedure: ProcedureProtocol>: ProcedureObserv
 
 /// DidExecuteObserver is an observer which will execute a
 /// closure when the operation starts.
+///
+/// - notes: this observer will be invoked directly after the
+/// `execute` returns.
+///
+/// - warning: there are no guarantees about when this observer
+/// will be called, relative to the lifecycle of the procedure. It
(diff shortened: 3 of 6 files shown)


danthorpe:
@swiftlyfalling - yes. Thoughts?

swiftlyfalling:
@danthorpe: Makes sense to me. We just need to document and highlight that `didExecuteObserver`s fire right after the Procedure's `execute()` function returns.

danthorpe:
This was a conversation between me and @swiftlyfalling about `didExecuteObserver`:

> danthorpe [6:43 PM]  
> Did you have any comments/concerns about the `didExecuteObserver`? I'm going to merge my changes for it.
> 
> swiftlyfalling [7:26 PM]  
> I don't have a preference as to when it should fire, but I think it's extremely important to document it - because it isn't necessarily clear from its name.
> 
> [7:27]  
> Because there's a function you have to override named "execute()" and a Procedure state called executing.
> 
> [7:28]  
> And asynchronous procedures may return from "execute()" and yet still be executing.
> 
> [7:29]  
> My impression was that that potential for confusion was why there wasn't previously a "didExecute" observer.
> 
> danthorpe [7:30 PM]  
> Yeah, I think it's just been an oversight.
> 
> [7:30]  
> And generally if you've got async execute behaviour, then you just do something in a block, or on did finish.
> 
> [7:31]  
> But in some cases (test cases) you might want to trigger an action after the procedure executed (whether async or not) but before it finishes
> 
> swiftlyfalling [7:32 PM]  
> willExecuteObservers are fired before the Procedure's state changes to executing *and* before "execute()" is called, whereas didExecuteObservers will fire after "execute()" is called but before the Procedure's state changes (and potentially long before the Procedure subclass finishes).
> 
> danthorpe [7:32 PM]  
> Correct.
> 
> [7:32]  
> I will add documentation to state this explicitly
> 
> swiftlyfalling [7:32 PM]  
> I'm okay with it - I just think it should be highlighted in the documentation.
> 
> [7:32]  
> Yeah
> 
> danthorpe [7:33 PM]  
> :+1:
> 
> swiftlyfalling [7:34 PM]  
> And I can totally see the use-case for wanting to handle right after execute() returns. And the other case can be handled with a will/didFinishObserver. (edited)
> 
> [7:35]  
> Perhaps this is another reason to map out a full state flow chart. Could be helpful for users.
> 
> danthorpe [7:37 PM]  
> Definitely.
> 
> [7:41]  
> Also as a heads-up: it is technically possible for the didExecuteObserver to fire *after* the Procedure finishes.
> 
> [7:42]  
> That would be due to a side-effect of how/when the Procedure subclass calls finish(). But it's technically possible
> 
> [7:42]  
> So all that's currently guaranteed is that didExecuteObservers will be called after execute() returns. The rest is up to the specifics of the Procedure subclass implementation.
> 
> danthorpe [7:43 PM]  
> That was what I was gonna ask you. Should we guarantee that the observers are called in order?
> 
> swiftlyfalling [7:44 PM]  
> That would technically mean blocking finish until the didExecuteObservers fire, in this case.
> 
> [7:44]  
> Which could mean blocking another thread that is doing who knows what. (edited)
> 
> danthorpe [7:45 PM]  
> Yeah. :unamused:
> 
> [7:45]  
> I'm kinda happy to not worry about this.
> 
> [7:45]  
> Call it out as something to check in the docs
> 
> swiftlyfalling [7:46 PM]  
> Yeah. If someone provides an obvious, common case, we can re-evaluate guaranteeing it.
> 
> [7:46]  
> But side-effects of such a change would have to be carefully thought through.
> 
> danthorpe [7:46 PM]  
> I might just add some more test cases around it
> 
> [7:47]  
> E.g. An empty BlockProcedure ?
> 
> swiftlyfalling [7:47 PM]  
> If the thread that is calling finish() is holding some sort of lock that is also acquired by a didExecuteObserver... deadlock.
> 
> [7:47]  
> Unless finish() becomes asynchronous and returns immediately.
> 
> danthorpe [7:48 PM]  
> That is a possibility
> 
> swiftlyfalling [7:48 PM]  
> But that's also a significant behavior change
> 
> danthorpe [7:48 PM]  
> Yeah. We discussed it before I think
> 
> swiftlyfalling [7:48 PM]  
> I think so
> 
> [7:49]  
> In any case, didExecute ordering is only an issue in certain circumstances
> 
> danthorpe [7:49 PM]  
> We'd need our own queue though I think
> 
> [7:49]  
> I basically don't want to do that.
> 
> swiftlyfalling [7:49 PM]  
> That's one way to do it, sure
> 
> [7:49]  
> And yeah, I agree (edited)
> 
> danthorpe [7:50 PM]  
> I think copious warnings.
> 
> swiftlyfalling [7:50 PM]  
> Yep
> 
> danthorpe [7:51 PM]  
> One option, would be to add a timeout observer to each procedure in debug
> 
> [7:51]  
> To log/warn if an operation is running for too long
> 
> [7:52]  
> Indicating a possible deadlock
> 
> [7:52]  
> Sort of like a deadlock assertion
> 
> swiftlyfalling [7:52 PM]  
> Except I'm not sure how it's possible to pick a good default value for the timeout
> 
> [7:52]  
> People use operations for all sorts of tasks
> 
> danthorpe [7:53 PM]  
> Yeah, would need be overrideable.
> 
> swiftlyfalling [7:53 PM]  
> Additionally, look at CloudKitProcedure. Asynchronous. No guarantee when CloudKit returns a response.
> 
> [7:53]  
> There are a lot of cases like that.
> 
> danthorpe [7:53 PM]  
> Yeah
> 
> [7:53]  
> That is why I've not done it already
> 
> swiftlyfalling [7:53 PM]  
> I think there's a limit to what the framework can or should do to prevent people from making mistakes.
> 
> danthorpe [7:54 PM]  
> Totally agree with that
> 
> swiftlyfalling [7:54 PM]  
> If their procedure takes longer than they expect, they will presumably know.
> 
> danthorpe [7:54 PM]  
> So, document possible dangers from misuse
> 
> swiftlyfalling [7:54 PM]  
> Yep
> 
> danthorpe [7:54 PM]  
> :+1:
> 
> swiftlyfalling [7:55 PM]  
> Trying to solve the halting problem is, I think, reasonably out of scope. :smile:
> 
> danthorpe [7:55 PM]  
> Yes. Do you mind if I copy/paste this conversation into GitHub?
> 
> swiftlyfalling [7:56 PM]  
> Sure. That's fine.
> 
> [7:56]  
> Good to have the thoughts documented
> 
> [7:58]  
> As far as possible future changes go, I think there's a possible proposal around making finish() (and cancel()?) asynchronous - as handling something *after* the procedure finishes can (and probably should) be done with a didFinishObserver.
> 
> [7:59]  
> But it does require a bunch more thought

Tests/BlockProcedureTests.swift
```diff
@@ -30,4 +30,14 @@ class BlockProcedureTests: ProcedureKitTestCase {
         wait(for: block)
         XCTAssertProcedureFinishedWithErrors(block, count: 1)
     }
+
+    func test__block_did_execute_observer() {
+        let block = BlockProcedure { /* does nothing */ }
+        var didExecuteBlockObserver = false
+        block.addDidExecuteBlockObserver { procedure in
+            didExecuteBlockObserver = true
```
danthorpe: Not that at this point it is possible that `procedure.isFinished` will be true.

Issue #548, opened 2016-10-26
Compiling failed with Xcode 8.1 GM seed


pomozoff:
I've downloaded Xcode Version 8.1 (8T61a) GM seed and try to build a ProcedureKit 4.0.0 Beta 3 with Carthage and it fails with error:

> duplicate symbol __TMRbBp in:
>     /Users/apomozov/Library/Developer/Xcode/DerivedData/ProcedureKit-erdkskxxwyhdaoanlnvglvwxlmle/Build/Intermediates/ProcedureKit.build/Release-iphoneos/ProcedureKit.build/Objects-normal/arm64/Condition.o
>     /Users/apomozov/Library/Developer/Xcode/DerivedData/ProcedureKit-erdkskxxwyhdaoanlnvglvwxlmle/Build/Intermediates/ProcedureKit.build/Release-iphoneos/ProcedureKit.build/Objects-normal/arm64/ProcedureQueue.o
> ld: 1 duplicate symbol for architecture arm64
> clang: error: linker command failed with exit code 1 (use -v to see invocation)

danthorpe:
Thanks @pomozoff - will try to look into this asap, as I need to tag Beta 4.

wfilleman:
pomozoff,

I'm using ProcedureKit 4.0.0 Beta 3, but I'm seeing the duplicate symbol __TMRbBp error in another part of my code base. Specifically a TodayView extension on three custom swift files I've written. None of which use or import ProcedureKit.

XCode 8.0 builds without issues, but the 8.1 GM and final 8.1 both have this issue. I'm stumped and opened up a discussion on the Apple Developer forum for hopefully some insight: https://forums.developer.apple.com/message/191894#191894

If you've been able to work through this, please let me know.

Wes

danthorpe:
@wfilleman thanks for the info.

danthorpe:
Has anyone tried building `development` from source with Xcode 8.1 final yet?

jshier:
I build with the GM through CocoaPods every day.

danthorpe:
@jshier on `development`?

jshier:
Yes, 
```ruby
pod 'ProcedureKit', git: "https://github.com/ProcedureKit/ProcedureKit", branch: "development"
  pod 'ProcedureKit/Location', git: "https://github.com/ProcedureKit/ProcedureKit", branch: "development"
```

danthorpe:
[...]

PR #558, opened 2016-10-28
Workaround for a Swift3 bug, issue #548

pomozoff:


danthorpe:
This is fixed in #559. Thanks so much @pomozoff for figuring this out! Will be released this weekend with beta 4.

PR #561, opened 2016-10-31
NetworkProcedure

danthorpe:
This is a refactor of the Network procedure wrappers, to unify both _reachable_ and _resilient_ wrappers into a single `RetryProcedure` subclass.


danthorpe:
### `NetworkProcedure`

`NetworkProcedure` is a "wrapper" procedure, which means that it provides a context in which to run other `Procedure` instances, in this case for reliable networking.

`NetworkProcedure` provides the following basic features:

1. Automatic handling of client network reachability error handling
2. Semi-automatic handling for client & server network resiliency features, including request timeouts.

### Usage

For example - performing a network data request, using default parameters: 

```swift
let procedure = NetworkProcedure { NetworkDataProcedure(session: URLSession.shared, request: aRequest) }
```

in the happy case, the `NetworkProcedure` will just execute the `NetworkDataProcedure` and all is good. 

However, if that instance finishes with an error, `NetworkProcedure` will inspect it and perform some automatic retries. In cases where the error indicates that the client should wait for its own network reachability status to change before retrying, this will be done automatically. There is no extra configuration needed.

If the procedure finishing with a "transient error", Apple's recommendation is to retry as normal, this is also handled automatically.

In situations where the client request succeeds, and the procedure finishes with any errors, `NetworkProcedure` will still look at the HTTP status code, and grant the ability to perform a retry in the case of client or server error codes. By default, `NetworkProcedure` will retry client error codes 408 (request timeout), and 429 (too many requests) and all server error codes.

### Customization

To customise behavior, initialise `NetworkProcedure` with a type conforming to `NetworkResilience` protocol, like this:

```swift

public struct CustomNetworkResilience: NetworkResilience {

    public let maximumNumberOfAttempts: Int
    public let backoffStrategy: WaitStrategy
    public let requestTimeout: TimeInterval?

    public init(maximumNumberOfAttempts: Int = 6, backoffStrategy: WaitStrategy = .constant(5), requestTimeout: TimeInterval? = 12.0) {
        self.maximumNumberOfAttempts = maximumNumberOfAttempts
        self.backoffStrategy = backoffStrategy
        self.requestTimeout = requestTimeout
    }

    public func shouldRetry(forResponseWithHTTPStatusCode statusCode: HTTPStatusCode) -> Bool {
        return code.isServerError
    }
}

let procedure = NetworkProcedure(resilience: CustomNetworkResilience(), connectivity: .wifi) { NetworkDataProcedure(session: URLSession.shared, request: aRequest) }
```

See `DefaultNetworkResilience` for default resiliency behavior.

The type of network request is entirely customisable, so long as it conforms to the `NetworkOperation` protocol.

PR #565, opened 2016-11-02
Add NetworkDownloadProcedure.

yageek:
The implementation for class and test is largely inspired by
`NetworkDataProcedure` class.


danthorpe:
@yageek - awesome. That's pretty much _exactly_ what I would have written! 😁

PR #566, opened 2016-11-02
Fix Condition to support ResultInjection

danthorpe:
At the moment, `Condition` has a result property, which is not a `PendingValue<T>`. Subclasses cannot therefore conform to `ResultInjection`.

### DIFF
--- a/Sources/Condition.swift
+++ b/Sources/Condition.swift
@@ -11,9 +11,6 @@
  */
 public enum ConditionResult {
 
-    /// Indicates that the condition is pending
-    case pending
-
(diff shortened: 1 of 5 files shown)


danthorpe:
With these changes - custom conditions which implement `ResultInjection` is now possible:

```swift
class CustomCondition: Condition, ResultInjection {

    var requirement: PendingValue<String> = .pending

    override func evaluate(procedure: Procedure, completion: @escaping (ConditionResult) -> Void) {
        if let requirement = requirement.value {
            // Use the requirement here

        }
    }
}
``` 

/cc @bizz84

PR #567, opened 2016-11-02
Add NetworkUploadProcedure.

yageek:
The implementation class and test has been largely inspired by
`NetworkDataProcedure`.

The requirement type is currently a pending value of a tuple.
Maybe would it make more sense to create a dedicated struct to add more
clarity in the case of dependencies injection ?

### DIFF
--- a/ProcedureKit.xcodeproj/project.pbxproj
+++ b/ProcedureKit.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0B15EDE91DCA0AFE0060D2D7 /* NetworkUpload.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0B15EDE71DCA07CB0060D2D7 /* NetworkUpload.swift */; };
+		0B15EDEC1DCA147C0060D2D7 /* NetworkUploadProcedureTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0B15EDEA1DCA13810060D2D7 /* NetworkUploadProcedureTests.swift */; };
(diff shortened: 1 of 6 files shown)


danthorpe:
@yageek how about this for the requirement:

```swift
public struct HTTPRequirement<Payload: Equatable>: Equatable {
    public static func == (lhs: HTTPRequirement <Payload>, rhs: HTTPRequirement <Payload>) -> Bool {
        return lhs.payload == rhs.payload && lhs.request == rhs.request
    }
    public var payload: Payload
    public var request: URLRequest
}
```

danthorpe:
So, that it would be used like this:

```swift
open class NetworkUploadProcedure<Session: URLSessionTaskFactory>: Procedure, ResultInjection {
    public var requirement: PendingValue<HTTPRequirement<Data>> = .pending
    public var result: PendingValue<HTTPResult<Data>> = .pending

```

yageek:
Completely. Do we want also to pass the `HTTPRequirement` requirement struct at the initializer or we keep an initializer with both `URLRequest`  and `Data` ? May be both ?

danthorpe:
Hmm, good question - perhaps designated initialiser with the struct, and then a `convenience`? As a consumer, what would you want?

yageek:
The problem by having both: 

```swift
convenience init(session: Session, request: URLRequest? = nil, data: Data? = nil, completionHandler: @escaping (HTTPResult<Data>) -> Void = { _ in })

public init(session: Session, requirements: HTTPRequirement<Data>? = nil, completionHandler: @escaping (HTTPResult<Data>) -> Void = { _ in }) 
```
will lead to an ambiguous case if only the `session` parameter is provided.

I think that the version with `URLRequest` and `Data` will be more usable.

danthorpe:
Okay, so you'll have this: 

```swift
init(session: Session, request: URLRequest? = nil, data: Data? = nil, completionHandler: @escaping (HTTPResult<Data>) -> Void = { _ in }) {
    requirement = request.flatMap { HTTPRequirement(data: nil, request: $0) } ?? .pending
    // etc
}
```

Issue #568, opened 2016-11-02
TaskProcedure


danthorpe:
Needs adding to _ProcedureKitMac_, essentially equivalent to: [TaskOperation](https://github.com/ProcedureKit/ProcedureKit/blob/3.4.0/Sources/Features/Shared/TaskOperation.swift).

PR #570, opened 2016-11-03
Add ProcessProcedure.

yageek:
The class has been largely inspired by the original `TaskOperation`.
Refers to #568.

### DIFF
--- a/Sources/Mac/Process.swift
+++ b/Sources/Mac/Process.swift
@@ -1,0 +1,82 @@
+//
+//  ProcedureKit
+//
+//  Copyright © 2016 ProcedureKit. All rights reserved.
+//
+
+import Foundation
+import ProcedureKit
+
+open class ProcessProcedure: Procedure {
+
+    /// Error type for ProcessProcedure
+    public enum Error: Swift.Error, Equatable {
+        case LaunchPathNotSet
+        case TerminationReason(Process.TerminationReason)
+    }
+
+    /// Closure type for testing if the task did exit cleanly
+    public typealias ProcessDidExitCleanly = (Int) -> Bool
+
+    /// The default closure for checking the exit status
+    public static let defaultProcessDidExitCleanly: ProcessDidExitCleanly = { status in
+        switch status {
+        case 0: return true
+        default: return false
+        }
+    }
+
+    // - returns process: the Process
+    public let process: Process
+
+    /// - returns processDidExitCleanly: the closure for exiting cleanly.
+    public let processDidExitCleanly: ProcessDidExitCleanly
+
+    /**
+     Initializes ProcessProcedure with a Process.
+     - parameter task: the Process
+     - parameter processDidExitCleanly: a ProcessDidExitCleanly closure with a default.
+     */
+    public init(process: Process, processDidExitCleanly: @escaping ProcessDidExitCleanly = ProcessProcedure.defaultProcessDidExitCleanly) {
+        self.process = process
+        self.processDidExitCleanly = processDidExitCleanly
+        super.init()
+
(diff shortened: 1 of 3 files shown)


Sources/Mac/Process.swift
```diff
@@ -0,0 +1,82 @@
...
+     Initializes ProcessProcedure with a Process.
+     - parameter task: the Process
+     - parameter processDidExitCleanly: a ProcessDidExitCleanly closure with a default.
+     */
+    public init(process: Process, processDidExitCleanly: @escaping ProcessDidExitCleanly = ProcessProcedure.defaultProcessDidExitCleanly) {
+        self.process = process
+        self.processDidExitCleanly = processDidExitCleanly
+        super.init()
+
+        addWillCancelBlockObserver { [unowned self] procedure, errors in
```
danthorpe: So, there is no need here to capture `self` in these observer blocks anymore. In this case `procedure` is already guaranteed to be the receiver, with full type safety. So you can just do this:

```swift
addWillCancelBlockObserver { procedure, errors in
    guard procedure.process.isRunning else { return }
    procedure.process.terminate()
}
```

Sources/Mac/Process.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    open override func execute() {
+        guard let _ = process.launchPath else {
+            finish(withError: Error.LaunchPathNotSet)
+            return
+        }
+
+        let previousTerminationHandler = process.terminationHandler
+
+        process.terminationHandler = { [unowned self] task in
```
danthorpe: I think here, we'd prefer this to be changed to `[weak self]`, throughout the framework we're changing all of these sort of captures to be weak, as it's not necessarily guaranteed that `self` will not have been deallocated when the closure is invoked. So, something like:

```swift
process.terminationHandler = { [weak self] task in
    guard let strongSelf = self else { return }
    // etc etc
}
```

Sources/Mac/Process.swift
```diff
@@ -0,0 +1,82 @@
...
+
+import Foundation
+import ProcedureKit
+
+open class ProcessProcedure: Procedure {
+
+    /// Error type for ProcessProcedure
+    public enum Error: Swift.Error, Equatable {
+        case LaunchPathNotSet
+        case TerminationReason(Process.TerminationReason)
```
danthorpe: Can we change these cases to be Swift 3.0 style, which is camel case. 😀

Sources/Mac/Process.swift
```diff
@@ -0,0 +1,82 @@
...
+            else {
+                self.finish(withError: Error.TerminationReason(self.process.terminationReason))
+            }
+        }
+
+        process.launch()
+    }
+}
+
+public func == (lhs: ProcessProcedure.Error, rhs: ProcessProcedure.Error) -> Bool {
```
danthorpe: Not really sure what is going on here, but in Swift 3.0, conforming to equatable means putting this as a static function inside the type?

yageek: I was not aware about that change. I will fix that.

yageek:
Regarding the implementation of `Equatable` protocol, it is still possible to implement it in the global scope. I tried with the `Identifiable` protocol and implement it like:

```swift
public protocol Identifiable {
    var identifier: UUID { get }
}

public func ==<T: Identifiable>(lhs: T, rhs: T) -> Bool {
    return lhs.identifier == rhs.identifier
}

```

danthorpe:
@yageek - yeah, I guess that's a little bit confusing. 🤔 So far, I've been adding the static functions. And I think I prefer it, as it helps keep implementations collected together.

danthorpe:
Anyway, this is great! Thanks @yageek!

PR #571, opened 2016-11-03
Procedure: cancel() should not call finish()

swiftlyfalling:
Previously, the default behavior of the base Operation/Procedure class was to automatically call `finish()` when `cancel()` is called.

This behavior is problematic once the Procedure's `execute()` function has been called. The framework cannot safely handle this itself because it cannot know what the Procedure is doing, has done, or will do. Additionally, if cancelled while executing, the Procedure may finish with the cancellation errors instead of the errors that the subclass passes to its later call to finish().

**Once `execute()` is called, it must be the responsibility of the Procedure to call finish() in all cases**, and this includes the case of handling cancellation (as appropriate for that Procedure).

Thus, this PR changes the base Procedure class behavior of cancellation and finishing from:

- When a Procedure is cancelled (at any time), `finish()` will be automatically called (unless `disableAutomaticFinishing == true`). If cancelled before it executes, `execute()` will not be called.

to:

- When a Procedure is cancelled before it executes, the Procedure will automatically transition to finished upon executing (unless `disableAutomaticFinishing == true`), and `execute()` will not be called.
- Once a Procedure's `execute()` function is called, it is responsible for calling `finish()`.
- Once a Procedure's `execute()` function is called, it is responsible for handling cancellation (as appropriate/required).


danthorpe:
Hey @swiftlyfalling - will have a look though this. Overall, I'm pretty happy with the direction - but might need a bit of time to think of the consequences/impact.

swiftlyfalling:
I tested the existing Procedure subclasses in the framework, and only found a few that required modifications.

In general, the simplified "rules of thumb" will be:

- A Procedure subclass must always call `finish()`, once `execute()` has been called.
- When handling cancellation, if you do something that would prevent `finish()` from being called, you must call `finish()`.
- Don't call `finish()` before the Procedure has been executed.

For example, `DelayProcedure` uses a DidCancel observer to respond to cancellation, in which it cancels the dispatch timer it uses. Since this also prevents the completion block from being called (which is where `DelayProcedure` calls `finish()`), it must call `finish()` from within the DidCancel observer.

So DidCancel observer becomes:

    addDidCancelBlockObserver { procedure, _ in
        procedure.timer?.cancel()
        if procedure.isExecuting {
            procedure.finish()
        }
    }

jshier:
This has always been a confusing aspect of building `Procedure` subclasses for me, especially when the procedure is doing something asynchronous like a network request. Also, there's nothing in the updated header docs about overriding `procedureDidCancel`. Is that still a suggested method of handling cancellation? And in the case of using it, instead of just cancelling my asynchronous task, I would issue that cancellation, followed by a check against `isExecuting` to issue a `finish()` myself?

swiftlyfalling:
@jshier: I am writing up an overview that goes into detail on how to handle cancellation in different situations. There are indeed several methods, and different methods may be better for different types of Procedures. Draft coming soon.

swiftlyfalling:
A draft of new documentation on how to handle cancellation. (Comments/feedback/suggestions appreciated.)

# Handling Cancellation

By default, all Procedures provide a `cancel()` method, which sets `isCancelled` to `true`.

Built-in Procedures in ProcedureKit handle responding to cancellation.  Some examples are:

>  - **GroupProcedure** = Cancels its children. Finishes once all of its children have finished.

> - **DelayProcedure** = Immediately finishes when cancelled.

If you want your custom Procedure subclasses to respond to cancellation (i.e. finish early when cancelled), you must do some additional work.

## When ProcedureKit handles cancellation for you:

- When using a built-in Procedures in ProcedureKit (such as GroupProcedure or CloudKitProcedure).

- By default, when a custom Procedure subclass is cancelled **before it has begun executing** (i.e. before your `execute()` function is called).

## When you need to handle cancellation:

- When you need to take any additional steps to properly handle cancellation.

- In your custom Procedure subclass: When cancellation occurs after your Procedure's `execute()` function is called.

## Overview of methods to handle cancellation:

When a Procedure is cancelled, the following order of events occurs:

1. **`procedureWillCancel(withErrors:)`** is called

2. **WillCancel observers** are called

3. **`isCancelled`** is set to **`true`**

4. **`procedureDidCancel(withErrors:)`** is called

5. **DidCancel observers** are called

### You can handle cancellation by any combination of:

1. Checking `isCancelled`.

2. Will/Did Cancel observers.

3. Overriding `procedureWillCancel(withErrors:)` / `procedureDidCancel(withErrors:)`

Generally, the method you choose will be dependent upon the type of Procedure you are writing.

## Handling cancellation in a synchronous Procedure:

A **synchronous Procedure** performs all of its work in its`execute()` function, and calls `finish()` before returning.

For example, take a look at the following example Procedure that generates numbers in the Fibonacci sequence, and provides them one-by-one to a block:

```swift
// NOTE: Do not use this Procedure. It is not responsive to cancellation.
class FibonacciA: Procedure {
    let count: Int
    let block: ((Int) -> Void)
    init(count: Int = 10, block: @escaping (Int) -> Void) {
        self.count = count
        self.block = block
        super.init()
    }
    override func execute() {
        guard count > 0 else { finish(); return }
        block(0)
        var result = 0
        var prevResult = 1
        for _ in 1..<count {
            let nextResult = prevResult + result
            prevResult = result
            result = nextResult
            block(nextResult)
        }
        finish()
    }
}
```

The `FibonacciA` class can be cancelled, but does not ***respond*** to cancellation. 

If it is cancelled while it is executing, it will continue providing numbers from the Fibonacci sequence until it is complete.

To respond to cancellation, we must modify the loop in the `execute()` function to periodically check `isCancelled` and finish early if needed:

```swift
class FibonacciB: Procedure {
    let count: Int
    let block: ((Int) -> Void)
    init(count: Int = 10, block: @escaping (Int) -> Void) {
        self.count = count
        self.block = block
        super.init()
    }
    override func execute() {
        guard count > 0 else { finish(); return }
        block(0)
        var result = 0
        var prevResult = 1
        for i in 1..<count {
            if i % 2 == 0 { // no need to check `isCancelled` every time in this loop
                guard !isCancelled else {
                    // handle cancellation by finishing and returning immediately
                    finish()
                    return
                }
            }
            let nextResult = prevResult + result
            prevResult = result
            result = nextResult
            block(nextResult)
        }
        finish()
    }
}
```

The `FibonacciB` Procedure checks whether the Procedure has been canceled every 2 numbers in the sequence and, if so, finishes and returns immediately.

> 💡**NOTE:**

> You may want to throttle how frequently you check `isCancelled`, as each check involves a lock and will thus slightly impact performance.

> This is an important consideration for tight loops - if your loop can blow through 1000 items in 50ms, it probably doesn't make any sense to check `isCancelled` more frequently than every 1000 (or even 10,000) items. Figure out how quickly you want your Procedure to respond to cancellation, and try to calibrate how frequently you check for cancellation to that.

`FibonacciB` is **responsive** to cancellation. You can cancel it at any time, and even if it's currently executing it should quickly respond and finish.

## Handling cancellation in an asynchronous Procedure:

An **asynchronous Procedure** may return from its `execute()` function without finishing, and call `finish()` sometime later - often in a completion block in response to another API call.

For example, take a look at this simplified version of [DelayProcedure](https://github.com/ProcedureKit/ProcedureKit/blob/development/Sources/Delay.swift) (part of ProcedureKit):

```swift
// NOTE: Do not use this Procedure. It is not responsive to cancellation.
// ProcedureKit offers a better built-in alternative called "DelayProcedure".
// This is a simplified (and incomplete) example.
class OneMinuteDelayProcedureA: Procedure {
    var timer: DispatchSourceTimer? = nil
        timer = DispatchSource.makeTimerSource(flags: [], queue: DispatchQueue.default)
        timer?.setEventHandler { [weak self] in
            guard let strongSelf = self else { return }
            strongSelf.finish()
        }
        timer?.scheduleOneshot(deadline: .now() + .seconds(60))
        timer?.resume()
    }
}
```

The `OneMinuteDelayProcedureA` procedure waits for 60 seconds, and then finishes. `finish()` is called inside the eventHandler block for the DispatchSourceTimer.

> NOTE: The eventHandler block shows a best-practice of capturing `self` as a `weak` reference.
> It's not necessarily guaranteed that the Procedure will not have been deallocated when the closure is invoked, so it is generally recommend to capture self as `weak` and **not** `unowned`.

The `OneMinuteDelayProcedureA` procedure can be cancelled, but does not ***respond*** to cancellation.

If it is cancelled after it has begun executing, it will only finish when the timer fires.

To respond to cancellation, we need to do two things:

- Handle the DidCancel event.
- Cancel the timer and finish.

To handle the DidCancel event in our asynchronous Procedure, we can either override the `procedureDidCancel(withErrors:)` function, or add a DidCancel observer. Below, you'll see an example of the latter:

```swift
class OneMinuteDelayProcedureB: Procedure {
    var timer: DispatchSourceTimer? = nil
    override init() {
        super.init()
        addDidCancelBlockObserver { procedure, _ in
            procedure.timer?.cancel()
            // since cancelling the timer prevents it from calling its event handler
            // this will prevent the Procedure from finishing
            // thus, we must call finish() after we've cancelled the timer
            if procedure.isExecuting { // see NOTE below
                procedure.finish()
            }
        }
    }
        timer = DispatchSource.makeTimerSource(flags: [], queue: DispatchQueue.default)
        timer?.setEventHandler { [weak self] in
            guard let strongSelf = self else { return }
            strongSelf.finish()
        }
        timer?.scheduleOneshot(deadline: .now() + .seconds(60))
        timer?.resume()
    }
}
```

The `OneMinuteDelayProcedureB` procedure adds a DidCancel observer to handle the cancellation event. In the observer block, it:

1. Cancels the procedure's timer.

2. Explicitly calls `finish()` (since cancelling the timer prevents it from calling its event handler, which is the only place that `finish()` is called).

### Will/DidCancel observers VS overriding procedureWill/DidCancel():

The main differences between these two options are:
- A Procedure can have multiple Will/DidCancel observers.
- Will/DidCancel observers can be added to a specific instance of a Procedure.
- A slight difference in the order they are called, as outlined above.
- Observers cannot be removed / overridden / usurped by a subclass.

You can use a Will/DidCancel observer in all the places that you might use procedureWillCancel/procedureDidCancel().

> ### Why is `finish()` conditioned on `isExecuting`?

> Finished Procedures cannot be added to a queue.

> Since `cancel()` can be called at any time, guarding on `isExecuting` in the DidCancel observer block ensures that a `OneMinuteDelayProcedureB` procedure that is cancelled **before it is added to a queue or executed** does not finish immediately.

> This ensures that a `OneMinuteDelayProcedureB` can be cancelled and then added to a queue (where, when it executes, it will automatically handle finishing).

> For safety, the framework follows the above rule in its Procedures. If you ensure that your custom Procedures are never cancelled before they are added to a queue, this is not necessary.

## Cancellation and Thread-Safety:

A Procedure can be cancelled on any thread. Just like all other observers, you should be prepared for your Will/DidCancel observers (or your procedureWillCancel/procedureDidCancel() overrides) to be called at any time, on any thread.

danthorpe:
@swiftlyfalling @jshier - I've not looked though all this in detail yet. I appreciate that a breaking change to cancellation/finishing behaviour should go into v4 betas asap - but I think I'd prefer to get Beta 5 done before introducing this.

So, at the moment, I'm going to come back to this issue at the start of next week.

Sources/Network/NetworkData.swift
```diff
@@ -64,11 +67,21 @@ open class NetworkDataProcedure<Session: URLSessionTaskFactory>: Procedure, Inpu
...
-            guard !isCancelled else { return }
+            guard !isCancelled else {
+                finish()
+                return
+            }
             task = session.dataTask(with: request) { [weak self] data, response, error in
                 guard let strongSelf = self else { return }
 
                 if let error = error {
+                    let nsError = error as NSError
```
jshier: FYI, the appropriate way to perform this check is to cast to `URLError` and check for the `.cancelled` case.

```swift
if let error = error {
    if strongSelf.isCancelled, let error = error as? URLError, error == .cancelled {
        strongSelf.finish()
        return
    }

    strongSelf.finish(withResult: .failure(error))
    return
}
```

Issue #551, opened 2016-10-27
Investigate: ReverseGeocodeUserLocationProcedureTests


danthorpe:
We've seen some random test failures.

swiftlyfalling:
I have been able to very sporadically reproduce an EXC_BAD_ACCESS crash in these tests locally. Trying to get something consistent...

danthorpe:
Gonna close this for now - as likely fixed by #572.

PR #572, opened 2016-11-04
TestableReverseGeocoder should call completionHandler block on the main thread

swiftlyfalling:
To replicate [CLGeocoder.reverseGeocodeLocation(_:completionHandler:)](https://developer.apple.com/reference/corelocation/clgeocoder/1423621-reversegeocodelocation), the completion block must be called on the main thread.


swiftlyfalling:
Speaking of #551, yet another error in:
ReverseGeocodeProcedureTests.test__completion_is_executed_and_receives_placemark()

That went away re-running the location test suite.

danthorpe:
Thanks @swiftlyfalling!

PR #573, opened 2016-11-04
Profiler

yageek:
I think we have the basics: translation of the `Profiler` class and some basic tests.

As mentionned within the second commit message, I have some trouble with the commented test with `GroupProcedure` and can not figure out why it is failing.

### DIFF
--- a/Sources/Profiler.swift
+++ b/Sources/Profiler.swift
@@ -22,9 +22,10 @@ extension ProcedureEvent: CustomStringConvertible {
         }
     }
 }
+
 // MARK: - ProcedureProfilerReporter
 public protocol ProcedureProfilerReporter {
-    func finishedProfilingWithResult(result: ProfileResult)
+    func finishedProfiling(withResult result: ProfileResult)
 }
 
 // MARK: - ProfileResult
@@ -38,30 +39,7 @@ public struct ProfileResult {
     public let children: [ProfileResult]
 }
 
-extension PendingValue where T: Equatable {
-    var pending: Bool {
-        if case .pending = self {
-            return true
-        }
-        return false
-    }
-
-    static func==(lhs: PendingValue<T>, rhs: PendingValue<T>) -> Bool {
-        switch (lhs, rhs) {
-        case (.pending, .pending):
-            return true
-        case (.void, .void):
-            return true
-        case let (.ready(lhsValue), .ready(rhsValue)):
-            return lhsValue == rhsValue
-        default:
-            return false
-        }
-    }
-}
-
 // MARK: - PendingResult
-
 struct PendingResult {
 
     let created: TimeInterval
@@ -72,12 +50,12 @@ struct PendingResult {
     let finished: PendingValue<TimeInterval>
     let children: [ProfileResult]
 
-    var pending: Bool {
-        return identity.pending || attached.pending || started.pending || (cancelled.pending && finished.pending)
+    var isPending: Bool {
+        return identity.isPending || attached.isPending || started.isPending || (cancelled.isPending && finished.isPending)
     }
 
     func createResult() -> ProfileResult? {
-        guard !pending,
+        guard !isPending,
             let
             identity = identity.value,
             let attached = attached.value,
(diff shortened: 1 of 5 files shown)


swiftlyfalling:
The broken GroupProcedure test is due to a bug in adding observers that conform to both GroupObserverProtocol and ProcedureObserver to a GroupProcedure. See #574.

yageek:
This is what I suspected. Thanks for giving an evidence :)

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+        case .started: return "Started"
+        case .cancelled: return "Cancelled"
+        case .produced: return "Produced"
+        case .finished: return "Finished"
+        }
+    }
+}
+// MARK: - ProcedureProfilerReporter
+public protocol ProcedureProfilerReporter {
+    func finishedProfilingWithResult(result: ProfileResult)
```
danthorpe: I think a Swift 3.0 naming here would be:

```swift
func finishedProfiling(withResult result: ProfileResult)
```

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+    public let children: [ProfileResult]
+}
+
+extension PendingValue where T: Equatable {
+    var pending: Bool {
+        if case .pending = self {
+            return true
+        }
+        return false
+    }
```
danthorpe: Lets take this computed property out of this extension, and add it to `PendingValue` directly in *ResultInjection.swift*, and rename it:

```swift
public enum PendingValue {
    // etc
    public var isPending: Bool {
        if case .pending = self { 
            return true
        }
    }
}
```

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+
+    static func==(lhs: PendingValue<T>, rhs: PendingValue<T>) -> Bool {
+        switch (lhs, rhs) {
+        case (.pending, .pending):
+            return true
+        case (.void, .void):
+            return true
```
danthorpe: Can we put these onto one line..

```swift
case (.pending, .pending), (.void, .void): 
    return true
```

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+    let created: TimeInterval
+    let identity: PendingValue<Procedure.Identity>
+    let attached: PendingValue<TimeInterval>
+    let started: PendingValue<TimeInterval>
+    let cancelled: PendingValue<TimeInterval>
+    let finished: PendingValue<TimeInterval>
+    let children: [ProfileResult]
+
+    var pending: Bool {
+        return identity.pending || attached.pending || started.pending || (cancelled.pending && finished.pending)
```
danthorpe: And I guess this should be renamed `isPending` - this is just to be more consistent with `Foundation.Operation` properties, e.g. `isFinished` etc.

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+        guard cancelled.pending else { return self }
+        return PendingResult(created: created, identity: identity, attached: attached, started: started, cancelled: .ready(now - created), finished: finished, children: children)
+    }
+
+    func finish(now: TimeInterval = CFAbsoluteTimeGetCurrent() as TimeInterval) -> PendingResult {
+        guard finished.pending else { return self }
+        return PendingResult(created: created, identity: identity, attached: attached, started: started, cancelled: cancelled, finished: .ready(now - created), children: children)
+    }
+
+    func addChild(child: ProfileResult) -> PendingResult {
```
danthorpe: I think should now be renamed:

```swift
func add(child: ProfileResult) -> PendingResult {
    // etc
}
```

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+            let
+            identity = identity.value,
+            let attached = attached.value,
+            let started = started.value
+            else { return .none }
+
+        return ProfileResult(identity: identity, created: created, attached: attached, started: started, cancelled: cancelled.value, finished: finished.value, children: children)
+    }
+
+    func setIdentity(newIdentity: Procedure.Identity) -> PendingResult {
```
danthorpe: And I think this should be renamed:

```swift
func set(identity newIdentity: Procedure.Identity) -> PendingResult 
```

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// MARK: ProcedureProfiler
+public final class ProcedureProfiler: Identifiable, Equatable {
+
+    public let identifier = UUID()
+
+    enum Reporter {
+        case Parent(ProcedureProfiler)
+        case Reporters([ProcedureProfilerReporter])
```
danthorpe: Can we rename these to be Swift 3.0 style, 😁

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+
+    let queue = DispatchQueue(label: "me.danthorpe.ProcedureKit.Profiler")
```
danthorpe: Can we change the queue label to be: `run.kit.ProcedureKit.Profiler`

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+    let reporter: Reporter
+
+    var result = PendingResult(created: CFAbsoluteTimeGetCurrent() as TimeInterval, identity: .pending, attached: .pending, started: .pending, cancelled: .pending, finished: .pending, children: [])
+    var children: [Procedure.Identity] = []
+    var finishedOrCancelled = false
+
+    var pending: Bool {
+        return result.pending || (children.count > 0)
```
danthorpe: Can we also rename to `isPending` please

Sources/Profiler.swift
```diff
@@ -0,0 +1,326 @@
...
+            case .attached:
+                self.result = self.result.attach(now: now)
+            case .started:
+                self.result = self.result.start(now: now)
+            case .cancelled:
+                self.result = self.result.cancel(now: now)
+                self.finishedOrCancelled = true
+            case .finished:
+                self.result = self.result.finish(now: now)
+                self.finishedOrCancelled = true
```
danthorpe: I'm thinking that because this is a `.sync` method, that we don't need to capture `self` and can just clean this up to be:

```swift
queue.sync {
    switch event {
        case .attached:
            result = result.attach(now: now)
        // etc
```

yageek:
I updated the pull request. Sorry for the wrong naming, my brain has not completed the migration to Swift 3.0 yet :sweat_smile:.

danthorpe:
Thanks @yageek for this - will merge now, and then look into this issue with the `GroupProcedure` and the failing test. 👍😀

Issue #576, opened 2016-11-05
NetworkObserver thread-safety


swiftlyfalling:
NetworkObserver in ProcedureKit 4.0 beta 4 is not currently thread-safe, because NetworkActivityController's `start()` and `stop()` methods are not currently thread-safe.

This can result in crashes and incorrect behavior if multiple Procedures with NetworkObservers perform the right state transitions (and observer callbacks) concurrently.

PR #577, opened 2016-11-09
NetworkObserver thread-safety (#576)

swiftlyfalling:
NetworkObserver in ProcedureKit 4.0 beta 4 is not currently thread-safe, because NetworkActivityController's start() and stop() methods are not currently thread-safe.

This can result in crashes and incorrect behavior if multiple Procedures with NetworkObservers perform the right state transitions (and observer callbacks) concurrently.

Also fixes some of the associated tests.

### DIFF
--- a/Tests/Mobile/NetworkObserverTests.swift
+++ b/Tests/Mobile/NetworkObserverTests.swift
@@ -78,9 +78,9 @@ class NetworkObserverTests: ProcedureKitTestCase {
     }
 
     func test__network_indicator_only_changes_once_when_multiple_procedures_start() {
-        let procedure1 = TestProcedure(delay: 1.0)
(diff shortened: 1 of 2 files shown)


Tests/Mobile/NetworkObserverTests.swift
```diff
@@ -77,19 +78,22 @@ class NetworkObserverTests: ProcedureKitTestCase {
     }
 
     func test__network_indicator_only_changes_once_when_multiple_procedures_start() {
-        let procedure1 = TestProcedure()
+        let procedure1 = TestProcedure(delay: 1.0)
```
danthorpe: Really minor quibble, but can we reduce these delays a couple of orders of magnitude? Just trying to keep the execution time of the tests down.

PR #579, opened 2016-11-10
Additional GroupProcedure tests

swiftlyfalling:
Tests to ensure that the base GroupProcedure ProcedureQueue delegate implementation ignores delegate callbacks from other queues.


danthorpe:
@swiftlyfalling - awesome!

Issue #574, opened 2016-11-04
Bug: Adding observers that conform to both GroupObserverProtocol and ProcedureObserver to a GroupProcedure


swiftlyfalling:
GroupProcedure has the following function to enable adding `GroupObserverProtocol` observers:

```swift
func add(observer: GroupObserverProtocol) {
    groupProtectedObservers.append(observer)
}
```

However, if adding an observer that conforms to **both** `GroupObserverProtocol` and `ProcedureObserver` (as in the case of ProcedureProfiler - see #573), this breaks the `ProcedureObserver` part of the observer.

`GroupProcedure.add(observer: GroupObserverProtocol)` gets called, and the observer only gets added to the groupProtectedObservers (and does not get added to the underlying Procedure observers).

yageek:
Any suggestion to address this ?

swiftlyfalling:
This may involve a bit of refactoring, so will wait on @danthorpe's input. (May be a short while.)

danthorpe:
Hmm - will have a think. We kinda want to just call `super.add(observer: theObserver)` right? But the generics I guess make this a bit complicated? Is that the key issue here?

yageek:
I tried to make a call to `super.add(observer: theObserver)` and yes, the generics did not help as `GroupObserver` is not related explicitly to `ProcedureObserver`.

danthorpe:
Okay, I've been thinking about this, and I think the best solution is to combine the two protocols. Essentially, we have `ProcedureObserver` which already has:

```swift
func procedure(_ procedure: Procedure, didProduce newOperation: Operation)
```

and we have `GroupObserverProtocol` which has:

```swift
    func group(_ group: GroupProcedure, willAdd: Operation)

    func group(_ group: GroupProcedure, didAdd: Operation)
```

Given that `ProcedureObserver` is generic, as the `Procedure` argument is an associated type, when used from `GroupProcedure`, it is essentially exactly the same as `group(: didAdd:)`.

So, here's my proposal - get rid of `GroupObserverProtocol` entirely, and modify `ProcedureObserver` so that it has:

```swift
func procedure(_ procedure: Procedure, willAdd newOperation: Operation)
func procedure(_ procedure: Procedure, didAdd newOperation: Operation)
```

which `GroupProcedure` will use when new child operations are added, and `Procedure` will also use when the `produce(operation: Operation)` API is used - although I'm actually tempted to deprecate this API entirely.

Thoughts?

[...]

PR #583, opened 2016-11-15
Changes to address GroupProcedure observer bug

danthorpe:
As discussed in #574.

### DIFF
--- a/ProcedureKit.xcodeproj/project.pbxproj
+++ b/ProcedureKit.xcodeproj/project.pbxproj
@@ -914,6 +914,7 @@
 				6583F95E1D8C4C7F00000A8D /* NoFailedDependenciesConditionTests.swift */,
 				653C9FE81D6099F10070B7A2 /* ProcedureKitTests.swift */,
 				65245D1C1D721A7100340A2D /* ProcedureTests.swift */,
+				0B7922301DCCD88A000AC3CA /* ProfilerTests.swift */,
 				6594840F1DA98E700028F83B /* ReduceProcedureTests.swift */,
 				655594F01D9E482D00B08990 /* RepeatProcedureTests.swift */,
 				65403ABD1D7C9E7600D6036B /* ResultInjectionTests.swift */,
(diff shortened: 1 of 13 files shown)


swiftlyfalling:
@danthorpe: I like how this is evolving.

As a point of clarification: previously, if a GroupProcedure child produced an operation, that operation would be added to the Group's queue, correct? (i.e. as a child of the Group?)

danthorpe:
@swiftlyfalling we did have this special handling previously in _Operations_ (although not originally, I think it was introduced in version 2). However - in the v4 rewrite of `ProcedureQueue` I didn't include this - as there was already some discussions that it wasn't consistent behaviour.

However, that if you have a `GroupProcedure`, and a child in that group (i.e. on the group's queue) calls `produce()`, then this new operation will get added to the receivers queue, which *is* the group's queue. However... the thing is, it will not go through the `add(child: )` API, and so the group will essentially be unaware of it - i.e. it will not wait for it to finish.

This is not a bug, or a defect - that is the intended behaviour of `produce()`, while `add(child:)` is for adding a new child.

swiftlyfalling:
@danthorpe:

With the changes in this PR, a child of a Group that produces an operation will result in the produced operation being waited on by the Group.

(The Group's `procedureQueue(_ queue: ProcedureQueue, willAddOperation operation: Operation)` is called with the produced operation.)

danthorpe:
@swiftlyfalling - I don't think so, in the observer in `ProcedureQueue` there are explicit checks for this [here](https://github.com/ProcedureKit/ProcedureKit/blob/development/Sources/ProcedureQueue.swift#L172).

danthorpe:
In terms of a group waiting on _produced_ operations - I guess it'll be best to double check this with some more unit tests - as we don't have that specific test. I'll create another issue to investigate.

swiftlyfalling:
@danthorpe: Please see PR #588. Added a simple test that fails.

swiftlyfalling:
If we want to ensure that:

1. Produced operations do not get waited on by the Group.

2. Produced operations do not trigger multiple calls of observers.

Then we need to ensure that the group does not execute its `willAddOperation` QueueDelegate handler for operations produced by children.

Issue #580, opened 2016-11-13
MutuallyExclusive with arbitrary category


pronebird:
Is it possible to use MutuallyExclusive condition to group operations based on arbitrary category?

I see that MutuallyExclusive simply uses class name, but would that be possible to have a condition for sub-type of operation, i.e MyClass.Action1, MyClass.Action2?

I can imagine that I can achieve the same with observers and simply chain procedures, but thought that MutuallyExclusive should have similar function implemented already.

danthorpe:
Hi @pronebird. At the moment, this is not possible, as the `category` property is only defined in an internal extension on `ConditionProtocol`.

However, I think, this could be changed to support arbitrary category groupings. In such a scenario, you could then just attach a `TrueCondition` or `BlockCondition` with a custom category.

pronebird:
Hi @danthorpe,

Thanks for suggestion. Looking at TrueCondition's code, if I get it right, I can simply provide arbitrary name and set mutuallyExclusive = true, then it should work the same way as mutual exclusivity right?

pronebird:
Am I right to think that this should work?

```swift
let exclusiveCondition = TrueCondition(name: "MutuallyExclusive<\(self)>.washingMachine1.doLaundry", mutuallyExclusive: true)

let operation = FetchExchangeRateProcedure()
operation.add(condition: exclusiveCondition)
        
operationQueue.addOperation(operation)
```

I am not sure whether it makes the condition mutually exclusive or entire operation..

pronebird:
Ugh, I just skimmed through the code and what you said makes sense. ConditionProtocol defines category but it's internal.

danthorpe:
Yeah, so the name argument of `TrueCondition` is just the name of the underlying `Operation`, not the category - but this is what I'm suggesting will be changed.

PR #587, opened 2016-11-17
Support Arbitrary mutual exclusion category names

danthorpe:
As discussed in #580.

### DIFF
--- a/Sources/Capability.swift
+++ b/Sources/Capability.swift
@@ -249,10 +249,10 @@ public class AuthorizedFor<Status: AuthorizationStatus>: Condition {
 
     fileprivate let capability: AnyCapability<Status>
 
-    public init<Base>(_ base: Base) where Base: CapabilityProtocol, Status == Base.Status {
+    public init<Base>(_ base: Base, category: String? = nil) where Base: CapabilityProtocol, Status == Base.Status {
         capability = AnyCapability(base)
         super.init()
-        mutuallyExclusiveCategory = String(describing: base)
+        mutuallyExclusiveCategory = category ?? String(describing: type(of: base))
         add(dependency: AuthorizeCapabilityProcedure(base))
     }
 
--- a/Sources/Capability.swift
+++ b/Sources/Capability.swift
@@ -252,7 +252,7 @@ public class AuthorizedFor<Status: AuthorizationStatus>: Condition {
     public init<Base>(_ base: Base) where Base: CapabilityProtocol, Status == Base.Status {
         capability = AnyCapability(base)
         super.init()
-        mutuallyExclusive = true
+        mutuallyExclusiveCategory = String(describing: base)
         add(dependency: AuthorizeCapabilityProcedure(base))
     }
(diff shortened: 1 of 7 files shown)


pronebird:
So using the following we can set up mutually exclusive operations with arbitrary categories?

```swift
let mutuallyExclusiveWithArbitraryCategory = TrueCondition(name: "Condition 1", mutuallyExclusiveCategory: "Condition 1")
```

danthorpe:
Yes. You can do something like this:

```swift
let foo = Foo()
foo.add(condition: TrueCondition(mutuallyExclusiveCategory: "Red"))
let bar = Bar()
bar.add(condition: TrueCondition(mutuallyExclusiveCategory: "Red"))
let baz = Baz()
baz.add(condition: TrueCondition(mutuallyExclusiveCategory: "Green"))
let bat = Bat()
bat.add(condition: TrueCondition(mutuallyExclusiveCategory: "Green"))
bat.add(condition: TrueCondition(mutuallyExclusiveCategory: "Blue"))
```

danthorpe:
@pronebird feedback on this would be much appreciated - especially when adding multiple exclusive conditions - the test coverage around those scenarios is probably a bit weak, so might want to bear that in mind.

pronebird:
@danthorpe cool. As of now I still maintain a dictionary of category => operations. I'll probably convert it to TrueCondition and let you know if I hit any issues.

pronebird:
Works on a very limited test case.

```swift
    func testSerialExecutionForSimilarRequests() {
        var counter = 0
        
        var expectations = [XCTestExpectation]()
        
        for i in 0..<50 {
            let expect = expectation(description: "Wait for fetch \(i)")
            
            expectations.append(expect)
            
            runOperationWithCategory("MyCategory") { getResult in
                let _ = try! getResult()
                
                DispatchQueue.main.async {
                    print("Finished query \(i)")
                    
                    XCTAssertEqual(counter, i)
                    counter += 1
                    
                    expectations[i].fulfill()
                }
            }
        }
        
        waitForExpectations(timeout: 20)
    }
```

ryanjm:
Using the same setup as before I get this error:

> Type of expression is ambiguous without more context

For this line of code:

```
clearCoreData.add(condition: CoreDataExclusive())
```

My `CoreDataExclusive` is defined as:

```
public enum CoreDataWork { }
public typealias CoreDataExclusive = MutuallyExclusive<CoreDataWork>
```

Any ideas why this error would show up?

swiftlyfalling:
@ryanjm: I am unable to reproduce that error using the latest `development` branch and `Xcode 8.2.1 (8C1002)`. Could you try with `development`?

ryanjm:
Sorry @swiftlyfalling, my fault. I forgot to change my `clearCoreData` from a `BlockOperation` to `BlockProcedure`.

Thank you for taking a look.

swiftlyfalling:
@ryanjm: Ah, yup - that would do it!

PR #563, opened 2016-11-17
ReachableProcedure

danthorpe:
- [x] Adds `Reachable` API wrappers back
- [x] Define protocols for constraints. Definitely require an error protocol, so that we can know that the procedure finishes with a network error.
- [x] Create a `RetryProcedure` subclass. Something like this:

```swift
final public class ReachableProcedure<T: ProcedureProtocol>: RetryProcedure<T> 
    where T: AssociatedErrorProtocol, T.AssociatedError: ReachableNetworkProcedureError {
    // etc
}
```

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -8,4 +8,11 @@ disabled_rules:
 file_length:
   warning: 500
   error: 1200
-  
+variable_name:
+  excluded:
+    - id
+    - URL
+    - of
+    - in
+    - x
+    - y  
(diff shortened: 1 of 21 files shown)


danthorpe:
This is certainly a stepping point for _ProcedureKitNetwork_. There are still some bits to be sorted out however which will be done in separate issues.

## Incorporate network resiliency and reachability
    This will likely mean either, we have a single "network retry procedure" which incorporates handling of both reachability & other errors using the resilient behavior protocol. Or, the resiliency procedure is a subclass of this reachable one, which I think is less desirable.

## ProcedureKitNetworkTestCase
    There is quite a lot of faffing about in the test cases here - so all the network testing needs a bit of refactoring on how to make it easy to test these classes, and for consumers to test their procedures which may compose them.

PixelPirate:
Is there any alternative for `ReachableProcedure` in the current version of ProcedureKit?

danthorpe:
Hi @PixelPirate - yup, its just got renamed to `NetworkProcedure`. See #561 for details. If you have any issues with `NetworkProcedure` - please create a new issue, would like any feedback on this 😊

PR #588, opened 2016-11-17
Producing operations from Group children fails

swiftlyfalling:
Test fails with:

`assertion failed: Attempting to perform illegal cyclic state transition, executing -> pending`

### DIFF
--- a/Sources/Errors.swift
+++ b/Sources/Errors.swift
@@ -29,7 +29,7 @@ public struct ProcedureKitError: Error, Equatable {
 
         public static func == (lhs: Context, rhs: Context) -> Bool {
             switch (lhs, rhs) {
-            case (.unknown, .unknown), (.dependencyFinishedWithErrors, .dependencyFinishedWithErrors), (.parentCancelledWithErrors, .parentCancelledWithErrors), (.requirementNotSatisfied, .requirementNotSatisfied):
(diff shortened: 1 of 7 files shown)


danthorpe:
@swiftlyfalling great! Thanks!

swiftlyfalling:
Copied from the other (now closed) PR (to keep everything in one place):

@danthorpe [said](https://github.com/ProcedureKit/ProcedureKit/pull/583#issuecomment-261365072):
> @swiftlyfalling - I don't think so, in the observer in ProcedureQueue there are explicit checks for this here.

The problem is that:

1. In the first call to the queue's `WillAddOperationBlockObserver`, the produced operation is not yet enqueued, so the check passes.
2. This then calls `queue.delegate?.procedureQueue(queue, willAddOperation: operation)`, which enters the Group's delegate handler.
3. The Group's delegate handler *calls all the same observers again*, (`observers.forEach { $0.procedure(self, willAdd: operation) }`) which results in the queue's `WillAddOperationBlockObserver` being called for a second time.
4. In the **second** call to the queue's `WillAddOperationBlockObserver`, the produced operation is not yet enqueued, so the check passes.
5. This then ultimately enqueues the produced operation.
6. Ultimately, the Group's delegate handler returns.
7. And then the **first** call to the queue's `WillAddOperationBlockObserver` tries to proceed with enqueuing the produced operation... again.

If we want to ensure that:

1. Produced operations do not get waited on by the Group.

2. Produced operations do not trigger multiple calls of observers.

Then we need to ensure that the group does not execute its `willAddOperation` QueueDelegate handler for operations produced by children.

PR #593, opened 2016-11-21
Support merging collections of Procedures which conform to ResultInjection

danthorpe:


danthorpe:
This PR adds convenience APIs to `Collection` where the elements are procedures which conform to `ResultInjection` protocol. Usage would be like this:

### Gather the results of the collection of procedures  into an array:
```swift
let gathered = myCollection.gather()

gathered.addDidFinishBlockObserver { gathered, _ in
    // Will print the array of result values from the collection
    print(gathered.result.value)
}

queue.add(operations: myCollection)
queue.add(operation: gathered)
```

### Reduce the results of the collection of procedures into a single type
```swift
let reduced = myCollection.reduce(initialValue) { accumulator, element in
    // etc
}

reduced.addDidFinishBlockObserver { reduced, _ in
    // Will print the result of combining the results from the collection
    print(reduced.result.value)
}

queue.add(operations: myCollection)
queue.add(operation: reduced)
```

PR #594, opened 2016-11-21
Improve how stress tests are run in CI

danthorpe:
We need to 
- [x] run stress tests on the same dedicated build machine
- [x] run stress tests in release configuration
- [x] run stress tests with Thread Sanitizer disabled


danthorpe:
FYI @swiftlyfalling

Issue #555, opened 2016-10-27
Stress Tests: Batch.counters (and associated methods) are not thread-safe.


swiftlyfalling:
This can result in occasional (random) StressTest failures (or testing crashes), because multiple threads may attempt to call `Batch.counter(named:)` or `Batch.incrementCounter(named:withBarrier:)` simultaneously, mangling the `counters` dictionary.

Simply wrapping the `counters` dictionary in a Protector would fix the issue, but would also add a lock to a hot path that was originally intended to avoid them (with the Counter class and OSAtomicIncrement32).

However, OSAtomicIncrement32 and friends are deprecated as of macOS 10.12/iOS 10.10. And while Xcode 8 suggests to use functions from `<stdatomic.h>` instead, doing so in Swift is [currently problematic](https://openradar.appspot.com/27161329). So maybe the best solution for now is to wrap things in a Protector (and then forget about OSAtomicIncrement32)? We'll need to test to see how this impacts the stress tests' speed / other characteristics.

PR #596, opened 2016-11-21
Fix StressTestCase Batch counters thread-safety (#555)

swiftlyfalling:
Per #555, since OSAtomicIncrement32 and friends are deprecated, moved this to a simple lock-based implementation that fixes the issues.

### DIFF
--- a/Sources/Testing/StressTestCase.swift
+++ b/Sources/Testing/StressTestCase.swift
@@ -6,24 +6,7 @@
 
 import Foundation
 import XCTest
-
-public class Counter {
-    private(set) var _count: Int32 = 0 // swiftlint:disable:this variable_name
-
-    public var count: Int {
-        return Int(_count)
-    }
-
-    @discardableResult public func increment() -> Int32 {
-        return OSAtomicIncrement32(&_count)
-    }
-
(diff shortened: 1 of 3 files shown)


Sources/Testing/StressTestCase.swift
```diff
@@ -6,24 +6,7 @@
...
-        return OSAtomicIncrement32(&_count)
-    }
-
-    @discardableResult public func barrierIncrement() -> Int32 {
-        return OSAtomicIncrement32Barrier(&_count)
-    }
-
-    public init() { }
-}
+@testable import ProcedureKit
```
danthorpe: This is a bit of a problem - it means that we can't create Release configuration builds of `TestingProcedureKit`, so, for example integration via Carthage will fail.

I'm guessing it's the `.withCriticalScope { }` API that you need? I have no issue making that public.

swiftlyfalling: Ah! Did not realize that. And yes, `.withCriticalScope { }` is what was needed.

danthorpe: I'll fix it, no problem.

PR #598, opened 2016-11-22
Procedure & Support thread-safety fixes

swiftlyfalling:
Fixes issues identified by an audit and Thread Sanitizer.

`add(directDependency:)`
`remove(directDependency:)`
`add(condition:)`
are now thread-safe.

In addition, accessing the various properties are now thread-safe.

And a data race that was not prevented by the existing use of locks surrounding `_error` is resolved.

### DIFF
--- a/Sources/Support.swift
+++ b/Sources/Support.swift
@@ -4,6 +4,8 @@
 //  Copyright © 2016 ProcedureKit. All rights reserved.
 //
 
+import Dispatch
+
 internal func _abstractMethod(file: StaticString = #file, line: UInt = #line) {
     fatalError("Method must be overriden", file: file, line: line)
 }
--- a/Sources/Support.swift
+++ b/Sources/Support.swift
@@ -75,12 +75,8 @@ public class Protector<T> {
         return read { $0 }
     }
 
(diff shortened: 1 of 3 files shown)


Sources/Support.swift
```diff
@@ -24,13 +26,14 @@ extension Dictionary {
 
 protocol ReadWriteLock {
     mutating func read<T>(_ block: () throws -> T) rethrows -> T
-    mutating func write(_ block: @escaping () -> Void, completion: (() -> Void)?)
+    mutating func write_async(_ block: @escaping () -> Void, completion: (() -> Void)?)
+    mutating func write_sync<T>(_ block: () throws -> T) rethrows -> T
```
danthorpe: I'm thinking that it might be better to follow Dispatch here. The API for `read<T>` is the same as that for `write_sync<T>` so, how about we do this instead:

```swift
protocol ReadWriteLock {
    mutating func sync<T>(_ block: () throws -> T) rethrows -> T
    mutating func async(_ block: @escaping () -> Void, completion: (() -> Void)?)
}

swiftlyfalling: Yeah, that was primarily why I did not change it to sync/async.

PR #599, opened 2016-11-22
CKProcedure has internal properties which should be public

danthorpe:


danthorpe:
Thanks @leotumwattana!

PR #601, opened 2016-11-22
[Breaking Changes]: Refactor of Result Injection APIs

danthorpe:
At the moment, `PendingValue<T>` is a structure which can represent a value which isn't yet ready. i.e. `.pending` vs `.ready(value)`. However, we also want to be able to support `.failed(error)` or equivalent.

We'll take inspiration from the excellent [Result](https://github.com/antitypical/Result) type here, and refactor to include as simple `Result`:

```swift
enum Result<T> {
    case success(T)
    case failure(Error)
}
```


danthorpe:
## Description of the Breaking Changes

Before, there was the protocol `ResultInjection`:

```swift
public protocol ResultInjection {
    associatedtype Requirement
    associatedtype Result

    var requirement: PendingValue<Requirement> { get set }
    var result: PendingValue<Result> { get set }
}
```

This has been changed (back) to two protocols, with a rename of the properties.

### `InputProcedure`

```swift
public protocol InputProcedure: ProcedureProtocol {
    associatedtype Input

    var input: Pending<Input> { get set }
}
```

### `OutputProcedure`

```swift
public protocol OutputProcedure: ProcedureProtocol {
    associatedtype Output

    var output: Pending<Result<Output>> { get }
}
```

where `Pending` is mostly `PendingValue` as before, and `Result` is a pretty single either type for `success(T) and `failure(Error)`.

## Ramifications of this change

1. The `output` of a procedure which conforms to `OutputProcedure` is now explicit with regard to its ready state. e.g. if finishing with a successful output value:
    ```swift
    output = .ready(.success(value))
    ```
    if finishing with an error
    ```swift
    output = .ready(.failure(anError))
    ```

2. A `.pending` output means that that the procedure has not completed executing, i.e. either it hasn't started yet, or it has been cancelled.

3. `Condition` has been refactored, as now `Result<Bool>` represents `ConditionResult`, where
    - `.success(true)` indicates a _satisfied_ condition
    - `.success(false)` indicates an _ignored_ condition
    - `.failure(error)` indicates a _failed_ condition

4. `BlockCondition` has a breaking change. To *fail* the attached procedure, the block must throw and error. Returning `false` from the block will result in an ignored state, meaning that the procedure will be cancelled without an error.

danthorpe:
@swiftlyfalling @jshier @yageek reviews welcome on this 😁

Sources/Reduce.swift
```diff
@@ -4,56 +4,22 @@
 //  Copyright © 2016 ProcedureKit. All rights reserved.
 //
 
-open class ReduceProcedure<Element, U>: Procedure, ResultInjection {
-
-    public let initial: U
-    public let nextPartialResult: (U, Element) throws -> U
-    public var requirement: PendingValue<AnySequence<Element>> = .pending
-    public var result: PendingValue<U> = .pending
+open class ReduceProcedure<Element, U>: TransformProcedure<AnySequence<Element>, U> {
```
danthorpe: Refactored `ReduceProcedure` to subclass `TransformProcedure`

Sources/Testing/RepeatTestCase.swift
```diff
@@ -45,7 +45,10 @@ open class RetryTestCase: ProcedureKitTestCase {
         let info = RetryTestCaseInfo()
         return AnyIterator {
             let procedure = TestProcedure()
-            procedure.add(condition: BlockCondition { info.numberOfFailures == failureThreshold })
+            procedure.add(condition: BlockCondition {
+                guard info.numberOfFailures == failureThreshold else { throw ProcedureKitError.conditionFailed() }
+                return true
+            })
```
danthorpe: This change is a result of the changed behaviour of `BlockCondition` - not 100% sure on this breaking change.

danthorpe:
In terms of finishing, at the moment, these changes just introduce the use of `Result`. But, we can/will add a convenience API for finishing. Something like this:

```swift
public extension OutputProcedure {

    func finish(withResult result: Result<Output>) {
        // Sets the output property
        output = .ready(result)
        // Finish the procedure, using any optional error from the result
        finish(withError: result.error)
    }
}
```

I think I prefer maintaining a consistent usage - i.e. subclass is always required to called `finish()`, over automatically calling `finish()` when the output is set.

An alternative _might_ be something like this:

```swift
public extension OutputProcedure {

    var output: Pending<Result<Output>> {
        didSet {
            guard !isFinished else { return } // this is actually not good enough
            finish(withError: output.error)
        }
    }
}
```

But the finishing behaviour of `Procedure` is pretty involved, and we'd now have to deal with setting the output and calling finish concurrently.

Additionally, it is helpful to be able to set `output` periodically while a procedure is running, see `UserLocationProcedure` for example.

yageek:
I'm asking because inside the new `NetworkDataProcedure` we have this:

```swift
if let error = error {
                strongSelf.output = .ready(.failure(ProcedureKitNetworkError(error as NSError)))
                return
            }
``` 

There is no call to `finish()` and only an update on the `output` property.

danthorpe:
@yageek there is a `defer` statement:

```swift
defer { strongSelf.finish(withError: strongSelf.output.error) }
```

yageek:
My bad :( Not really awake this morning.

danthorpe:
@yageek no worries! But yeah, essentially adding `finish(withResult: )` would actually improve the readability in cases like this.

PR #604, opened 2016-11-22
OPR-595 NetworkProcedure Progress

blg-andreasbraun:
There are no tests right now. Unfortunately I have not figured out how to properly test this functionality.
If someone has ideas, I'd like to hear them :)

### DIFF
--- a/Sources/Procedure.swift
+++ b/Sources/Procedure.swift
@@ -586,18 +586,6 @@ open class Procedure: Operation, ProcedureProtocol {
         didChangeValue(forKey: .finished)
     }
 
-    /**
-     Public override which deliberately crashes your app, as usage is considered an antipattern
-
-     To promote best practices, where waiting is never the correct thing to do,
-     we will crash the app if this is called. Instead use discrete operations and
-     dependencies, or groups, or semaphores or even NSLocking.
-
-     */
-    public final override func waitUntilFinished() {
-        fatalError("Waiting on operations is an anti-pattern. Remove this ONLY if you're absolutely sure there is No Other Way™. Post a question in https://github.com/danthorpe/Operations if you are unsure.")
-    }
-
     // MARK: - Observers
 
     /**


blg-andreasbraun:
@danthorpe - I can't see whats wrong with the buildkite checks... Am I missing something?

danthorpe:
@blg-andreasbraun - all the network CI jobs have failed because: 

> ProcedureKitNetworkTests.swift:12:1: type 'TestableURLSessionTask' does not conform to protocol 'URLSessionTaskProgressProtocol'

danthorpe:
@blg-andreasbraun - you can DM me on Twitter @danthorpe your email address, and I can invite you to a Slack team where _ProcedureKit_ discussions happen 😀

Sources/Network/NetworkUpload.swift
```diff
@@ -63,7 +72,31 @@ open class NetworkUploadProcedure<Session: URLSessionTaskFactory>: Procedure, Re
...
+            return
+        }
+
+        if let keyPath = keyPath, let task = object as? Session.UploadTask {
+            switch keyPath {
+            case #keyPath(URLSessionTaskProgressProtocol.countOfBytesExpectedToSend):
+                progress.totalUnitCount = task.countOfBytesExpectedToSend
+            case #keyPath(URLSessionTaskProgressProtocol.countOfBytesSent):
+                progress.completedUnitCount = task.countOfBytesSent
+            case #keyPath(URLSessionTaskProgressProtocol.countOfBytesExpectedToReceive):
```
swiftlyfalling: It seems like this case is unnecessary for `NetworkUploadProcedure`? (It's not being registered for KVO.)

Sources/Network/NetworkUpload.swift
```diff
@@ -63,7 +72,31 @@ open class NetworkUploadProcedure<Session: URLSessionTaskFactory>: Procedure, Re
...
+                progress.totalUnitCount = task.countOfBytesExpectedToReceive
+            case #keyPath(URLSessionTaskProgressProtocol.countOfBytesReceived):
```
swiftlyfalling: It seems like this case is unnecessary for `NetworkUploadProcedure`? (It's not being registered for KVO.)

Sources/Network/NetworkData.swift
```diff
@@ -59,6 +72,32 @@ open class NetworkDataProcedure<Session: URLSessionTaskFactory>: Procedure, Resu
...
+            switch keyPath {
+            case #keyPath(URLSessionTaskProgressProtocol.countOfBytesExpectedToSend):
+                progress.totalUnitCount = task.countOfBytesExpectedToSend
+            case #keyPath(URLSessionTaskProgressProtocol.countOfBytesSent):
+                progress.completedUnitCount = task.countOfBytesSent
+            case #keyPath(URLSessionTaskProgressProtocol.countOfBytesExpectedToReceive):
+                progress.totalUnitCount = task.countOfBytesExpectedToReceive
+            case #keyPath(URLSessionTaskProgressProtocol.countOfBytesReceived):
+                progress.completedUnitCount = task.countOfBytesReceived
+            default: break
```
swiftlyfalling: Presuming that a data task eventually sets both `countOfBytesExpectedToSend` and `countOfBytesExpectedToReceive`, won't handling both cases as they are above cause progress to increase then regress and then increase again?

i.e.

1. `progress.completedUnitCount` is updated with task.countOfBytesSent until the request is fully sent

2. `progress.totalUnitCount` is **reset** to task.countOfBytesExpectedToReceive

3. `progress.completedUnitCount` is updated with the value of task.countOfBytesReceived, which starts at 0 and increases

Sources/Network/NetworkData.swift
```diff
@@ -59,6 +72,32 @@ open class NetworkDataProcedure<Session: URLSessionTaskFactory>: Procedure, Resu
...
+
+    override open func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?) {
+        guard context == &observerContext else {
+            super.observeValue(forKeyPath: keyPath, of: object, change: change, context: context)
+            return
+        }
+
+        if let keyPath = keyPath, let task = object as? Session.DataTask {
+            switch keyPath {
+            case #keyPath(URLSessionTaskProgressProtocol.countOfBytesExpectedToSend):
```
swiftlyfalling: Does this case get fired, since the observers are added after the task is created? (And, presumably, after the task has already set its `countOfBytesExpectedToSend`?)

swiftlyfalling:
Also, have you seen https://github.com/AFNetworking/AFNetworking/issues/3380 and https://github.com/AFNetworking/AFNetworking/pull/3607 ?

It seems as though there may be some edge cases with KVO that can result in crashes (or, at least, there were in AFNetworking).

Sources/Network/NetworkUpload.swift
blg-andreasbraun: Oh, copy and paste error

swiftlyfalling: Figured as much - happens to me all the time. 😊

Sources/Network/NetworkData.swift
blg-andreasbraun: Isn't that only the case when an upload tasks becomes a download task?
Is that even possible with the completion handler API?

swiftlyfalling: `countOfBytesReceived` is just "the number of bytes that the task has received from the server in the response body".

Can't any request potentially have a response?

Sources/Network/NetworkData.swift
blg-andreasbraun: I did some manual testing by uploading some files and they got fired.

But you are totally right, we actually know the content length at init time.

```
public init(session: Session, request: URLRequest? = nil, completionHandler: @escaping (HTTPResult<Data>) -> Void = { _ in }) {

        self.progress = Progress(totalUnitCount: Int64(request?.httpBody?.count ?? -1))
        ...
}
```

Sources/Network/NetworkData.swift
swiftlyfalling: According to the docs for [NSURLSessionDataTask](https://developer.apple.com/reference/foundation/nsurlsessiondatatask), it would seem as though a data task moves through the following flow:

- During upload of the body data (if your app provides any), the session periodically calls its delegate’s `URLSession:task:didSendBodyData:totalBytesSent:totalBytesExpectedToSend:` method with status information.
**NOTE: According to the docs for [countOfBytesSent](https://developer.apple.com/reference/foundation/urlsessiontask/1410444-countofbytessent), this should map to `countOfBytesSent` being incremented.**

- After receiving an initial response, the session calls its delegate’s `URLSession:dataTask:didReceiveResponse:completionHandler:` method to let you examine the status code and headers, and optionally convert the data task into a download task.
**NOTE: According to the docs for [countOfBytesExpectedToReceive](https://developer.apple.com/reference/foundation/urlsessiontask/1410663-countofbytesexpectedtoreceive), this should map to `countOfBytesExpectedToReceive` being updated to whatever the `Content-Length` header of the server's response indicates.**

- During the transfer [of the response], the session calls its delegate’s `URLSession:dataTask:didReceiveData:` method to provide your app with the content as it arrives.
**NOTE: According to the docs for [countOfBytesReceived](https://developer.apple.com/reference/foundation/urlsessiontask/1411581-countofbytesreceived), this should map to `countOfBytesReceived` being incremented.**

blg-andreasbraun:
@swiftlyfalling - Now that you mention it, I had such a crash in a project a while ago regarding to KVO.

I guess the delegation pattern is not really a possibility without having a private `URLSession` because a consumer could completely mess up the functionality by just setting the `delegate` of the provided `URLSession` while the procedure is running.

Sources/Network/NetworkData.swift
blg-andreasbraun: I guess this could be resolved by having to child progresses - one for the up- and one for the download.

danthorpe:
Hi, sorry, not sure what we're doing with this issue...

@blg-andreasbraun - would it be possible for you to update the branch and resolve the conflicts?

@swiftlyfalling - we've got concerns regarding KVO and thread-safety if I recall correctly, what are your thoughts on this?

PR #606, opened 2016-11-24
Async block procedures 

danthorpe:
Currently, `ResultProcedure`, `BlockProcedure` and `TransformProcedure` are only suitable for non-concurrent _work_. Previously, in _Operations_, `BlockOperation` supported concurrent _work_. This issue is to bring back that functionality, with some additional features.


danthorpe:
### `AsyncResultProcedure`

These procedures are initialised with a closure which receives a _finishing block_, this block in turn receives a `ProcedureResult<Output>` value. So, they should be used like this in the happy case:

```swift
let procedure = AsyncResultProcedure { finishWithResult in
    asyncTask { result, _ in
        finishWithResult(.success(result))
    }
}
```

and in the failure case:

```swift
let procedure = AsyncResultProcedure { finishWithResult in
    asyncTask { _, error in
        finishWithResult(.failure(error))
    }
}
```

### `AsyncBlockProcedure`

In the case of `AsyncBlockProcedure` the `Output` is `Void` - and _ProcedureKit_ provides a special argument for the successful finishing case:

```swift
let procedure = AsyncBlockProcedure { finishWithResult in
    asyncTask {
        finishWithResult(success)
    }
}
```

but the un-happy failure is the same:

```swift
let procedure = AsyncBlockProcedure { finishWithResult in
    asyncTask { error in
        finishWithResult(error.map { .failure($0) } ?? success)
    }
}
```

### `AsyncTransformProcedure`

In a similar vein, `AsyncTransformProcedure`, supports asynchronous transforms:

```swift
let procedure = AsyncTransformProcedure { input, finishWithResult in
    asyncTask(input) { output, error in
        finishWithResult(error.map { .failure($0) } ?? .success(output))
    }
}
```

leotumwattana:
I just tried AsyncBlockProcedure and ran into the same issue with my custom implementation earlier. `finishWithResult` is marks as `non-escaping` by default. Hence we'll get the compiler error `Closure use of non-escaping paramter 'finishWithResult' may allow it to escape`

danthorpe:
@leotumwattana - sorry - fixed now!

leotumwattana:
@danthorpe - No worries at all. Thanks!

PR #608, opened 2016-11-25
NetworkActivityController's delayed hide should use its internal queue

swiftlyfalling:


swiftlyfalling:
It also fixes a data race.

PR #611, opened 2016-11-25
Add missing imports to make Xcode happier

swiftlyfalling:
The projects compiled without issues, but Xcode seems to greatly prefer having explicit import statements at the top of each Swift file.

**Exception:**

We don't currently `import ProcedureKit` at the top of Swift files in the other pieces (ProcedureKitCloud, ProcedureKitMobile, etc) because of CocoaPods.


danthorpe:
@swiftlyfalling awesome 🙌

PR #613, opened 2016-11-26
NetworkObserver and NetworkActivityController cross-platform work

swiftlyfalling:
On all platforms, the following initializers are available:
```swift
NetworkObserver.init(controller: NetworkActivityController)
NetworkActivityController.init(timerInterval: TimeInterval = 1.0, indicator: NetworkActivityIndicatorProtocol)
```

To implement your own network activity indicator display, create a **NetworkActivityController** with an `indicator` that conforms to `NetworkActivityIndicatorProtocol`. In many cases, a shared indicator should be used. (i.e. Use the same indicator for all **NetworkObservers**.)

**NetworkActivityController** ensures that all accesses of the indicator’s `networkActivityIndicatorVisible` property are serialized on the main queue.

On **iOS**, the following convenience initializers are available that utilize **`UIApplication.shared`** to display/hide the network activity indicator in the status bar:
```swift
NetworkObserver.init()
NetworkActivityController.init(timerInterval: TimeInterval = 1.0)
static let NetworkActivityController.shared // the shared NetworkActivityController that uses UIApplication.shared
```
These convenience initializers are not available in Application Extensions.


swiftlyfalling:
@danthorpe: Remaining TODO: Move this to another part of the framework?

danthorpe:
@swiftlyfalling so - the issue with moving it out of _ProcedureKitMobile_ is that the other frameworks are Extension-API compatible, so it didn't compile. I think the `@available(iOSApplicationExtension, unavailable, message: " etc ")` does not work if the product is an extension API compatible. I totally got what you were doing - it's just the compiler didn't like it.

So I put all the iOS related stuff in extensions in ProcedureKitMobile, but left the main class in _ProcedureKit_.

swiftlyfalling:
Sounds good to me.

PR #582, opened 2016-11-27
Add "batch modify" support to CloudKitProcedure

danthorpe:
Some CloudKit operations work in batches, like modify records, or modify record zones. And it is possible for us to provide some rudimentary error handling when the number of items exceeds the limit.


danthorpe:
Gonna close this for now.

PR #636, opened 2016-12-05
OPR-595 NetworkProcedure Progress

blg-andreasbraun:


swiftlyfalling:
To address @danthorpe on the previous issue: https://github.com/ProcedureKit/ProcedureKit/pull/604#issuecomment-264862857

> Hi, sorry, not sure what we're doing with this issue...
> 
> @blg-andreasbraun - would it be possible for you to update the branch and resolve the conflicts?
> 
> @swiftlyfalling - we've got concerns regarding KVO and thread-safety if I recall correctly, what are your thoughts on this?

@danthorpe: AFNetworking ended up ripping out **all** of their KVO of tasks (including for monitoring progress and providing NSProgress) because of a number of issues that did not have easy (or readily apparent) solutions. Unfortunately, those issues were often crashes (with background sessions, suspending, and more).

It is not clear to me that this can be done using KVO without opening up a can of worms (again - for those unaware, ProcedureKit/Operations v3 used KVO on the task's state, and we had crashes because of it). I haven't dug into it in-depth, but the various AFNetworking issues are concerning: https://github.com/AFNetworking/AFNetworking/issues/3380 and https://github.com/AFNetworking/AFNetworking/pull/3607

danthorpe:
Thanks for the clarification @swiftlyfalling - I'll have a look at the AFNetworking issues tomorrow, and we can make a decision on the best way forward.

danthorpe:
I've just been reading through all the AFNetworking issues related to KVO of these properties. 

Most of the issues there were highlighted due to crashes in background tasks. Essentially deallocating tasks before an observing delegate could un-observe them. Although it was pointed out that this can also happen with a default session configuration. In _Operations_ v3, we also had crashes related to KVO of the session's state, which have proven quite tricky to pin down.

I think all things considered, **we won't merge these changes.**

If a framework consumer needs progress reporting on their network procedures, it is best to write a bespoke `Procedure` subclass. This has been the longstanding advice from me and other contributors. Wrapping _Alamofire_ in a `Procedure` is relatively straightforward, and this opens up the use of `URLSessionDelegate` for reliable progress reporting.

Having said that, what is the point of `NetworkDataProcedure` et. al.? Well, these are still very useful procedures for prototyping/background/quick networking tasks. They also provide a minimal requirement for `NetworkProcedure` which can support custom networking procedures.

While I'm always happy to continue discussions and furthering the framework, unless anyone has any genius ideas on how to provide network progress without re-creating _Alamofire_, access to `URLSessionDelegate` or using KVO - then I'd like to move on.

blg-andreasbraun:
I guess we have to wait for Apple to add support for this out of the box.
I submitted a feature request, maybe someday we'll get it :)

PR #639, opened 2016-12-07
ProcessProcedure improvements & thread-safety

swiftlyfalling:
`NSTask`/`Process` is listed as a Thread-Unsafe Class in the [Foundation Framework Thread Safety guide](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html#//apple_ref/doc/uid/10000057i-CH12-125664).

Additionally, when an NSTask/Process is `launch()`ed on a thread that goes away before it does, [a crash can result](https://web.archive.org/web/20111213185028/http://cocoadev.com/index.pl?NSTaskTermination).

As such, some major modifications to ProcessProcedure are necessary.

- `init(…)` now takes all the possible configuration parameters for the Process, instead of a Process instance (the only required parameter is `launchPath`)
- the ProcessProcedure is now responsible for creating the Process object and managing it
- there is no public accessor for the Process object (it’s now an internal implementation detail of ProcessProcedure)
- various additional methods / properties were added to expose Process object functionality directly on the ProcessProcedure

Internally, the ProcessProcedure ensures that `launch()`, `terminate()`, and other important methods of the Process are dispatched to the main queue/thread.

### DIFF
--- a/ProcessProcedureTests.swift
+++ b/ProcessProcedureTests.swift
@@ -46,10 +46,9 @@ class ProcessProcedureTests: ProcedureKitTestCase {
 
     func test__cancel_process_after_launched() {
         processProcedure = ProcessProcedure(launchPath: "/bin/bash/", arguments: ["-c", "sleep 2"])
-        processProcedure.addDidExecuteBlockObserver { (procedure) in
+        check(procedure: processProcedure, withTimeout: 1) { procedure in
             procedure.cancel()
         }
-        wait(for: processProcedure, withTimeout: 1)
         XCTAssertProcedureCancelledWithoutErrors(processProcedure)
     }
 
@@ -60,12 +59,10 @@ class ProcessProcedureTests: ProcedureKitTestCase {
     func test__processIdentifier_after_launched() {
         processProcedure = ProcessProcedure(launchPath: "/bin/bash/", arguments: ["-c", "sleep 1"])
         weak var didExecuteExpectation = expectation(description: "DidExecute: \(#function)")
-        var processIdentifier = Protector<Int32>(-1)
-        processProcedure.addDidExecuteBlockObserver { (procedure) in
+        let processIdentifier = Protector<Int32>(-1)
+        processProcedure.addDidExecuteBlockObserver { procedure in
             let retrievedProcessIdentifier = procedure.processIdentifier
-            processIdentifier.write({ (processIdentifier) in
-                processIdentifier = retrievedProcessIdentifier
-            })
+            processIdentifier.overwrite(with: retrievedProcessIdentifier)
             DispatchQueue.main.async {
                 didExecuteExpectation?.fulfill()
             }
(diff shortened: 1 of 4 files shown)


Sources/Mac/Process.swift
```diff
@@ -38,47 +33,151 @@ open class ProcessProcedure: Procedure {
...
+    /// - Parameters:
+    ///   - launchPath: the path to the executable to be launched.
+    ///   - arguments: (optional) the command arguments that should be used to launch the executable.
+    ///   - currentDirectoryPath: (optional) the current directory to be used when launching the executable.
+    ///   - environment: (optional) the environment to be used when launching the executable.
+    ///   - standardError: (optional) the standard error (FileHandle or Pipe object)
+    ///   - standardInput: (optional) the standard input (FileHandle or Pipe object)
+    ///   - standardOutput: (optional) the standard output (FileHandle or Pipe object)
+    ///   - processDidExitCleanly: a ProcessDidExitCleanly closure with a default.
+    public init(launchPath: String, arguments: [String]? = nil, currentDirectoryPath: String? = nil, environment: [String : String]? = nil, standardError: Any? = nil, standardInput: Any? = nil, standardOutput: Any? = nil, processDidExitCleanly: @escaping ProcessDidExitCleanly = ProcessProcedure.defaultProcessDidExitCleanly) {
```
danthorpe: This is so much better! 👍

ProcessProcedureTests.swift
```diff
@@ -8,24 +8,224 @@
...
+        wait(for: processProcedure)
+        XCTAssertProcedureCancelledWithoutErrors(processProcedure)
+    }
+
+    func test__cancel_process_after_launched() {
+        processProcedure = ProcessProcedure(launchPath: "/bin/bash/", arguments: ["-c", "sleep 2"])
+        processProcedure.addDidExecuteBlockObserver { (procedure) in
+            procedure.cancel()
+        }
+        wait(for: processProcedure, withTimeout: 1)
```
danthorpe: I guess, this is okay, but I think a better way of testing this is to use the `check` API:

```swift
processProcedure = ProcessProcedure(launchPath: "/bin/bash/", arguments: ["-c", "sleep 2"])
check(procedure: processProcedure, withTimeout: 1) { procedure in
    procedure.cancel()
}
XCTAssertProcedureCancelledWithoutErrors(processProcedure)
```

Because 
a. it's more consistent throughout the test suite
b. it has less reliance on _ProcedureKit_ machinery.

swiftlyfalling: For future reference, per Slack discussion:

- `check` does not enforce timing between the execution of the block and the execution of the Procedure
- Thus, replacing the didExecuteObserver in this test with `check` can result in the cancellation sometimes occurring *prior* to the ProcessProcedure executing.
- Which won't result in test failures, but doesn't end up testing what we want in all cases.

PR #643 fixes this by adding a new `checkAfterDidExecute` method, and using it in this test.

PR #662, opened 2016-12-22
TimeoutObserver improvements

swiftlyfalling:
- Utilize a new ProcedureTimeoutRegistrar to handle the lifetime of timeout timers.
- When a Procedure finishes (or a timer fires), cancel all outstanding timers for that Procedure. (Thus, a TimeoutObserver with a long timeout will immediately deallocate its timers and associated timeout blocks when the Procedure finishes, instead of keeping the timers / timeout blocks in memory until the timeout passes.)
- Eliminate impact on the main thread/queue by executing the timer blocks on a private queue.

### DIFF
--- a/Tests/TimeoutObserverTests.swift
+++ b/Tests/TimeoutObserverTests.swift
@@ -44,4 +44,28 @@ final class TimeoutObserverTests: ProcedureKitTestCase {
         wait(for: procedure)
         XCTAssertProcedureFinishedWithoutErrors()
     }
+
+    func test__multiple_timeout_observers_on_a_single_procedure() {
+        procedure = TestProcedure(delay: 0.5)
+        procedure.add(observer: TimeoutObserver(by: 0.1))
+        procedure.add(observer: TimeoutObserver(by: 0.1))
+        procedure.add(observer: TimeoutObserver(by: 0.2))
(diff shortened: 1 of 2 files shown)


danthorpe:
@swiftlyfalling should we add some more tests here. Just wondering what happens if you add a more than one `TimeoutObserver` to a procedure.

swiftlyfalling:
@danthorpe: Yup, that's a great idea. Will add some tests for that.

Tests/TimeoutObserverTests.swift
```diff
@@ -44,4 +44,28 @@ final class TimeoutObserverTests: ProcedureKitTestCase {
...
+        procedure = TestProcedure(delay: 0.5)
+        procedure.add(observer: TimeoutObserver(by: 0.1))
+        procedure.add(observer: TimeoutObserver(by: 0.1))
+        procedure.add(observer: TimeoutObserver(by: 0.2))
+        procedure.add(observer: TimeoutObserver(by: 3.0))
+        wait(for: procedure)
+        XCTAssertProcedureCancelledWithErrors(count: 1)
+    }
+
+    func test__add_single_timeout_observer_to_multiple_procedures() {
```
danthorpe: Oh, interesting - but yes, I guess that is a perfectly reasonable use case 👍

PR #668, opened 2017-01-09
Procedure EventQueue

swiftlyfalling:
The core of Procedure now utilizes an internal EventQueue - a serial FIFO queue onto which all user “events” are dispatched - which results in major thread-safety, performance, and ease-of-use wins, and a reduction in framework overhead.

Numerous additional tests are included, edge-case bugs are fixed, and code is simplified.

### What is the EventQueue?

The EventQueue is an internal queue that each Procedure uses to ensure that all user events are executed in a serial (non-concurrent) fashion, to:
- drastically simplify both internal code complexity and framework consumer code complexity
- provide more safety and functionality "by default" (i.e. built-in to the framework)

### What are "Procedure events"?

"Procedure events" include anything that calls user code, such as:
- overridden methods like: 
    - Procedure.`execute()`
    - Procedure.`procedureDidCancel(withErrors:)`
    - Procedure.`procedureWillFinish(withErrors: [Error])`
    - Procedure.`procedureDidFinish(withErrors: [Error])`
    - GroupProcedure.`groupWillAdd(child:)` **[NEW]**
    - GroupProcedure.`child(_:willAttemptRecoveryFromErrors:)`
    - GroupProcedure.`childWillFinishWithoutErrors(_:)`
- all observer callbacks
- injecting results from a dependency

The framework now handles this automatically in the background, **improving the thread-safety of your custom Procedures without any extra work on your part**, and dramatically reducing the number of scenarios in which locks are required for full thread-safety.

> For one simple example of the benefits, take a look at the changes to **`Delay.swift`**, and **`NetworkDownloadProcedure.swift`**. A lock and a bunch of code/complexity was made redundant by these changes, and was removed.

Additionally, ProcedureKit now provides helper functionality that makes it easy to synchronize observers on one Procedure, with another Procedure's EventQueue or even your own custom DispatchQueue that you use to process callbacks.

And all necessary methods of ProcedureKit are now (internally) asynchronous. It is safe to call any Procedure method from a block on *any* Procedure’s EventQueue / other DispatchQueue / thread, and most edge cases in which deadlock can be triggered by user code calling ProcedureKit methods have been *eliminated entirely*.

### Performance Improvement:

These changes result in a **~15%+ overhead reduction / performance improvement**
(measured using the Stress Tests running in a Release mode build).

### Breaking Changes (from prior Procedure v4 betas):

**TL/DR:** Almost all code built using the prior beta of ProcedureKit v4 will work with these new improvements while requiring only minimal changes.

The few areas of change that may impact existing code are:

#### [Breaking Changes]

- Observer changes:
    - **WillCancel** observers have been **removed**
        - **Solution:** use a DidCancel observer instead to react to cancellation
    - **WillExecute** and **WillFinish** observers **now receive an additional parameter** - a `PendingExecuteEvent` / `PendingFinishEvent` - which can be used to ensure that asynchronous tasks occur before the Procedure executes / finishes.
        ```diff
        public protocol ProcedureObserver {
        - func will(execute procedure: Procedure)
        + func will(execute procedure: Procedure, pendingExecute: PendingExecuteEvent)
        - func will(finish procedure: Procedure, withErrors errors: [Error]) 
        + func will(finish procedure: Procedure, withErrors errors: [Error], pendingFinish: PendingFinishEvent)
        }
        ```
        This also impacts the respective BlockObservers, which receive the pendingEvent as an additional parameter.
        - **Solution:** Update any custom Observer classes to use the new ProcedureObserver methods, and update any WillExecute / WillFinish BlockObserver blocks to expect the additional parameter.
- `OperationQueueDelegate` (a hold-over from Operations 3.x) is fully replaced by a revamped `ProcedureQueueDelegate`.
    > `ProcedureQueueDelegate` optionally offers the new ability to process `willAddOperation` / `willAddProcedure` delegate callbacks _asynchronously_ (by returning a `ProcedureFuture`), in addition to some new convenient functionality for linking a context passed-in to the ProcedureQueue.add(operation:withContext:) method to delegate Will/DidAdd callbacks.
    - **Solution:** If you utilize your own queue delegate, please check the new `ProcedureQueueDelegate` for delegate callback changes.

#### [Infrequently-Breaking Changes]

- `Procedure.produce()` is now asynchronous.
    If you must do something *after* it succeeds (i.e. after the produced operation is added to the queue), you can use the `ProcedureFuture` it returns to schedule a block afterwards:

    ```swift
    procedure.produce(operation: otherOperation).then(on: DispatchQueue.main) {
        // do something on the main queue after otherOperation
        // is scheduled on the ProcedureQueue
    }
    ```
    Alternatively, `produce()` now takes an optional `PendingEvent`, and will ensure that the operation is produced before the `PendingEvent` is ready to proceed. For example, to produce an operation before another operation finishes (from its WillFinish observer):

    ```swift
    procedure.addWillFinishObserver { procedure, errors, pendingFinish in
        // will produce an operation before `procedure` finishes
        procedure.produce(operation: otherOperation, before: pendingFinish)
    }
    ```

    In most cases, in which order/timing is not important (or is already enforced via dependencies), **no change will be required**.

### New API:

```diff
open class GroupProcedure {
+ /**
+   This method is called when a child will be added to the Group.
+   (It is called on the Group's EventQueue.)
+  */
+ open func groupWillAdd(child: Operation) { /* no-op */ }
}
```

#### Observers:

```diff
public protocol ProcedureObserver {
+ /**
+   Provide a queue onto which observer callbacks will be dispatched.
+  */
+ var eventQueue: DispatchQueueProtocol? { get }
}

public struct BlockObserver {
+ /// - returns: the queue onto which observer callbacks are dispatched
+ public let eventQueue: DispatchQueueProtocol?
...
+ /// - parameter synchronizedWith: a provider of the queue onto which observer callbacks are dispatched
+ public init(
+         synchronizedWith queueProvider: QueueProvider? = nil,
...
}
```

Additionally, `WillExecuteObserver`, `DidExecuteObserver`, `DidCancelObserver`, `WillAddOperationObserver`, `DidAddOperationObserver`, `WillFinishObserver`, `DidFinishObserver` now all take an optional `synchronizedWith:` parameter in their init, to which you can pass a `QueueProvider` (i.e. a `Procedure`, `EventQueue`, or `DispatchQueue`) that provides a queue onto which all observer callbacks are synchronously dispatched.

#### New Block Procedures:

Easily handle cancellation while running from within the block:

```diff
+ open class CancellableResultProcedure<Output>: Procedure, OutputProcedure { 
+ 
+    /// A block that receives a closure (that returns the current value of `isCancelled`
+    /// for the CancellableResultProcedure), and returns a value (which is set as the
+    /// CancellableResultProcedure's `output`).
+    public typealias ThrowingCancellableOutputBlock = (() -> Bool) throws -> Output
...
+    public init(cancellableBlock: @escaping ThrowingCancellableOutputBlock)
...
+ }
+ open class CancellableBlockProcedure: CancellableResultProcedure<Void> { }
```

Example usage:

```swift
let block = CancellableBlockProcedure { isCancelled in
    // process something in a loop while not cancelled:
    while !isCancelled() {
        // do something
    }
}
```


Sources/ProcedureKit/Collection+ProcedureKit.swift
```diff
@@ -9,16 +9,18 @@ import Foundation
 extension Collection where Iterator.Element: Operation {
 
     internal var operationsAndProcedures: ([Operation], [Procedure]) {
-        return reduce(([], [])) { result, element in
-            var (operations, procedures) = result
+        // this for-in loop has considerably better performance characteristics than reduce()
```
jshier: 👍 

If anyone's curious, see https://airspeedvelocity.net/2015/08/03/arrays-linked-lists-and-performance/ for more info.

Sources/ProcedureKit/Delay.swift
```diff
@@ -74,20 +74,15 @@ public class DelayProcedure: Procedure {
...
     private var _timer: DispatchSourceTimer? = nil
-    private let stateLock = NSLock()
 
     internal init(delay: Delay, leeway: DispatchTimeInterval = .milliseconds(1)) {
         self.delay = delay
         self.leeway = leeway
         super.init()
         name = "Delay \(delay)"
         addDidCancelBlockObserver { procedure, _ in
-            procedure.stateLock.withCriticalScope {
```
jshier: 👏 👏

Sources/ProcedureKit/DispatchQueue+ProcedureKit.swift
```diff
@@ -45,6 +45,10 @@ public extension DispatchQueue {
         }
         return try work()
     }
+
+    static var currentQoSClass: DispatchQoS.QoSClass {
+        return DispatchQoS.QoSClass(rawValue: qos_class_self()) ?? DispatchQoS.QoSClass.unspecified
```
jshier: You can likely shorten the right side to just `.unspecified`.

Sources/ProcedureKit/Group.swift
```diff
@@ -37,20 +37,53 @@ open class GroupProcedure: Procedure, ProcedureQueueDelegate {
...
+     The default service level to apply to the GroupProcedure and its child operations.
+     
+     This property specifies the service level applied to the GroupProcedure itself, and to
+     operation objects added to the GroupProcedure.
+     
+     If the added operation object has an explicit service level set, that value is used instead.
+     
+     For more, see the NSOperation and NSOperationQueue documentation for `qualityOfService`.
+     */
+    @available(OSX 10.10, iOS 8.0, tvOS 8.0, watchOS 2.0, *)
```
jshier: There's no such thing as `tvOS 8.0`. Guess the compiler doesn't validate that.

swiftlyfalling: Good catch - that code was previously elsewhere in the file (was moved because it's `open` to resolve a visibility scope warning), but this should indeed be fixed.

Sources/ProcedureKit/Delay.swift
swiftlyfalling: Yup! 🎉

Sources/ProcedureKit/PendingEvent.swift
```diff
@@ -0,0 +1,124 @@
...
+///
+/// ```swift
+/// procedure.addWillFinishObserver { procedure, errors, futureFinish in
+///     // produce a new operation before the procedure finishes
+///     procedure.produce(BlockOperation { /* do something */ }, before: futureFinish)
+/// }
+/// ```
+///
+final public class PendingEvent: CustomStringConvertible {
+    public enum PendingEventType: CustomStringConvertible {
```
jshier: Perhaps this can be shortened to just `Type`, so it can be referenced as `PendingEvent.Type` rather than `PendingEvent.PendingEventType`.

swiftlyfalling: `Type` is a bit annoying to use, per the Swift compiler:
```swift
Type member may not be named 'Type', since it would conflict with the 'foo.Type' expression
```

But perhaps just shorten it to `Event` instead?

jshier: Oh yeah. I think `EventType` works there, as the class is the event, the enum is just the type.

Sources/ProcedureKit/Procedure.swift
```diff
@@ -100,44 +122,69 @@ open class Procedure: Operation, ProcedureProtocol {
...
-    private let _stateLock = NSRecursiveLock()
+    // the state variable to be used *within* the stateLock
+    private var _state = ProcedureKit.State.initialized {
+        willSet(newState) {
+            _log.verbose(message: "\(_state) -> \(newState)")
+            assert(_state.canTransition(to: newState, whenCancelled: _isCancelled), "Attempting to perform illegal cyclic state transition, \(_state) -> \(newState) for operation: \(identity). Ensure that Procedure instances are added to a ProcedureQueue not an OperationQueue.")
+        }
+    }
+
+    fileprivate let _stateLock = PThreadMutex()
```
jshier: Would it be possible to expose a `stateLock` to subclasses? Perhaps not this one, but a more public one that developers can use for their own asynchronous `Procedure`'s state? e.g. In my Alamofire-wrapping `NetworkProcedure`, it seems like my `Request` value should be locked, as it's set during execution and has methods called during cancellation. Instead of having to keep my own `stateLock` value, perhaps it would be beneficial to have one already available.

Sources/ProcedureKit/Procedure.swift
```diff
@@ -212,27 +286,28 @@ open class Procedure: Operation, ProcedureProtocol {
     final public var log: LoggerProtocol {
         get {
             let operationName = self.operationName
-            return protectedProperties.read { LoggerContext(parent: $0.log, operationName: operationName) }
+            return _stateLock.withCriticalScope { LoggerContext(parent: protectedProperties!.log, operationName: operationName) }
```
jshier: Perhaps make `protectedProperties` an IUO, so you can avoid the `!` everywhere and still be able to set it to `nil`?

Sources/ProcedureKit/Procedure.swift
swiftlyfalling: Exposing the Procedure's internal _stateLock is very, very problematic (very easy to introduce deadlock if people don't know what they're doing), but this touches the part I haven't written up yet for the documentation:

You can use the Procedure's EventQueue. (See `Procedure.eventQueue`). You can simply dispatch blocks onto the EventQueue to do things that manipulate state.

Sources/ProcedureKit/Procedure.swift
```diff
@@ -478,46 +670,90 @@ open class Procedure: Operation, ProcedureProtocol {
...
     }
 
-    private var shouldCancel: Bool {
+    // used by GroupProcedure to bypass dispatching to the event queue for its cancellation handling
+    internal func _procedureDidCancel() {
+        // no-op
+    }
+
+    private enum ShouldCancelResult {
+        case ShouldCancel
```
jshier: Cases should be lower-cased to match Swift 3 style, no?

Sources/ProcedureKit/Procedure.swift
```diff
@@ -531,73 +767,171 @@ open class Procedure: Operation, ProcedureProtocol {
...
-        _finish(withErrors: errors, from: .finish)
+    public func finish(withErrors errors: [Error] = []) {
+        log.verbose(message: "finish() called")
+        finish(withErrors: errors, from: .finish)
+    }
+
+    // Used to queue an automatic finish from Procedure.start()/main()
+    // (i.e. if a Procedure should automatically finish prior to executing if, for example,
+    // it is cancelled prior to executing)
+    private func queueAutomaticFinish(from source: ProcedureKit.FinishingFrom) {
```
jshier: Does `FinishingFrom` need to be fully qualified like this?

Sources/ProcedureKit/Procedure.swift
swiftlyfalling: Yup. Good catch!

Sources/ProcedureKit/Procedure.swift
```diff
@@ -531,73 +767,171 @@ open class Procedure: Operation, ProcedureProtocol {
...
-            return true
+    private final func _finish(withInfo info: FinishingInfo) {
+        dispatchEvent {
+            self._finishImpl(withInfo: info)
         }
     }
 
-    private final func _finish(withErrors receivedErrors: [Error], from source: ProcedureKit.FinishingFrom) {
-        guard shouldFinish(from: source) else { return }
+    private final func _finishImpl(withInfo info: FinishingInfo) {
```
jshier: Perhaps the naming here could be a bit more descriptive so it's easier to see why both `_finish` and `_finishImpl` exist. Also `Impl` gives me C++ flashbacks. ;)

Sources/ProcedureKit/PendingEvent.swift
```diff
@@ -0,0 +1,124 @@
...
+        procedure.log.verbose(message: "(\(self)) is ready to proceed")
+    }
+
+    public var description: String {
+        return "Pending\(event.description) for: \(procedure.procedureName)"
+    }
+}
+
+internal extension PendingEvent {
+    static let executeEvent: (Procedure) -> PendingEvent = { PendingEvent(forProcedure: $0, withEvent: .execute) }
```
jshier: Seems cleaner to drop the `Event` from all of these.

Sources/ProcedureKit/Procedure.swift
swiftlyfalling: Yep, because it's actually part of an internal struct `ProcedureKit` earlier in the file. (Has been organized like that for quite a while, I believe.)

Sources/ProcedureKit/ProcedureEventQueue.swift
```diff
@@ -0,0 +1,248 @@
...
+}
+
+extension Procedure: QueueProvider {
+    public var providedQueue: DispatchQueueProtocol { return eventQueue }
+}
+
+// MARK: - DispatchQueueProtocol
+
+public protocol DispatchQueueProtocol: class {
+    @discardableResult func asyncDispatch(block: @escaping () -> Void) -> DispatchWorkItem
```
jshier: I think it would be a bit cleaner to drop the `Dispatch` from these functions, if possible. That might cause collisions with the `DispatchQueue` implementation but I don't think so. Just makes it a bit easier to read, for the same reasons they renamed the `DispatchQueue` functions in Swift 3.

Sources/ProcedureKit/ProcedureFuture.swift
```diff
@@ -0,0 +1,336 @@
...
+ */
+public class ProcedurePromise {
+    public let future = ProcedureFuture()
+    #if DEBUG
+    private var _didComplete = false
+    private var lock = PThreadMutex()
+    #endif
+
+    #if DEBUG
+    // in Debug builds, verify and assert if a promise is not completed before it is deinited
```
jshier: Just to make sure you know, `assert`s are only compiled in when using `-O0`. I know you're also using `DEBUG` to add properties in such, just wanted to make sure you know.

Sources/ProcedureKit/ProcedureFuture.swift
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    fileprivate init() {
+        group.enter()
+    }
+
+    public func then(on eventQueueProvider: QueueProvider, block: @escaping (ProcedureResult<Result>) -> Void) {
+
+        let eventQueue = eventQueueProvider.providedQueue
+        eventQueue.dispatchNotify(withGroup: group, block: {
```
jshier: Looks like you lost the trailing closure syntax here.

Sources/ProcedureKit/ProcedureFuture.swift
```diff
@@ -0,0 +1,334 @@
...
+
+    /// Schedules a block for execution after the ProcedureFuture completes.
+    ///
+    /// - Parameters:
+    ///   - on: a queue [DispatchQueue, EventQueue, Procedure (which schedules it on its internal event queue)]
+    ///   - block: The block to be scheduled after the ProcedureFuture completes.
+    public func then(on eventQueueProvider: QueueProvider, block: @escaping (Void) -> Void) {
+
+        let eventQueue = eventQueueProvider.providedQueue
+        eventQueue.dispatchNotify(withGroup: group, block: {
```
jshier: Looks like these calls are losing trailing closures a lot.

Sources/ProcedureKit/ProcedureEventQueue.swift
swiftlyfalling: The concern was indeed naming collisions. Will look into this a bit more. This protocol is mostly built for internal use.

Sources/ProcedureKit/ProcedureFuture.swift
```diff
@@ -0,0 +1,334 @@
...
+
+// Used by ProcedureFutureGroup
+fileprivate extension ProcedureFuture {
+    // If the future is immediately available, execute the block synchronously on the current thread
+    // otherwise, queue a future dispatch onto the QueueProvider's queue.
+    @discardableResult
+    func thenOnSelfOrLater(on eventQueueProvider: QueueProvider, block: @escaping (Void) -> Void) -> ProcedureFuture {
+        let promise = ProcedurePromise()
+        let eventQueue = eventQueueProvider.providedQueue
+        if group.wait(timeout: .now()) == DispatchTimeoutResult.success {
```
jshier: Shouldn't need to fully qualify the `DispatchTimeoutResult` here. 

(Let me know if pointing out these naming things is annoying.)

swiftlyfalling: (Not annoying at all - much appreciated! Although was already about to check in several that you mentioned. 😁)

Sources/ProcedureKit/ProcedureFuture.swift
```diff
@@ -0,0 +1,334 @@
...
+        }
+        return promise.future
+    }
+}
+
+/// Easily wait on multiple ProcedureFutures.
+/// Provides a future that is signaled once all ProcedureFutures in an array are signaled.
+///
+/// ```swift
+/// let futures: [ProcedureFuture] = ... // multiple ProcedureFutures
```
jshier: Perhaps an extension on `Collection` or some such may make this even more convenient?

Sources/ProcedureKit/ProcedureQueue.swift
```diff
@@ -6,103 +6,148 @@
...
+ Delegate callbacks may occur concurrently and on any thread.
+ 
+ Examples:
+    - It is possible for `didAddOperation` / `didAddProcedure` to be called concurrently with
+    `didFinishOperation` / `will/didFinishProcedure` for an Operation / Procedure instance.
+    - It is possible for `didAddOperation` to be called concurrently for two different Operation
+    instances added to a ProcedureQueue simultaneously. (i.e. Once for each Operation.)
+ 
+ */
+public protocol ProcedureQueueDelegate: class {
```
jshier: There's been talk of removing `class` as an inheritable type for protocols. I usually use `AnyObject` to convey this requirement.

Sources/ProcedureKit/ProcedureQueue.swift
```diff
@@ -137,129 +184,273 @@ open class ProcedureQueue: OperationQueue {
...
+
+            // if no delegate, proceed to Step 2 directly:
+            _add_step2(procedure: procedure, context: context, promise: promise)
+        }
+    }
+
+    // Step2 2: process conditions
+    //          add the evaluator procedure to self (using async method), .then(onQueue: )
+    //
+    private func _add_step2(procedure: Procedure, context: Any?, promise: ProcedurePromise) {
```
jshier: I know these functions are private, but the naming is fairly odd. I'm not sure if it would be better or worse to name these step functions after what they do rather than just ordering. The number makes the sequence easy to reason about but hides all info about what the function actually does. Plus the `_` after `add` isn't Swift style.

Sources/ProcedureKit/ProcedureQueue.swift
```diff
@@ -137,129 +184,273 @@ open class ProcedureQueue: OperationQueue {
...
 
-    /**
-     Adds the operations to the queue.
+            // Filter out the indirect dependencies which have already been added to the queue
+            let indirectDependenciesToProcess = indirectDependencies.filter { !self.operations.contains($0) }
 
-     - parameter ops: an array of `NSOperation` instances.
-     - parameter wait: a Bool flag which is ignored.
+            // Check to see if there are any which need processing
+            if indirectDependenciesToProcess.count > 0 {
```
jshier: A `guard` here lets you drop a level of indentation.

Sources/ProcedureKit/ProcedureQueue.swift
```diff
@@ -137,129 +184,273 @@ open class ProcedureQueue: OperationQueue {
...
+     - parameters operations: a variadic array of `NSOperation` instances.
+     - parameter context: an optional parameter that is passed-through to the Will/DidAdd delegate callbacks
+     - returns: a ProcedureFuture that is signaled once the operation has been added to the ProcedureQueue
      */
-    open override func addOperations(_ ops: [Operation], waitUntilFinished wait: Bool) {
-        ops.forEach { addOperation($0) }
+    @discardableResult
+    final func add<S: Sequence>(operations: S, withContext context: Any? = nil) -> ProcedureFuture where S.Iterator.Element: Operation {
+
+        let futures = Array(operations).map {
```
jshier: No need to create an array first if all you're doing is a map, `Sequence` has a `map` that returns an `Array`.

Sources/ProcedureKit/ProcedureFuture.swift
swiftlyfalling: Yup. Great suggestion.

Sources/ProcedureKit/ProcedureQueue.swift
swiftlyfalling: Yup. Carry-over from when this wasn't broken out into its own function (i.e. prior commits).

Tests/ProcedureKitTests/KVONotificationTests.swift
```diff
@@ -0,0 +1,378 @@
...
+class KVOTests: ProcedureKitTestCase {
+
+    class NSOperationKVOObserver: NSObject {
+
+        let operation: Operation
+        private var removedObserved = false
+        private var isFinishedBlock: (() -> Void)?
+
+        enum KeyPath: String {
+            case Cancelled = "isCancelled"
```
jshier: I know it's only a testing class, but these should be lower-cased as well.

Tests/ProcedureKitTests/KVONotificationTests.swift
```diff
@@ -0,0 +1,378 @@
...
+            case Executing = "isExecuting"
+            case Finished = "isFinished"
+            case Ready = "isReady"
+            case Dependencies = "dependencies"
+            case QueuePriority = "queuePriority"
+            case CompletionBlock = "completionBlock"
+        }
+
+        struct KeyPathSets {
+            static let State = Set<String>([KeyPath.Cancelled.rawValue, KeyPath.Executing.rawValue, KeyPath.Finished.rawValue, KeyPath.Ready.rawValue])
```
jshier: Should this be `states` instead?

Tests/ProcedureKitTests/KVONotificationTests.swift
```diff
@@ -0,0 +1,378 @@
...
+//  ProcedureKit
+//
+//  Copyright © 2016 ProcedureKit. All rights reserved.
+//
+
+import XCTest
+import TestingProcedureKit
+@testable import ProcedureKit
+
+class KVOTests: ProcedureKitTestCase {
```
jshier: It's great to see this huge source of uncertainty finally locked down with some tests.

Tests/ProcedureKitTests/KVONotificationTests.swift
```diff
@@ -0,0 +1,378 @@
...
+        }
+        let procedure = NoFinishOperation()
+        weak var expDidExecute = expectation(description: "")
+        procedure.addDidExecuteBlockObserver { _ in
+            DispatchQueue.main.async {
+                expDidExecute?.fulfill()
+            }
+        }
+        let kvoObserver = NSOperationKVOObserver(operation: procedure)
+        procedure.willEnqueue(on: queue) // trigger state transition from .Initializing -> .WillEnqueue
```
jshier: May want to audit comments for proper case naming (capitalization).

jshier:
This is really really great @swiftlyfalling! If I get some time later this week I'll try updating my project to use this version and see how it goes. I've gone through the changes and made some naming notes, but I may have more if I get a chance to update my code. 

I really hope that no one uses ProcedureKit just for Promises / Futures, but they make an interesting addition the library.

Sources/ProcedureKit/ProcedureQueue.swift
swiftlyfalling: Until any Swift language change on this is finalized, I think it's reasonable to match the current Swift documentation and the pattern used elsewhere in ProcedureKit (which is `class`).

(I do see the argument for changing the language, though.)

Sources/ProcedureKit/ProcedureQueue.swift
danthorpe: Yeah, I agree with @jshier here. I think things like "step 2" are ill-defined, and will be hard to read & maintain in 3 months time.

Not sure whether renaming is something we can do a bit later though, as I think we can tackle style/naming later on, after merging, to avoid as much risk.

Sources/ProcedureKit/Support.swift
```diff
@@ -113,6 +113,107 @@ public extension Protector where T: RangeReplaceableCollection {
...
+    /// - Returns: returns the return value of the block
+    public func withCriticalScope<T>(block: () -> T) -> T {
+        lock()
+        defer { unlock() }
+        let value = block()
+        return value
+    }
+}
+
+public class MutexProtector<T> {
```
danthorpe: Is this used instead of `Protector`? We could just name this `Protector` and remove the previous version?

Sources/TestingProcedureKit/XCTAsserts.swift
```diff
@@ -268,3 +268,19 @@ public extension ProcedureKitTestCase {
...
+    func XCTAssertProcedureNoConcurrentEvents<T: ProcedureProtocol & EventConcurrencyTrackingProcedureProtocol>(_ exp: @autoclosure () throws -> T, minimumConcurrentDetected: Int = 1, _ message: @autoclosure () -> String = "", file: StaticString = #file, line: UInt = #line) where T: Procedure {
+        __XCTEvaluateAssertion(testCase: self, message, file: file, line: line) {
+            let procedure = try exp()
+            let detectedConcurrentEvents = procedure.concurrencyRegistrar.detectedConcurrentEvents
+            guard procedure.concurrencyRegistrar.maximumDetected >= minimumConcurrentDetected && detectedConcurrentEvents.isEmpty else {
+                return .expectedFailure("\(procedure.procedureName) detected concurrent events: \n\(detectedConcurrentEvents)")
+            }
+            return .success
+        }
+    }
```
danthorpe: Very nice 😁

Sources/ProcedureKit/Support.swift
swiftlyfalling: Good suggestion - had a chance to audit the rest of the uses. Done!

Sources/ProcedureKit/Procedure.swift
```diff
@@ -151,25 +198,52 @@ open class Procedure: Operation, ProcedureProtocol {
...
+    // Grouped in a class to allow for easily deinitializing in `deinit`.
+    fileprivate class ProtectedProperties {
         var log: LoggerProtocol = Logger()
         var errors = [Error]()
         var observers = [AnyObserver<Procedure>]()
         var directDependencies = Set<Operation>()
         var conditions = Set<Condition>()
     }
-    fileprivate let protectedProperties = Protector(ProtectedProperties())
+    fileprivate var protectedProperties: ProtectedProperties! = ProtectedProperties()
```
swiftlyfalling: To permit easy (and explicit) deinit in `Procedure.deinit` (inside the lock, to resolve a data race).

jshier:
So I upgraded my latest project to this branch today and it went pretty smoothly. Aside from some additional closure parameters there wasn't anything breaking. Also, I really like the `synchronizedWith` API, as it makes it easy to guarantee my observers are running on a given queue. Very nice. 

However, running my project through the Leaks instrument brought up fairly extensive leaks of the `_NativeSetStorageImpl/Owner<Condition>`, so it seems like either the procedures are leaking the conditions set or there's a false positive. Also, possible leaks of the array of `Operation`s in `canFinishGroup`, as well as one `PThreadMutex` (perhaps just the one in my persistent queue).

jshier:
Nothing in the Memory Sanitizer though. 🤷‍♂️ 

Nothing in the Thread Sanitizer. 🎉

swiftlyfalling:
@jshier: Yeah, those are a false-positives. Xcode 8.2 fixed a number of the leak false-positives, but `Set` still causes them, and there are a handful of others.

While a Procedure exists, you'll see these false-positive `_NativeSetStorageImpl` leaks. But if you trigger the Memory Graph Debugger again after the Procedure deinits, they magically disappear. Hence, the "Leaks" instrument, which triggers a leak detection at specific intervals (that likely coincide with when one or more Procedures are initialized) shows them.

This same false-positive is reproducible with even a simple struct with two Sets.

The `PThreadMutex` "leak" is also a false-positive. It belongs to the LogManager, but (like the Set issues) the ownership information isn't properly being determined by the Leak detector.

Issue #515, opened 2016-10-13
Bug evaluating NoFailedDependenciesCondition with NegatedCondition


martnst:
I run into some edge case in which NoFailedDependenciesCondition to evaluate wrong.  If I add a condition that is wrapped in a `NegatedCondition`. 

```swift
let operation1 : Procedure = …
let operation2 : Procedure = …
let authorizedCondition : AuthorizedFor = ... 

operation2.addDependency(operation1)
operation2.attach(condition: NegatedCondition(authorizedCondition))
operation2.attach(condition: NoFailedDependenciesCondition())
```

Now when `operation1` finishes successfully and authorizedCondition evaluates with a negative result which should be turned into a positive by `NegatedCondition` `operation2` still won't execute. After some debugging I found that the `authorizedCondition` results in a direct dependency of `operation2` which failed and therefore lets the `NoFailedDependenciesCondition` also fail. 

I guess this is a bad side effect of running the Conditions as Operations. I think the `NoFailedDependenciesCondition` should filter Operations that are `Conditions` during the evaluation.

danthorpe:
Hmm, interesting, I will have a think.

martnst:
Hi @danthorpe, 

FYI: For now I patched it locally in [NoFailedDependenciesCondition.swift#L43](https://github.com/ProcedureKit/ProcedureKit/blob/development/Sources/NoFailedDependenciesCondition.swift#L43) as following:

```diff
-        let dependencies = procedure.dependencies
+        let dependencies = procedure.dependencies.filter { ($0 as? Condition) == nil }
```

However there is probably a better way. As mentioned the issue comes from using Operations to evaluate Conditions. Maybe such operations should be filtered inside `procedure.dependencies`. Since I did not know what side effect changed `procedure.dependencies` might have I just fixed it inside `NoFailedDependenciesCondition` for now.

danthorpe:
Hi @martnst - I haven't had a chance to look into this properly yet - but will get to it. 👍

martnst:
No worries about timing, I just wanted to provide a possible solution.

swiftlyfalling:
@martnst: This should be fixed as part of a new Conditions PR (with many other Condition improvements). See #681, and branch `feature/OPR-681_condition_improvements`.

PR #681, opened 2017-01-30
Improve Conditions

swiftlyfalling:
This PR encompasses a major refactor of Conditions in ProcedureKit, improving performance, fixing bugs, and adding new features (while also expanding test coverage and documentation).

## Highlights:

- `Conditions` are no longer `Procedure` subclasses, and are lighter-weight.
- `Condition` processing is faster, and supports short-circuit evaluation for considerable performance improvements.
- MutualExclusivity is now evaluated post-dependencies and condition evaluation (i.e. immediately before execute), resolving numerous issues and edge cases.
- Bugs have been fixed.
- Internal implementation details are no longer exposed to user code unexpectedly.
- Added a new `CompoundCondition` (and `AndCondition` / `OrCondition`) to support logical “&&” and “||” operations on `Conditions`.
- A *lot* more documentation.

## Breaking Changes (from prior ProcedureKit v4 betas):

**TL/DR:** Almost all code built using the prior beta of ProcedureKit v4 will work with these new improvements while requiring only minimal changes.

The few areas of change that may impact existing code are:

### For Condition subclasses:
- `Condition.add(dependency:)` now functions exactly the same as `Procedure.add(dependency:)`. It does **_not_** result in the framework automatically enqueuing / running the dependency.
    - **Solutions:**
        - If you add the dependency to a queue yourself (or otherwise cause it to run), continue using `add(dependency:)`.
        - If you want the framework to schedule the dependency for execution immediately prior to evaluating the `Condition`, use the new **`Condition.produce(dependency:)`** method.
- `var mutuallyExclusiveCategory: String?` is no longer available.
    - **Solutions:**
        - Use **`Condition.addToAttachedProcedure(mutuallyExclusiveCategory:)`** instead to add a mutually-exclusive category to the `Procedure` to which your custom `Condition` is added.
        - Use `var mutuallyExclusiveCategories` to query the currently-set categories (rarely needed outside the framework).
- You can no longer call `add(condition:)` on a `Condition`.
    - Since `Condition` is no longer a `Procedure` subclass, it makes no sense to add a `Condition` (which is evaluated on a `Procedure`) to a `Condition`. However, `ComposedCondition` and `CompoundCondition` provide supported alternative patterns.
    - **Solution:**
        - Investigate `ComposedCondition` and `CompoundCondition`. These support almost all common scenarios.
        - For cases in which you previously attached a `NoFailedDependenciesCondition` to a `Condition`, you can **use the new `Condition.dependencyRequirements` to automate checking for dependency failures / cancellation** (so you don't have to bother doing this yourself in your `evaluate` override).

## How Conditions are Scheduled and Evaluated

Here’s the new order of operations / evaluation / scheduling:

- All of a `Procedure`’s (direct) **dependencies** must finish.
- Then, asynchronous processing of the`Conditions` begins. For each `Condition`:
    - If it has `dependencies` or `producedDependencies`:
        - The `Condition` evaluation (call to `evaluate(procedure:completion:)`) is wrapped in a `BlockOperation`.
        - All `dependencies` and `producedDependencies` are added as dependencies to the `BlockOperation`.
        - The `producedDependencies` and `BlockOperation` are added to an internal queue.
        - (Thus, all `Condition` dependencies must finish before the wrapped `Condition` evaluation is executed.)
    - Otherwise, the `Condition`’s `evaluate(procedure:completion:)` is called directly.
- **Condition results are aggregated until the final result is known**. This may be before all Conditions are processed (as short-circuit evaluation is now supported). If the final result is known before all Conditions have been evaluated, any remaining Conditions may be skipped (and any remaining Condition-produced dependencies may be cancelled / never scheduled for execution).
- If the final result is a failure, **the attached `Procedure` is cancelled** (with / without errors depending on whether the final result is `.failure(error)` or `.success(false)`). **Processing is complete.**
- Otherwise, if the final result is a success (`.success(true)`):
    - If no mutually-exclusive categories exist on any of the `Procedure`’s `Conditions`, **the `Procedure` is ready to execute.**
    - If mutually-exclusive categories exist on any of the `Procedure`’s `Conditions`:
        - A single asynchronous lock request is made for the union of all of the mutually-exclusive categories.
        - **Once the lock request is granted**, the internal `EvaluateConditions` operation finishes, and **the `Procedure` is ready to execute**.
        - (The locks are released once the `Procedure` has finished.)

## Additional Details

- Mutual Exclusivity is now evaluated when `Procedures` are **ready to execute** (as the final step of Condition evaluation), instead of when `Procedures` are added to the queue.
    - This **resolves several deadlock and unexpected dependency scenarios**, while still ensuring that only one Procedure with each mutual exclusivity category is running simultaneously.
    - It also **improves performance** by only acquiring locks in a single request, and only when the `Procedure` is otherwise ready to execute.
- The internal `EvaluateConditions` operation - and any `Condition`-produced dependencies - are no longer added to a user-visible queue.
    - Improved encapsulation of internal implementation details.
    - Prevents unexpected user queue starvation as a result of many long-running, outstanding condition evaluations / produced dependencies.
- `Conditions` and `Condition` dependencies no longer appear in the `Procedure.dependencies`.
    - Improved encapsulation of internal implementation details.
    - Fixes edge-case bugs with things like NoFailedDependenciesCondition (issue: #515).
- `Condition` evaluation now supports **short-circuit evaluation, dramatically speeding-up performance** in many cases.
    - The result of `Condition` evaluation may be known before all `Conditions` are evaluated. (For example, if a Procedure has multiple Conditions, as soon as one of them fails, the ultimate result is guaranteed to be a failure.) Thus, we can avoid evaluating any remaining conditions as soon as the result is known.
    - **Minor Backwards-Compatibility Note**: 
        - This means that `Procedures` will fail with the **_first_ failure error that is returned by a `Condition`**.
        - `Condition` evaluation is _asynchronous_, and _the order in which a `Procedure`’s conditions are evaluated is not guaranteed_. Whichever `Condition` fails first will provide the failure error.
        - If you are relying on `Procedures` cancelling with information about _every_ failing `Condition` (or a specific `Condition` error amongst many possible), you will need to modify your expectations to match the new behavior. (This will likely only affect exhaustive test-cases that verify the cancellation errors in-depth.)
        - `OrCondition` may fail with multiple errors, as its evaluation can only be short-circuited if at least one `Condition` succeeds.

## New Features:

- **`CompoundCondition`** is now supported, to provide support for logical “&&” and “||” operations on a sequence of `Conditions`. (As a short-cut, you can also use `AndCondition` and `OrCondition`.)

    ```swift
    let condition1 = TrueCondition()
    let condition2 = FalseCondition()

    // The following are each equivalent to a Condition encompassing “condition1 || condition2”
    // i.e. if either condition1 or condition2 succeeds, the “orCondition” succeeds
    let orCondition1 = OrCondition(condition1, condition2)
    let orCondition2 = CompoundCondition(orPredicateWith: [condition1, condition2])
    let orCondition3 = condition1 || condition2
    ```

    `CompoundCondition` supports short-circuit evaluation, for improved performance.

### DIFF
--- a/Sources/ProcedureKit/Procedure.swift
+++ b/Sources/ProcedureKit/Procedure.swift
@@ -76,6 +76,70 @@ internal struct ProcedureKit {
     }
 }
 
+/**
+ Procedure is an Operation subclass. It is an abstract class which should be subclassed.
+
+ ```swift
+ import ProcedureKit
+
+ class MyFirstProcedure: Procedure {
+     override func execute() {
+         guard !isCancelled else { return }
+         print("Hello World")
+         finish()
+     }
+ }
+
+ let queue = ProcedureQueue()
+ let myProcedure = MyFirstProcedure()
+ queue.addOperation(myProcedure)
+ ```
+
+ The key points here are:
+
+ 1. Subclass `Procedure`
+ 2. Override `execute` but do not call `super.execute()`
+ 3. Check the `isCancelled` property before starting any work.
+ 4. If not cancelled, always call `finish()` after the work is done. This could be done asynchronously.
+ 5. Add procedures to instances of `ProcedureQueue`.
+
+ ### Built-in Procedures
+
+ ProcedureKit includes a number of built-in Procedure subclasses, such as `GroupProcedure`,
+ `RetryProcedure`, `RepeatProcedure`. Many of these built-in subclasses can be used as-is
+ (without subclassing them). See their documentation for information on what they do and
+ how to utilize them.
+
+ ### Cancellation
+
+ Once you add a Procedure to a ProcedureQueue, the queue takes over and handles the
+ scheduling of the task at some point in the future (based on dependencies, and
+ qualityOfService, etc).
+
+ If you later decide that you do not want to execute the procedure after all, you can cancel
+ the procedure to prevent it from running needlessly. You do this by calling the `cancel()`
+ method on the Procedure instance itself.
+
+ Cancelling a Procedure before it has been started by the queue will cause the queue to:
+    - Ignore any unfinished dependencies
+    - Automatically finish the Procedure (without calling your `execute()` override)
+
+ This helps clear the cancelled Procedure from the queue as quickly as possible. Since it
+ hasn't yet started to execute, the framework can help handle this case for you.
+
+ However, cancelling a Procedure that has already been started by the queue will *not*
+ automatically stop it from executing. It is the responsibility of your Procedure subclass,
+ once it has started to execute, to check its cancelled state and respond to cancellation
+ by finishing as quickly as possible.
+
+ You can implement this via two different methods:
+    1. Checking `isCancelled` periodically (for example, in your `execute()` override)
+    2. Adding a DidCancel observer or overriding `produceDidCancel(withErrors:)`, which will be called after your Procedure is cancelled.
+
+ Which method you should use is likely determined by how your Procedure performs the
+ bulk of its task (synchronously - in its `execute()` override - or asynchronously, for
+ which a DidCancel observer may be more useful).
+ */
 open class Procedure: Operation, ProcedureProtocol {
 
     private var _isTransitioningToExecuting = false
@@ -137,9 +201,13 @@ open class Procedure: Operation, ProcedureProtocol {
     internal let queueAddContext = ProcedureQueueContext()
 
     deinit {
+        // ensure that any EvaluateConditions operation is cancelled
+        evaluateConditionsProcedure?.cancel()
+
         // ensure that the Protected Properies are deinitialized within the lock
         stateLock.withCriticalScope {
             self.protectedProperties = nil
+            _evaluateConditionsProcedure = nil
         }
     }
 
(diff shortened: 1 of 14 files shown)


swiftlyfalling:
This PR also resolves several race conditions and TOCTOU (time-of-check time-of-use) issues.

One example:

Previously, Conditions only had an `add(dependency:)` method. ProcedureQueue attempted to handle scheduling Condition dependencies as-needed, by [filtering Condition dependencies already scheduled on the current queue](https://github.com/ProcedureKit/ProcedureKit/blob/08a0d37ba66f084073fd687bbd99a34fa9fe1fff/Sources/ProcedureKit/ProcedureQueue.swift#L397):

```swift
// Filter out the indirect dependencies which have already been added to the queue
let indirectDependenciesToProcess = indirectDependencies.filter { !self.operations.contains($0) }
```

This approach has several problems, including:
- The dependency could be scheduled on another queue.
- The dependency could be added to the queue post-check, but before the call to add.
- The dependency could have already finished (if something else previously scheduled it).

It's not safe to decide whether to add an Operation to a queue based on whether it's already on the queue.

This PR resolves both these issues in part by splitting Condition dependency management into two functions:
 - **`add(dependency:)`** adds a dependency. You are responsible for scheduling it. (Just like `Procedure.add(dependency:)`)
 - **`produce(dependency:)`** adds a dependency that will be produced and finished prior to Condition evaluation. This handles scheduling at the appropriate time for you.

Splitting these options absolves the framework from trying to do things that are necessarily race-condition-prone, and enables the programmer to specify their intent more clearly.

swiftlyfalling:
Not-Frequently-Asked-Questions:

- What if I previously added a `MutualExclusivity` condition to a `Condition`, to ensure that the `Condition` itself wasn't executed concurrently with other things?
    - Mutual Exclusivity can only be applied to `Procedures`. Since `Condition` is no longer a `Procedure` subclass, you can no longer add a `MutualExclusivity` condition directly to a `Condition`.
    - **SOLUTION**: 
        1. Extract out the code / logic that requires mutual exclusivity, and put it in a `Procedure`. (SeeAlso: `BlockProcedure`)
        2. Add the `MutualExclusivity` condition to the new Procedure.
        3. Modify your Condition to `produce(dependency:)` the new Procedure.
        4. Reduce your Condition's evaluate function to simply evaluating the output of your new Procedure (and any other logic that does not require mutual exclusivity).

        Thus, your logic that requires mutual exclusivity now lives in a `Procedure` embedded in the `Condition` (that the `Condition` produces as a dependency), and the `Condition`'s `evaluate()` function simply transforms the output of that `Procedure` to success / failure of the `Condition`.

Sources/ProcedureKit/Procedure.swift
```diff
@@ -1087,116 +1254,190 @@ internal extension Procedure {
...
+                finish()
+                return
+            }
 
-        // Add the direct dependencies of the procedure as direct dependencies of the evaluator
-        directDependencies.forEach {
-            evaluator.add(dependency: $0)
-        }
+            let conditions = procedure.conditions
+            conditions.evaluate(procedure: procedure, withContext: context) { result in
```
danthorpe: This is really nice too.

Sources/TestingProcedureKit/XCTAsserts.swift
```diff
@@ -196,6 +196,18 @@ public extension ProcedureKitTestCase {
         }
     }
 
+    public func XCTAssertProcedure<T: ProcedureProtocol, E: Error>(_ exp: @autoclosure () throws -> T, firstErrorEquals firstError: E, _ message: @autoclosure () -> String = "", file: StaticString = #file, line: UInt = #line) where E: Equatable {
```
danthorpe: Fantastic - been meaning to add this for a while!

Tests/ProcedureKitTests/ConditionTests.swift
```diff
@@ -5,11 +5,80 @@
 //
 
 import XCTest
-import TestingProcedureKit
+@testable import TestingProcedureKit
```
danthorpe: Hmm, this might be a problem. I can't remember exactly why this causes issues, but we really only want to testable import ProcedureKit itself. Can we make APIs public here instead?

PR #682, opened 2017-02-14
AsyncResultProcedure design: no way to access self in block

danthorpe:
Hi, 

I am implementing network requests as subclasses of `AsyncResultProcedure` / `AsyncBlockProcedure`. As of the current design with passing the super classes init a block there seems to be no way to access self (the subclass procedure). This prohibits accessing helper methods or the logging system. 

As of now I use this workaround approach: 

```swift
    init() {
        super.init(block: { _ in }) // passing in an empty block as it wont get called due to overriding execute()
    }
    
    // since it's not possible to access self in the block to be set in init() override execute instead
    override func execute() {
        fetch { self.finish(withResult: $0) }
    }

    func fetch(_ completion: @escaping FinishingBlock) {
        // ...
    }
```

Alternatively the `AsyncResultProcedure` could pass a reference to weak self into the block.

### DIFF
--- a/Sources/ProcedureKit/Block.swift
+++ b/Sources/ProcedureKit/Block.swift
@@ -29,7 +29,7 @@ open class BlockProcedure: ResultProcedure<Void> { }
 open class AsyncResultProcedure<Output>: Procedure, OutputProcedure {
 
     public typealias FinishingBlock = (ProcedureResult<Output>) -> Void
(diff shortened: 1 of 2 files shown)


danthorpe:
Rather annoyingly, this change will potentially break a lot of existing code (such as the failing CI task). Will just have a bit more of a think about it.

PR #699, opened 2017-04-05
Fixes Xcode 8.3 warnings for Swift 3.1

danthorpe:


danthorpe:
Most of these failures are related to a single unit test failing, which in turn is related to moving a piece of code....

Sources/ProcedureKit/Group.swift
```diff
@@ -52,6 +52,16 @@ open class GroupProcedure: Procedure, ProcedureQueueDelegate {
...
+    /// Override of errors
+    public override var errors: [Error] {
+        get { return groupStateLock.withCriticalScope { _groupErrors.fatal } }
+        set {
+            groupStateLock.withCriticalScope {
+                _groupErrors.fatal = newValue
+            }
+        }
+    }
+
```
danthorpe: @swiftlyfalling just wondering what your thoughts are on this....

Essentially this has been moved from an extension in the same file, as in Xcode 8.3 it creates a warning of "Cannot override a non-dynamic class declaration from an extension".

However, moving it into the main class definition for `GroupProcedure` causes [a failed test](https://github.com/ProcedureKit/ProcedureKit/blob/development/Tests/ProcedureKitTests/GroupTests.swift#L221):

```swift
    func test__group_cancel_with_errors_does_not_collect_errors_sent_to_children() {
        check(procedure: group) { $0.cancel(withError: TestError()) }
        XCTAssertProcedureCancelledWithErrors(group, count: 1)
    }
```

Where the test fails, as there are no group errors. Essentially, I think what is happening is that the additional error argument is not getting added to the group's fatal errors. I think, clearly, there is a timing/event issue going on.

Any thoughts?

danthorpe: I think essentially, that what is happening here is that `GroupProcedure` sort of handles errors a little bit differently so that it can have the concept of error recovery. This means, disambiguating between fatal and recovered errors. However, the error mechanisms in `Procedure` which are activated by `cancel(withError:)` and not overridden by `GroupProcedure` therefore mean that these errors are not picked up.

Although, given this explanation, I don't see why this has only just started happening now - possibly a false positive which the warnings in Xcode 8.3 has helped shine a light on.

So, possible solution would be to make the internal `_procedureDidCancel()` _micro-optimization_ in `Procedure` accept the additional errors argument. This would allow `GroupProcedure` to append the error to it's fatal array immediately.

danthorpe: Yep - this fixes it, see commit: 1bb31b62c14b05e459e945dab755557405d50557

danthorpe:
Stress tests are currently failing here. Looks like a deadlock related to cancellation.

Sources/ProcedureKit/Group.swift
```diff
@@ -156,10 +156,7 @@ open class GroupProcedure: Procedure, ProcedureQueueDelegate {
         // we must cancelAllOperations and also ensure the queue is not suspended.
         queue.cancelAllOperations()
         queue.isSuspended = false
-
-        // If you find that execution is stuck on the following line, one of the child
-        // Operations/Procedures is likely not handling cancellation and finishing.
-        queue.waitUntilAllOperationsAreFinished()
```
swiftlyfalling: (See the notes for this commit.)
The deadlock this fixes was consistently reproducible on some machines, but consistently *unreproducible* on others. Fun!

danthorpe: To be clear, this causes a dead lock on some machines - possible due to lack of hyper threading on the CPU.

danthorpe: deadlocked test run

swiftlyfalling: To add-on (and for future reference): There looks to be some difference in how things are being scheduled / pre-empted on different machines, as:
- The machine with 4 cores + Hyper-Threading **never** encountered this particular order of operations that could result in deadlock.
- The machine with 4 cores (but *no* Hyper-Threading) **consistently encountered** this particular order of operations that could result in deadlock (i.e. every time a particular Stress Test was run).
 > Overall failures only occurred in the second batch of `test__group_cancel()`, and often around 2/3 of the second batch of Group's would fail to finish.

PR #703, opened 2017-04-11
Investigate transforming output chains

danthorpe:
I quite often want to take the output of a procedure and then transform, combine, and otherwise process it into a new output. Often it is easier to achieve this by writing small procedures to perform each step. However, _ProcedureKit_ lacks a really cohesive API to chain procedures together.

This PR investigates (for like, the 3rd time I'm sure) another attempt at a "procedure chain" API.


jshier:
Interesting. So this allows you to chain normal (looking) closures onto a procedure that produces a result, in order to transform that result? All chained operations are executed on the queue of the initial procedure?

danthorpe:
@jshier exactly. Each closure is executed inside a Procedure, and ProcedureQueue gets a new method for adding a chain, so all of the Procedures get added to the same queue.

Issue #707, opened 2017-04-27
Operations scheduled on the main queue run on background threads


5sw:
Operations scheduled on `ProcedureQueue.main` (sometimes?) run on background threads, not the main thread. I don't think I have to explain why this is bad.

Reason is that the main procedure queue (class `ProcedureQueue.MainProcedureQueue`) is configured with `underlyingQueue = DispatchQueue.main`.  Apples documentation about this explicitly states that "the value of this property must not be the value returned by dispatch_get_main_queue()."

swiftlyfalling:
@5sw: You've raised several interesting points. After an initial look at this:

1.) `ProcedureQueue.MainProcedureQueue` (and, therefore, `ProcedureQueue.main`) sets the `underlyingQueue = DispatchQueue.main`. You are correct that Apple's documentation explicitly says not to do this.

The *reasons* behind this are complex, and require some careful thought and investigation.

2.) A Procedure scheduled on `ProcedureQueue.main` does not always have its `execute()` function called on the main thread.

**Importantly: 2 is not caused by 1.** It's a separate issue that I'm working on a PR to resolve (along with some solid tests around it).

---

On a related subject: what sort of use-case led you to use `ProcedureQueue.main`? Can you post some sample code?

5sw:
Ok, if 2 is not caused by 1 (which I actually find hard to believe, but then I didn't check the complete ProcedureKit code) it explains why this started happening more often after we last updated ProcedureKit, even though that part didn't change.

But no matter what the reason is, I feel very uncomfortable about leaving the main queue as the underlyingQueue. Apple could add an assertion or some other reason for this to crash with any OS update. Maybe setting the underlying queue can simply be removed, if some other code is responsible for selecting the target queue anyways?

The use case for this is rather simple - we do some work in the background (using multiple operations and dependencies), and once that is all done we have another operation that updates the UI. The UI operation depends on all the other operations and is enqueued on the main queue. If it will help I can create a short example next week.

swiftlyfalling:
@5sw: PR #708 should resolve the `execute()` issue for you.

As to the current way that `ProcedureQueue.main` is implemented, hopefully we can get @danthorpe in here to offer his thoughts as to future directions, but I'd like to make a bit of a broader point:

[...]

PR #708, opened 2017-04-28
Procedure's `execute()` function should utilize the ProcedureQueue.underlyingQueue, if set

swiftlyfalling:
If a Procedure is enqueued on a ProcedureQueue that specifies an `underlyingQueue`, the Procedure’s `execute()` function should run on the `underlyingQueue`.

### DIFF
--- a/Sources/ProcedureKit/Procedure.swift
+++ b/Sources/ProcedureKit/Procedure.swift
@@ -641,25 +641,25 @@ open class Procedure: Operation, ProcedureProtocol {
         }
 
         // Check the state again, as it could have changed in another queue via finish
-        func getNextStateAgain() -> ProcedureKit.State? {
+        func getNextStateAgain() -> (ProcedureKit.State?, ProcedureQueue?) {
             return stateLock.withCriticalScope {
-                guard _state <= .started else { return nil }
+                guard _state <= .started else { return (nil, nil) }
 
                 guard !_isHandlingFinish else {
                     // a finish is pending, simply exit from processing execute
-                    return nil
+                    return (nil, nil)
                 }
 
                 if _isCancelled && !isAutomaticFinishingDisabled && !_isHandlingFinish {
                     // Procedure was cancelled, and automatic finishing is enabled.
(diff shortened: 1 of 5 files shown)


5sw:
I looked over this, and it should solve my problem from #707. If I understand everything correctly the `start` method of an operation is dispatched by `NSOperationQueue` on it's own dispatch queue or the underlying queue. This queue doesn't really matter as any further code is dispatched to the event queue. This also explains why the actual work used to run on some background thread and not the underlying queue. In this PR you add code that then dispatches the actual work onto the underlying queue if it is set.

If I don't miss anything this means that the actual underlying queue of the `NSOperationQueue` doesn't matter at all. If this is true I would recommend to create a new property for the target queue and leave the inherited `underlyingQueue` property alone. Then we are safe to use the main queue as the target queue, even if Apple decides to add an assertion.

swiftlyfalling:
> If I don't miss anything this means that the actual underlying queue of the NSOperationQueue doesn't matter at all.

@5sw: Since this ultimately addresses your separate issue concerning the `ProcedureQueue.main` implementation, I've [responded in the prior issue](https://github.com/ProcedureKit/ProcedureKit/issues/707#issuecomment-298182457).

Sources/ProcedureKit/Procedure.swift
```diff
@@ -696,7 +696,39 @@ open class Procedure: Operation, ProcedureProtocol {
         log.notice(message: "Will Execute")
 
         // Call the execute() function (which should be overriden in Procedure subclasses)
+        if let underlyingQueue = queue?.underlyingQueue {
+            // The Procedure was enqueued on a ProcedureQueue that specifies an `underlyingQueue`.
+            //
+            // Explicitly call the `execute()` function on the underlyingQueue, while also
+            // pausing dispatch of any new blocks on the Procedure's EventQueue until the call
+            // to `execute()` returns, to ensure that `execute()` occurs on the underlyingQueue
+            // *and* non-concurrently with this Procedure's EventQueue.
```
danthorpe: Excellent!

PR #714, opened 2017-05-08
Fix a (rare) data race in DelayProcedure

swiftlyfalling:
Fix a (rare) data race in `DelayProcedure` (discovered via Thread Sanitizer) by ensuring that the `DispatchSourceTimer` is initialized to use the `DelayProcedure`’s internal EventQueue. (The race involved the release of the `DispatchSourceTimer`’s `eventHandler` block.)

### DIFF
--- a/Sources/ProcedureKit/Delay.swift
+++ b/Sources/ProcedureKit/Delay.swift
@@ -121,7 +121,7 @@ public class DelayProcedure: Procedure {
         switch delay.interval {
         case (let interval) where interval > 0.0:
             guard !isCancelled else { return }
-            _timer = eventQueue.makeTimerSource(flags: [])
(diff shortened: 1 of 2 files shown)


Sources/ProcedureKit/Delay.swift
```diff
@@ -121,7 +121,7 @@ public class DelayProcedure: Procedure {
         switch delay.interval {
         case (let interval) where interval > 0.0:
             guard !isCancelled else { return }
-            _timer = DispatchSource.makeTimerSource(flags: [], queue: DispatchQueue.default)
+            _timer = eventQueue.makeTimerSource(flags: [])
```
jshier: Looks like you've got a default argument for `flags`, so the explicit empty set is unnecessary.

PR #737, opened 2017-05-30
Simplify GroupProcedure child error handling behavior & customization

swiftlyfalling:
Key changes:
- **`GroupProcedure`** child error handling is now centralized in a single **`child(_:willFinishWithErrors:)`** method, which is called for all finishing child Procedures (even if the errors array is empty). The default implementation in `GroupProcedure` appends the received child errors to the Group’s errors.
- **`GroupProcedure`** now uses an internal `GroupProcedureDelegate` object to forward delegate callbacks from its internal queue. Thus, `GroupProcedure` itself is no longer a `ProcedureQueueDelegate` (nor does it expose all the `ProcedureQueueDelegate` methods).
- **`RepeatProcedure`** and **`RetryProcedure`** now override the `child(_:willFinishWithErrors:)` method, providing a clearer execution flow (instead of jumping back-and-forth between multiple different overridden methods and GroupProcedure's internals and GroupProcedure's ProcedureQueueDelegate, etc).

Diff:

```diff
open class GroupProcedure: Procedure {
    ...
-   open func child(_ child: Operation, willAttemptRecoveryFromErrors errors: [Error]) -> Bool
-   open func childWillFinishWithoutErrors(_ child: Operation)
+   open func child(_ child: Procedure, willFinishWithErrors errors: [Error])
    ...
-   final public func append(fatalError error: Error)
+   final public func append(error: Error, fromChild child: Operation? = nil)
-   final public func append(fatalErrors errors: [Error])
+   final public func append(errors: [Error], fromChild child: Operation? = nil)
-   final public func child(_ child: Operation, didEncounterFatalError error: Error)
-   final public func child(_ child: Operation, didEncounterFatalErrors errors: [Error])
-   final public func childDidRecoverFromErrors(_ child: Operation)
-   final public func childDidNotRecoverFromErrors(_ child: Operation)
    ...
-   func procedureQueue(_ queue: ProcedureQueue, willAddOperation operation: Operation, context: Any?) -> ProcedureFuture?
-   func procedureQueue(_ queue: ProcedureQueue, didAddOperation operation: Operation, context: Any?)
-   func procedureQueue(_ queue: ProcedureQueue, didFinishOperation operation: Operation)
-   func procedureQueue(_ queue: ProcedureQueue, willAddProcedure procedure: Procedure, context: Any?) -> ProcedureFuture?
-   func procedureQueue(_ queue: ProcedureQueue, didAddProcedure procedure: Procedure, context: Any?)
-   func procedureQueue(_ queue: ProcedureQueue, willFinishProcedure procedure: Procedure, withErrors errors: [Error]) -> ProcedureFuture?
-   func procedureQueue(_ queue: ProcedureQueue, didFinishProcedure procedure: Procedure, withErrors errors: [Error])
}

public struct RetryFailureInfo<T: Procedure> {
-   public let historicalErrors: [Error]
}
```

### DIFF
--- a/Sources/ProcedureKit/Group.swift
+++ b/Sources/ProcedureKit/Group.swift
@@ -15,34 +15,16 @@ import Dispatch
  related operations together, thereby creating higher
  levels of abstractions.
  */
-open class GroupProcedure: Procedure, ProcedureQueueDelegate {
-
-    internal struct GroupErrors {
-        typealias ByOperation = Dictionary<Operation, Array<Error>>
-        var fatal = [Error]()
-        var attemptedRecovery: ByOperation = [:]
-
-        var attemptedRecoveryErrors: [Error] {
-            return Array(attemptedRecovery.values.flatMap { $0 })
-        }
-
-        var all: [Error] {
-            get {
-                var tmp: [Error] = fatal
-                tmp.append(contentsOf: attemptedRecoveryErrors)
-                return tmp
-            }
-        }
-    }
+open class GroupProcedure: Procedure {
 
     internal let queue = ProcedureQueue()
+    internal var queueDelegate: GroupQueueDelegate!
 
     fileprivate let initialChildren: [Operation]
     fileprivate var groupCanFinish: CanFinishGroup!
     fileprivate var groupStateLock = PThreadMutex()
 
     // Protected private properties
-    fileprivate var _groupErrors = GroupErrors() // swiftlint:disable:this variable_name
     fileprivate var _groupChildren: [Operation] // swiftlint:disable:this variable_name
     fileprivate var _groupIsFinishing = false // swiftlint:disable:this variable_name
     fileprivate var _groupIsSuspended = false // swiftlint:disable:this variable_name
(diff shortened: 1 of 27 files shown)


jshier:
A long time ago I did use the delegate overrides in a subclass in order to dynamically add procedures based on whether some previous procedure succeeded or failed, but I'm not sure that would be needed any more, and there're likely better ways to accomplish the same thing now.

swiftlyfalling:
@jshier: Yeah. The problem with having those ProcedureQueueDelegate methods exposed is: It's easy to break the GroupProcedure's internal behavior, and it's easy to accidentally do things that open up unexpected behavior because of concurrency.

Whereas with this PR, you can just override `child(_:willFinishWithErrors:)`, (or use other classic methods like adding observers to the appropriate children).

jshier:
Sounds right. I think my usage was pre child error handling, so it was all unified in one path. This looks good!

danthorpe:
Not had a chance to look yet - will do later tonight / tomorrow morning.

With that in mind - @swiftlyfalling, is there an option of allowing error handling via a delegate to avoid having to subclass? Is that something which should be encouraged? Or do we consider GroupProcedure to be effectively an abstract class which must be subclassed?

jshier:
Considering it abstract would be a huge change, given you can use it without subclassing right now. Removing that ability now would seem to necessitate a replacement type that can be used directly to just bundle up procedures.

Issue #752, opened 2017-07-10
ProcedureObserver fails to compile in Xcode 9 beta 3


appsird:
I receive a new error when compiling with said release:

![image](https://user-images.githubusercontent.com/8471833/28039078-dce7c3b8-6575-11e7-8372-d5838d92d9b1.png)

danthorpe:
Oh, thanks @appsird not checked with Beta 3 yet.

danthorpe:
Looks like this has already been [reported](https://bugs.swift.org/browse/SR-5265?jql=text%20~%20%22ProcedureKit%22) as a "won't fix" bug - annoyingly I didn't see it until now.

Will investigate.

danthorpe:
This is a bit of a serious situation, as the "construct" which allows complete typing of a procedure inside an observer is actually considered a weakness of the generics system, which has now been fixed. Hmm. We're investigating and considering the approach.

However, for anyone who is on the bleeding edge (_ProcedureKit_ `development` with Xcode 9 betas) - then for now, stay on Beta 2.

Any fixes for this will likely bump us to version 5 (a bit earlier than I planned, but probably unavoidable).

dimat:
It seems that beta 3 now supports only swift 3.2 as the lowest version, but if I open my project in xcode 8, then it shows that procedurekit is on swift 3.0

danthorpe:
Hi - @dimat, yep, that is as expected.

swiftlyfalling:
As an update, #754 should fix this issue - and be sufficiently backwards-compatible to make it into a future 4.x point release. Will be merged into `development` soon.

appsird:
Thanks Daniel.

dimat:
Thanks a lot!
Any chance to bump the version as well?

swiftlyfalling:
@dimat: This should make it into the next 4.x release. Not sure exactly what the timeframe looks like for that just yet.

danthorpe:
Was hoping to get the documentation guides done before doing v4.3. Will bump a new version this week though, and maybe doc site will be v4.4.

PR #753, opened 2017-07-11
Remove associatedtype from ProcedureObserver

danthorpe:
This is to work around the changes in Xcode 9 Beta 3 as raised by issue #752.

### DIFF
--- a/Sources/ProcedureKit/AnyObserver.swift
+++ b/Sources/ProcedureKit/AnyObserver.swift
@@ -5,7 +5,7 @@
 //
 
 import Foundation
-
+/*
 class AnyObserverBox_<Procedure: ProcedureProtocol>: ProcedureObserver {
     func didAttach(to procedure: Procedure) { _abstractMethod() }
     func will(execute procedure: Procedure, pendingExecute: PendingExecuteEvent) { _abstractMethod() }
@@ -107,3 +107,4 @@ public struct AnyObserver<Procedure: ProcedureProtocol>: ProcedureObserver {
         return box.eventQueue
     }
 }
+*/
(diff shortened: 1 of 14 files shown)


swiftlyfalling:
### Breaking Changes:

#### Observers

- `ProcedureObserver` no longer has an associated type.
- `AnyObserver` no longer exists. Because `ProcedureObserver` no longer has an associated type, it can now be used in place of `AnyObserver`.
- The `ProcedureObserver` methods now take a `procedure` argument of type **`ProcedureProtocol`** (instead of the prior `Procedure` associated type).
	- **IMPORTANT BREAKING CHANGE**:
		- Custom `ProcedureObserver`s must implement the tweaked `ProcedureObserver` method signatures to have their functions called. Review `ProcedureObserver.swift` for the adjusted methods. If you previously relied on the associated type of `ProcedureObserver`, you will need to:
			1. Change your method signature to expect a `procedure` argument of type `ProcedureProtocol`.
			2. Use a conditional downcast to cast to your desired type (for example, a `Procedure` subclass) if needed. (An example of this can be seen in the implementation of `BlockObserver`.)

swiftlyfalling:
Some notes:

- Prior to ProcedureKit v4.3.0, attempting to add a custom `ProcedureObserver` that expects a non-`Procedure` type (for example, a `Procedure` subclass) did not compile.
    Example:
    ```swift
    class ExampleProcedure: Procedure {
        var testingVal: Bool { return true }
        override func execute() { finish() }
    }
    class ExampleObserver: ProcedureObserver {
        typealias Procedure = ExampleProcedure
        func didAttach(to procedure: ExampleProcedure) {
            print ("\(procedure.testingVal)")
        }
    }
    let procedure = ExampleProcedure()
    procedure.add(observer: ExampleObserver()) // Generic parameter 'Observer' could not be inferred. (Xcode 8.3.x)
    ```
- v4 use of `BlockObserver`, `WillFinishObserver`, etc without explicitly specifying the type defaulted to `BlockObserver<Procedure>` (etc).

PR #759, opened 2017-07-16
Remove push Travis Trigger

danthorpe:
Thanks to comment: https://github.com/ProcedureKit/ProcedureKit/pull/757#issuecomment-314947291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,3 @@ jobs:
       script: bundle exec fastlane mac test_mac
     - stage: Stress Test
       script: bundle exec fastlane mac stress_test
-branches:
-  only:
-    - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: objective-c
 osx_image: xcode8.3
-branches:
-  only:
-    - master
 jobs:
   include:
     - stage: ProcedureKit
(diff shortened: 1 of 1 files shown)


danthorpe:
@jshier err, not sure about this. Seems like it doesn't build anything adding in your suggestion to `travis.yml`. We still want to run the travis job on PRs. Doesn't seem to be a way to get it to build _only_ PRs.

I don't know, I'm not all that familiar with TravisCI.... any thoughts?

jshier:
Odd. We use this setting on Alamofire just fine. It's also the [accepted answer](https://stackoverflow.com/a/31882307/272952). Perhaps the stages feature is incompatible?

jshier:
I can't see the repo settings, but perhaps make sure the Travis integration is through the actual Travis integration and not a manual webhook?

danthorpe:
Yeah, I think it might be the stages.

- [x] Saw this "accepted answer" - copied settings exactly.
- [x] Setup via GitHub "Integration" not a manual web hook.

Hmm

jshier:
Stages and this setting seem like they should be orthogonal but perhaps not. I read through most of the stages issue for Travis and no one mentioned an issue like this, but I did skim towards the end.

PR #762, opened 2017-07-19
Clarification for using Cocoapods

alimoeeny:
Just made it more clear, I think.


danthorpe:
@alimoeeny The CI failure here is our CocoaPods integration tests, which don't work for 3rd party forks - no matter.

PR #763, opened 2017-07-20
[Documentation]: Link to reference docs

alimoeeny:
Really not sure what the plan with the migration is, since I read all the old docs and found them beneficial but uncomfortable. 
Or maybe I have missed something glaringly obvious about what is going on. 
In any case I thought you may have not yet got the time to add the link.


danthorpe:
Thanks for this! Yeah, I have got a branch which I'm currently working on: #751, and the documents are here: http://procedure.kit.run/feature/OPR-751_documentation_guides/.

PR #770, opened 2017-07-24
Fix unstable test: `test__profile_operation__which_produces_child`

swiftlyfalling:
Previously, there was no guarantee that the produced child would finish before the wait completed.

### DIFF
--- a/Tests/ProcedureKitTests/ProfilerTests.swift
+++ b/Tests/ProcedureKitTests/ProfilerTests.swift
@@ -88,10 +88,14 @@ class ProfilerTests: ProcedureKitTestCase {
     func test__profile_operation__which_produces_child() {
 
         let child = TestProcedure()
+        // Also wait for the produced child to complete
+        addCompletionBlockTo(procedure: child)
 
         procedure = TestProcedure(produced: child)
         procedure.add(observer: profiler)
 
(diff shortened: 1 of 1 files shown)


Tests/ProcedureKitTests/ProfilerTests.swift
```diff
@@ -88,10 +88,14 @@ class ProfilerTests: ProcedureKitTestCase {
...
         let child = TestProcedure()
+        // Also wait for the produced child to complete
+        addCompletionBlockTo(procedure: child)
 
         procedure = TestProcedure(produced: child)
         procedure.add(observer: profiler)
 
+        // Because of the addCompletionBlockTo line above, wait for the procedure *and*
+        // the child it produces to complete
         wait(for: procedure)
```
danthorpe: Why not put in `wait(for: procedure, child)`?

swiftlyfalling: The TestProcedure produces the `child`, hence calling `wait(for: child)` would result in attempting to add it to a queue twice. (Since `wait(for:)` adds the child to the queue.)

danthorpe: oh, right you are!

PR #777, opened 2017-07-26
Simplify Procedure ConditionEvaluator state management

swiftlyfalling:
Also fixes a rare race condition in which Conditions could begin to be evaluated prior to all of the parent Procedure’s dependencies finishing.


danthorpe:
@swiftlyfalling - will look at this tomorrow. 👍

Issue #792, opened 2017-08-23
Lock/race condition may occurs when adding children to a GroupProcedure with MutuallyExclusive condition


yageek:
I created a [sample project](https://github.com/yageek/MutualExclusiveError) helping to reproduce the error. I'm using ProcedureKit 4.3.0.

In certain conditions, ProcedureKit seems to be locked if the children of a `GroupProcedure` are added after call to the initializer and one of the children contains a `MutuallyExclusive` condition.

Inside the sample code, the code related to the issue seems to be the following:

```
   init() {
        super.init(operations: [])
        name = "TopLevelGroup"

        let op = SimpleLowLevelProcedure()
        add(children: [op])
}
```
Removing the `MutuallyExclusive` condition solves the issue.
Adding the child inside the initializer instead of using `add` solves also the issue:
```
    init() {
     let op = SimpleLowLevelProcedure()
        super.init(operations: [op])
        name = "TopLevelGroup"
    }
```

swiftlyfalling:
@yageek: I have a theory about what might be going on here, but it'll be a little bit before I have a chance to test it out. Have a fix in mind.

danthorpe:
@swiftlyfalling I might have a bit of time to try out any theories if you want to discuss it. Had a play with @yageek's demo project yesterday with him - confirming the bug.

swiftlyfalling:
@yageek, @danthorpe: The core problem is that condition evaluation (and several other steps, like mutual exclusivity locks) does not properly respect and handle the associated ProcedureQueue's `isSuspended` status.

The solution is complicated by the asynchronous nature and relationship of a number of tasks (condition evaluation, mutual exclusivity lock requests, suspending / un-suspending ProcedureQueues, starting Procedures, etc).

I should have a solid fix (and a bunch of additional tests) soon. This is a very nice edge-case catch!

danthorpe:
Thanks for the overview of the underlying issue @swiftlyfalling!

swiftlyfalling:
@yageek: A fix for this has now been merged into `development`. Please give it a try.

yageek:
I just tested again with the sample app and I am not able to reproduce the issue. Great work 👍

swiftlyfalling:
Fix now available in 4.4.0.

PR #795, opened 2017-08-25
Fix issues with Conditions and suspended ProcedureQueues

swiftlyfalling:
- Condition evaluation now properly waits until the associated ProcedureQueue is not suspended to begin.
- An edge-case bug involving Mutual Exclusivity and ProcedureQueue suspension is now resolved.
- Added lots of additional tests involving condition execution timing and mutual exclusivity.

References: #792


danthorpe:
@swiftlyfalling wow, will read through & review this in depth this weekend.

PR #799, opened 2017-09-19
Fix build using Carthage on Xcode 8.3.3

pomozoff:


danthorpe:
Hi @pomozoff - can you give a bit of info on this?

pomozoff:
I've tried to build a release v4.3.2 with the lastest Carthage v0.25.0:
`$ carthage update --no-use-binaries --platform iOS Operations`
Here are errors from Swift Lint v0.22.0:
[error #1](https://gist.github.com/pomozoff/4d02b09ed2154b214e1d2c579c2df18e#file-carthage-xcodebuild-edklzl-log) - disable rules in code
[error #2](https://gist.github.com/pomozoff/6118fb267c5665789e58963894e8e775#file-carthage-xcodebuild-st1ubf-log) - valid_docs

swiftlyfalling:
Looks like swiftlint errors. @danthorpe: Maybe a new version of swiftlint requires a tweaked `.swiftlint.yml` file?

danthorpe:
Looking into this now. To be honest, at this point, I'm more inclined to just get rid of SwiftLint from this project. For the following reasons:

1. They've had 42 releases, and are at version 0.22 - so, semantic versioning is certainly not followed, as pretty much almost every version breaks previous rule sets.
2. Without using SwiftLint via CocoaPods or similar, it's actually pretty tricky to pin the version of SwiftLint to the project.
3. At this point, I think the review process and body of code in this project is mature enough that the benefit of SwiftLint is minimal.

**Therefore, I'm proposing that we remove the SwiftLint build steps to avoid these issues.**

Happy to investigate other linting solutions which are more stable in the future.

danthorpe:
This is now fixed on `development` - new version forthcoming.

PR #802, opened 2017-09-20
Update iOS simulators to iOS 11

danthorpe:


danthorpe:
Had to comment out a couple of tests related to `QualityOfService`. There are some relevant changes in the [releases notes](https://developer.apple.com/library/content/releasenotes/Foundation/RN-Foundation/index.html) of iOS 11:

> NSOperation

> NSOperation now responds to KVO and KVC for representing the finished, cancelled, executing and ready states as the strings @“finished”, @“cancelled”, @“executing” and @“ready” in addition to the older versions like @“isReady”.

> NSOperation completionBlock now uniformly is set to nil upon completion of an operation to avoid circular references to operations.

> NSOperationQueue

> NSOperationQueue will no longer create a serial backing dispatch queue for each operation. Instead it now shares a single concurrent dispatch queue.

> Quality of service for NSOperations when enqueued to an operation will no longer be calculated in advance, instead the quality of service for an operation is only applied if it is requested upon the execution of an operation. If a NSOperationQueue specifies a quality of service, the backing dispatch queue will have that specified quality of service. Any operations run upon that queue will inherit the effective quality of service if they do not have a quality of service specified. However if an NSOperation has a quality of service specified it will enforce that quality of service when running upon the NSOperationQueue’s backing dispatch queue.

> Quality of service overrides now only apply when waiting for operations to finish. This means that the work done to calculate the effective quality of service is not paid up-front but instead calculated when waiting is invoked.

> NSOperationQueueDefaultMaxConcurrentOperationCount should no longer hit thread exhaustion states.

Issue #808, opened 2017-10-18
'scheduleOneshot(deadline:leeway:)' is deprecated


ryanjm:
I'm upgrading my project to Swift 4 and updated ProcedureKit to 4.4.0. I got the following warning in Xcode:

```
.../ProcedureKit/Sources/ProcedureKit/Delay.swift:134:21: 'scheduleOneshot(deadline:leeway:)' is deprecated: renamed to 'schedule(deadline:repeating:leeway:)'
```

There are two instances of it. One in [Delay.swift](https://github.com/ProcedureKit/ProcedureKit/blob/development/Sources/ProcedureKit/Delay.swift#L134) and one in [TimeoutObsever.swift](https://github.com/ProcedureKit/ProcedureKit/blob/development/Sources/ProcedureKit/TimeoutObserver.swift#L88).

Is there a Swift 4 specific branch I should be using? I'd be happy to create a PR for this, but I think this would be a Swift 4-only change and not backward compatible so I'm not sure where that'd fit in with the other branches.

swiftlyfalling:
@ryanjm: I'm not seeing this when compiling ProcedureKit 4.4.0 with Xcode 9.0.1 with the default settings.

However, if you upgrade the ProcedureKit project to Swift 4.0, you will then see this warning. The fix is simple - don't do that. 😊

The Swift 4 compiler is built to compile modules using Swift 3 and Swift 4 into the same application. ProcedureKit 4.4.0's project file is configured to properly use Swift 3.x mode even when using Xcode 9. You can use it as-is with a Swift 4.0 project without upgrading it.

The next major release of ProcedureKit will likely be Swift 4+ only. Until then, ProcedureKit 4.4.0 is designed to support both Xcode 8 and Xcode 9.

(That all said, I have a PR pending that will remove this warning *if* someone upgrades the ProcedureKit project.)

ryanjm:
Thanks. I guess I was upgrading the ProcedureKit project to Swift 4 as well. Thanks.

danthorpe:
@ryanjm @swiftlyfalling - I finish work in a couple of weeks for the rest of the year, so will be devoting some proper time to fixing Swift 4+ issues. This is the first one.

swiftlyfalling:
As a heads-up, PR #809 fixes this.

rayfix:
@danthorpe Wondering if you had any plans for shepherding the PR in or if someone else could.  Thank you. :]

[...]

PR #809, opened 2017-10-18
Resolves some warnings when compiling in Swift 4 mode

swiftlyfalling:


rayfix:
Are these checks false positives?  It would be great to get these fixes to #808 so that I didn't have to fork the repo. :]  (I like to build warnings-as-errors and Swift 4 migration is available is a warning which brings me here. Hah.)

danthorpe:
Done in #816

PR #822, opened 2018-02-13
Use swift version DSL

fuxx:
Since CocoaPod 1.4.0 a new property has been introduced which sets the correct swift version in a project. That means that CocoaPod is now handling what ever has to be done in the project to set the correct compiler version.

Maybe at some point you may want to create a version including this new property :)


fuxx:
I cant see the errors of buildkite as they require a login. Most likely the use a old CocoaPod version or the compiler is maybe 3.2 instead of 3.0. But i think you got the point of the DSL already :)

danthorpe:
Interesting. I guess this is a change introduced by CocoaPods 1.4.0 however, so I wouldn't want to remove the existing settings yet. No idea what adoption rate of CocoaPods versions is like.

I think what we'll do is release #821 for Swift 4.0 now, as is, and will make this change in `development` branch for the next release.

PR #833, opened 2018-05-03
Result Injection improvements

danthorpe:
Result injection works really well to chain units of work together, mapping state via a sequence of `Input -> Output` transitions.

However, the common scenario, once this has been built will be to encapsulate those smaller units of work with a `GroupProcedure` to hide the details of that state transition. At the moment, this is totally possible, but it does require unnecessary boiler plate, which we can remove.

Essentially, we want to be able to do something like this:

```swift
final class MyGroup: GroupProcedure, InputProcedure, OutputProcedure {

    typealias StageA = TransformProcedure<Foo, Bar>
    typealias StageB = TransformProcedure<Bar, Baz>
    typealias StageC = TransformProcedure<Baz, Bat>

    var input: Pending<Foo>
    var ouput: Pending<ProcedureResult<Bat>>

    init() {
 
        let a = StageA { /* etc */ }
        let b = StageB { /* etc */ }.injectResult(from: a)
        let c = StageC { /* etc */ }.injectResult(from: b)

        super.init(operations: [a, b, c])

        // We want to "bind" the input property 
        //   of the StageA procedure to the group itself
        bind(to: a)

        // We want to "bind" the output property 
        //    of the group (i.e. self) to the StageC procedure output
        bind(from: c) 
    }
}
```

Those two "bind" APIs will use observers to set the `input` and `output` properties to the equivalent of the property on the receiver.

### DIFF
--- a/Sources/ProcedureKit/AnyObserver.swift
+++ b/Sources/ProcedureKit/AnyObserver.swift
@@ -8,6 +8,7 @@ import Foundation
 
 class AnyObserverBox_<Procedure: ProcedureProtocol>: ProcedureObserver {
     func didAttach(to procedure: Procedure) { _abstractMethod() }
+    func didSetInputReady(on procedure: Procedure) { _abstractMethod() }
     func will(execute procedure: Procedure, pendingExecute: PendingExecuteEvent) { _abstractMethod() }
     func did(execute procedure: Procedure) { _abstractMethod() }
     func did(cancel procedure: Procedure, withErrors errors: [Error]) { _abstractMethod() }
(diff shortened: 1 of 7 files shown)


slessans:
@danthorpe could you provide a hint or reference for the boilerplate required to do this when working with the 4.5.0 version of the library?

danthorpe:
@slessans the bind APIs are not in v4.5.0, they will be in v5.0.0. They work pretty much as in this boiler-plate code.

The development branch is 5.0.0.beta.1. There are some breaking changes however from 4.5.0. You might want to do a test to see how the changes impact your codebase - I’ve added deprecation Fix-its where possible to make it easy to upgrade.

Also be aware that more breaking changes around logging (beta.2) & potentially block procedures (beta.3) are in the works - so might want to hold off. 

Please post feedback on any upgrade issues.

PR #839, opened 2018-05-08
Fixes potential memory leak

danthorpe:
Removes a strong reference to self.

### DIFF
--- a/Sources/ProcedureKit/Block.swift
+++ b/Sources/ProcedureKit/Block.swift
@@ -41,7 +41,10 @@ open class AsyncResultProcedure<Output>: Procedure, OutputProcedure {
     }
 
     open override func execute() {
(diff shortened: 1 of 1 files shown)


Sources/ProcedureKit/Block.swift
```diff
@@ -41,7 +41,10 @@ open class AsyncResultProcedure<Output>: Procedure, OutputProcedure {
     }
 
     open override func execute() {
-        block { self.finish(withResult: $0) }
+        block { [weak self] result in
+            guard let strongSelf = self else { return }
+            strongSelf.finish(withResult: result)
```
jshier: Since there's only one line, I usually just use the weak reference, since making it strong is usually to stop the weak reference from nil'ing out in the middle of several lines.

PR #848, opened 2018-06-07
ProcedureKit Instruments

danthorpe:
Still pretty early days here, but looks like a fairly obvious easy win, for _ProcedureKit_ (the project/org) to produce an Instrument for use in Xcode 10 to help diagnose, debug & optimise framework usage.

Creating this issue as a holding place as I will probably have a look into this while at WWDC.


danthorpe:
instruments example

👆 this is what this looks like. Unfortunately, the Instruments UI does not support arbitrary nested hierarchies. Created a radar.

jshier:
Looks great. Now if only I could get the signposts to show up for Alamofire.

ArkadiYoskovitz:
@danthorpe A question: where can i find the client app element that you're using?

I wanted to try some work on this but i can't locate the client app that your using it's not in the git...

P.S. Are you sure that it's a bug maybe we need a different configuration?

danthorpe:
Hey, @ArkadiYoskovitz I have been making (in my spare time) a sample app of _Earthquakes_ to demonstrate _ProcedureKit_. It's currently closed source, but I will open source it when ready as part of v5 of ProcedureKit.

---

If you want to play around though, you could check out this branch, and do the following:

1. Use Xcode 10, and either an iOS device running iOS 12, or a Mac running Mojave.
2. Use a client app which depends on _ProcedureKit_ - and use the build products from this branch in the client app.
3. Build the ProcedureKitInstruments target, and it will load into Instruments and will allow you see the procedures which are running in your app. At the moment, this is pretty rudimentary. But it does show some promise - I've got some work to do on it before it can be safely merged.

---

So, I've spent a fair bit of time in the Custom Instruments labs at WWDC this year, to further understand how this all works, and get this branch to where it is at the moment. I spent about an hour going through what I want to do with Chad from the session: https://developer.apple.com/videos/play/wwdc2018/410/

He was pretty clear that it's not a bug, its just not a visual representation that they've needed up to now - tellingly there isn't an NSOperation/Queue instrument.

PR #851, opened 2018-06-15
Refactor Errors

danthorpe:
In _ProcedureKit_ the core type, `Procedure` can be finished with multiple `Error`s. This is a fundamental issue, and significantly contributes to the difficulty of handling errors. 

It is also unnecessary, as each `Procedure` instance should properly define its own error type, and take ownership of defining the behaviour where there are multiple errors.

- [x] Remove `[Error]` as finishing argument.
- [x] Clean up tasks related to above in places like observers etc


danthorpe:
@swiftlyfalling - not sure if you want to look into these changes.

Essentially, I've changed `Procedure` to only receive an `Error?` property instead of `[Error]`. This is to remove a complexity when it comes to understanding how to process/deal with errors.

PR #865, opened 2018-07-08
Make Location Procedures Testable in client code

ericyanush:
This PR exposes the ability for client code to inject mock LocationFetcher (CLLocationManager), and ReverseGeocoder (CLGeocoder) into location procedures. 

This is necessary to allow any client code that depends on or uses, especially in GroupProcedures, a procedure from ProcedureKitLocation. Without these changes, client code is subject to the state of the device/simulator during unit tests, making it impossible to validate behaviour when location services is disabled, no location is available, etc.


danthorpe:
Hi @ericyanush will look at this asap. Thanks for the contribution!

PR #875, opened 2018-08-05
Refactor BlockProcedure

danthorpe:
- [x] Remove API complexity with `BlockProcedure` and `AsyncProcedure`
- [x] Support accessing `self` inside the block, to expose `log` etc
- [x] Maintain existing APIs with deprecation warnings

### DIFF
--- a/ProcedureKit.xcodeproj/project.pbxproj
+++ b/ProcedureKit.xcodeproj/project.pbxproj
@@ -54,6 +54,7 @@
 		65A72C2B1DA7F37C008597CA /* TestingProcedureKit.framework in Copy Dependencies */ = {isa = PBXBuildFile; fileRef = 65CFC6161D60900000CAD875 /* TestingProcedureKit.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		65AB7DDA209B95B400726796 /* Batch.swift in Sources */ = {isa = PBXBuildFile; fileRef = 65AB7DD9209B95B400726796 /* Batch.swift */; };
 		65AEF19F20A3AE340089B6C8 /* ViewControllerContainment.swift in Sources */ = {isa = PBXBuildFile; fileRef = 65AEF19E20A3AE340089B6C8 /* ViewControllerContainment.swift */; };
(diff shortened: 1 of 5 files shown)


danthorpe:
Now, the following is possible:

1. Simple synchronous block (existing functionality):
```swift
let block = BlockProcedure { 
    print("Hello World")
}
```

2. Synchonous block, accessing the procedure inside the block:
```swift
let block = BlockProcedure { procedure in
    procedure.log.debug.message("Hello World")
    procedure.finish()
}
```
Note that here, the block is responsible for finishing itself - i.e. call `.finish()` or `.finish(with:)` to finish the Procedure. Using this initializer, by default, `BlockProcedure` will add a `TimeoutObserver` to itself, using `defaultTimeoutInterval` which is set to 3 seconds. This can be modified if needed.
```swift
BlockProcedure.defaultTimeoutInterval = 5
```

3. `AsyncBlockProcedure` is now deprecated, but still works as before. For example:
```swift
let block = AsyncBlockProcedure { finishWithResult in
    DispatchQueue.default.async {
        print("Hello world")
        finishWithResult(success)
    }
}
```
but this will raise a deprecation warning. This can now be achieved with `BlockProcedure`, as:
```swift
let block = BlockProcedure { procedure in
    DispatchQueue.default.async {
        print("Hello world")
        procedure.finish()
    }
}
```

4. `CancellableBlockProcedure` is deprecated, just use `BlockProcedure` and check `isCancelled` directly, e.g.
```swift
let block = BlockProcedure { procedure in
    guard !procedure.isCancelled else { procedure.finish() }
    DispatchQueue.default.async {
        print("Hello world")
        procedure.finish()
    }
}
```

5. `ResultProcedure` as been re-written as a subclass of `BlockProcedure` (previously, it was the superclass). Existing functionality has been maintained:
```swift
let hello = ResultProcedure { "Hello World" }
```
For completeness sake, it also supports receiving itself as a first argument to the block:
```swift
let hello = ResultProcedure { procedure in
    procedure.log.debug.message("Making a string")
    return "Hello World"
}
```

6. `AsyncResultProcedure` as before - just re-written but is still a subclass of `ResultProcedure`.

7. `CancellableResultProcedure` has been deprecated - just use `ResultProcedure`

PR #882, opened 2018-08-24
Refactor `AlertProcedure`

danthorpe:
As `UIProcedure` is deprecated, `AlertProcedure` needs some TLC.

### DIFF
--- a/Sources/ProcedureKitMobile/Alert.swift
+++ b/Sources/ProcedureKitMobile/Alert.swift
@@ -7,12 +7,31 @@
 import Foundation
 import UIKit
 
-public class AlertProcedure: UIProcedure {
+/**
+ `AlertProcedure` is a Procedure subclass which will present a `UIAlertController`.
 
-    /// - returns: the presented `UIAlertController`.
-    public var alert: UIAlertController {
-        return presented as! UIAlertController // swiftlint:disable:this force_cast
-    }
+ The alert controller is entirely managed by the procedure. The title, message,
+ text fields and actions are set via properties and methods on `AlertProcedure`.
+
+ A key understanding here is that `AlertProcedure` can finish when it is
+ dismissed, which is the default behaviour, or after it has presented the alert
+ controller.
+
+ In order to finish when the alert is dismissed, it is critical that the alert
+ action handler includes logic to call finish. Therefore, the procedure itself
+ exposes API to add actions, so that finish is called correctly. Therefore,
+ by restricting exposure to the underlying `UIAlertController` this behaviour
+ can be guaranteed.
+ */
+open class AlertProcedure: Procedure {
+
+    internal let controller: UIAlertController
+
+    internal let waitForDismissal: Bool
+
+    // - returns: The presenting view controller
+    public weak var viewController: PresentingViewController?
+
 
     /**
      The style of the alert controller. (read-only)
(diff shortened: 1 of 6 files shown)


Sources/ProcedureKitMobile/Alert.swift
```diff
@@ -76,53 +109,82 @@ public class AlertProcedure: UIProcedure {
...
+     - parameter from: a `UIViewController` instance.
+     - parameter waitForDismissal: a Bool, defaults to true, which indicates whether the
+          procedure should wait until the alert controller is dismissed until finishing.
 
-     - returns: the optional array of UITextField instances
+     - notes: The presenting view controller is weakly held.
      */
-    public var textFields: [UITextField]? {
-        return alert.textFields
+    public init(title: String? = nil, message: String? = nil, style: UIAlertControllerStyle = .alert, from viewController: PresentingViewController, waitForDismissal: Bool = true) {
```
ArkadiGiniApps: @danthorpe I think that we need to remove the option of changing the style from the Init method. Either that or extend the procedure to currently support the action sheet style on both iPhone and iPad. The current version will crash. Due to a missing presentation controller

danthorpe: Yeah, is this because it needs to be put in a pop-over presentation controller?

ArkadiYoskovitz: thats right, otherwise you'll get a crash on the iPad, you also nee do configure the presentation controller with the relevant data or it will still crash

ArkadiYoskovitz: you need something along this lines;


let alertController = UIAlertController(title: "aaaaaa", message: "bbbbbbbb", preferredStyle: .actionSheet)
        alertController.addAction(UIAlertAction(title: "test", style: UIAlertActionStyle.destructive, handler: { (some) in
            
        }))
        present(alertController, animated: true, completion: nil)
        
        if let presentationPopover = alertController.popoverPresentationController {
            
            if let sourceView = ShareHelper.defaultsShareHelper.sourceView, !ShareHelper.defaultsShareHelper.sourceRect.isEmpty {
                
                presentationPopover.sourceView = sourceView
                presentationPopover.sourceRect = ShareHelper.defaultsShareHelper.sourceRect
                
            } else if let barButtonItem = ShareHelper.defaultsShareHelper.barButtonItem {
                
                presentationPopover.barButtonItem = barButtonItem
            }
        }
        present(alertController, animated: true) {
            ShareHelper.defaultsShareHelper.sourceRect = CGRect.zero
            ShareHelper.defaultsShareHelper.sourceView = nil
            ShareHelper.defaultsShareHelper.barButtonItem = nil
        }


Note that you can pass two different combinations to the popoverPresentationController. 
Think that it will be simpler to just make a different procedure for this, also did you ever try to use the actionSheet style in your code? it's not really a common use case

danthorpe: Yeah, I agree - I think we'll fix `AlertProcedure` to just alerts.

blg-andreasbraun:
Seems like you missed the style 😢

```
public init(title: String? = nil, message: String? = nil, from viewController: PresentingViewController, waitForDismissal: Bool = true) {
        self.controller = UIAlertController(title: title, message: message, preferredStyle: .alert)
...
}
```

sets it to alert and the deprecated initialiser has a property for it, but does not set it

```
public convenience init(presentAlertFrom presenting: PresentingViewController, withPreferredStyle preferredAlertStyle: UIAlertController.Style = .alert, waitForDismissal: Bool = true) {
        self.init(title: nil, message: nil, from: presenting, waitForDismissal: waitForDismissal)
} 
```

PR #893, opened 2018-10-05
Added support to specify dispatch queue for event queue in Procedure

evilutioner:
Added possibility to expose DispatchQueue from EventQueue for RxSwift Scheduler

see comments here: https://github.com/ProcedureKit/ProcedureKit/issues/892


danthorpe:
As discussed, we cannot merge this - as it compromises the internal guarantees that are replied upon for `Procedure` to execute correctly.

Issue #897, opened 2018-10-10
`injectResult` retain cycle


kdubb:
With the following code...
```swift
class FetchContentMessageProcedure: GroupProcedure {

  init(_ ctx: MessageContext, summary: MessageSummary) {
    
    let resolveProc = ResolveRoutingProcedure(ctx, aliases: [summary.senderAlias])
    let processProc = ContentMessageProcessProcedure(ctx, summary: summary)
    processProc.injectResult(from: resolveProc)

    let ackProc = AcknowledgeMessageProcedure(ctx, messageId: summary.id, onOrBefore: summary.sent)
    ackProc.addDependency(processProc)

    super.init(operations: [resolveProc, processProc, ackProc])
  }

}
```
neither `ResolveRoutingProcedure` nor `ContentMessageProcessProcedure` are being released properly.

Replacing the `injectResult` line with the following (not suggesting it's a perfectly "correct" solution):
```swift
    processProc.addDependency(resolveProc)
    resolveProc.addCompletionBlock {
      processProc.input = .ready(resolveProc.output.success!)
    }
```
seems to solve the issue.

kdubb:
The offending block appears to be here...
https://github.com/ProcedureKit/ProcedureKit/blob/2bf45aac793c6cb035743d953fe1fbef392a5fc7/Sources/ProcedureKit/ProcedureResult.swift#L195


Adding a `[weak self]` to this block & appropriate strong capture when using `self` the procedures are released correctly.

danthorpe:
Thanks @kdubb - yes, definitely a bug. Fixing it now.

danthorpe:
@kdubb - can you try your code against the branch in #898?

I did write a unit test in an attempt to catch this error - but found it a little bit inconclusive. So will continue testing in my own projects.

I'm pretty sure the error is caused by the `self` which is passed into the block, this is a mistake, as it should use the `procedure` argument which it receives and is already weak.

kdubb:
Yep. Appears to work for me! Thanks!

danthorpe:
Thanks @kdubb - will aim to merge this branch into `development` asap.

danthorpe:
Closing this now, as the fix is merged.

PR #898, opened 2018-10-10
Fixes retain cycle due to strong self capture in inject result

danthorpe:
See #897.


Sources/ProcedureKit/ProcedureResult.swift
```diff
@@ -198,7 +195,7 @@ public extension ProcedureProtocol {
             case let .cancel(e):
                 procedure.cancel(with: e)
             case let .success(output):
-                block(self, output)
+                block(procedure, output)
```
danthorpe: This is the fix.

Issue #899, opened 2018-10-10
Cannot redirect all logging


kdubb:
The initialization of the `system` log here:
https://github.com/ProcedureKit/ProcedureKit/blob/2bf45aac793c6cb035743d953fe1fbef392a5fc7/Sources/ProcedureKit/Procedure.swift#L393
... hardwires it to your system `LogWriter`; which for my platforms is an `OSLogWriter`.

I assume this should be pointing to a setting, much the way the `log` above it is initialized.  Without this it seems we cannot redirect all logging.

kdubb:
Scratch that... we _*could*_ redirect all logging if we set this variable in every single procedure we define but that is cumbersome and error prone.

kdubb:
Also, all the documentation about custom logging appears woefully out of date with the latest.  

Currently we are setting the `Log.writer` variable directly; maybe part of the issue is we're not redirecting it correctly.

danthorpe:
Thanks @kdubb for this. Looking into it. 👍

danthorpe:
Hi @kdubb,

The `system` logger, which is a property inside every `Procedure`, is intended to be used internally by _ProcedureKit_ for its internal logging. These are the messages which previously would have been sent if the severity was set to `.verbose`.

Of course, these are useful debugging messages too and where supported they are directed into `os_log` under a `run.kit.procedure` subsystem. This potentially is an issue for framework consumers, though, looking for log messages under their bundle ID perhaps - so this could be changed.

If it was made a property, which defaulted to a `Log` static property, this would allow you to provide any Log[Channel|Writer|Formatter] that you wished. _However_ I'm curious to learn what you would like to set this to? Perhaps my expectation of its usage is wrong?

kdubb:
Not being able to redirect your `system` means we lose the ability to trace problems related to procedures after the fact using our online logging platform.  The messages are ___extremely___ useful for debugging and we want them saved in certain cases.

Without being able to redirect them we are left to simply just recreate them; which is senseless give the great amount of information you're already logging.

danthorpe:
Hey @kdubb - thanks for getting back to me.

See #906 - I think following these changes you should be able to make your custom logger conform to either `LogChannel` or just `LogWriter` and set it via the static properties on `Log`. Can you let me know what you think?

I will update the docs in another PR.

PR #906, opened 2018-10-24
Updates to Logging

danthorpe:
- [x] Exposes the log channel via settings.
        It defaults to the shared instance, but framework consumers will be able to implement their own `LogChannel` and set it using:

    ```swift
    Log.channel = mySpecialChannel
    ```
    
- [x] Removes `system` log (which used a private log writer)
        This will allow framework consumers to re-direct all logging through a central log channel/writer/formatter as they wish. This means, that framework consumers can use the default `Log.Channel` instance, but could set a special log writer, for example:

    ```swift
    Log.writer = mySpecialLogWriter
    ```
    which is equivalent of:

    ```swift
    Log.channel.writer = mySpecialLogWriter
    ```

- [x] For platforms which support `OSLog`
        Added an API to use the default LogWriter (`Log.Writers.OSLogWriter`) but with a given instance of `OSLog` - this will allow framework consumers to setup the log using their own subsystem and category. Unfortunately, at this time, we can only use a single OSLog instance. It is recommended that app developers create a `OSLog` instance dedicated for _ProcedureKit_ and use their bundle ID as the subsystem, and "ProcedureKit" as the category.

Created following discussions in #899 

So, for example, it's now possible to do this:

```swift
Log.writer = Log.Writers.Redirecting(writers: [
    Log.Writers.OSLogWriter(log: myOSLog), 
    remoteCloudLogger])
```


pronebird:
@danthorpe are devs able or supposed to use Log facilitates outside of Operations?

danthorpe:
Well, when I initially refactored the logger, I wanted _ProcedureKit_ contributors to be able to get logs on how the framework works, which would be separate from the logger that framework consumers might want to use for their own purposes.

But to be honest, following discussions in #899 - it's kinda easier to put everything into a single logger really.

PR #908, opened 2018-10-28
Add variable name in signatures that include "with: Error?"

pronebird:
The Xcode auto-completes the `ProcedureObserver` methods with `with` preposition without the appropriate variable name.

```swift
func did(finish procedure: Procedure, with: Error?) {
  if let error = with { // ??? clearly a mistake

  }
}
```

### DIFF
--- a/Sources/ProcedureKit/ProcedureObserver.swift
+++ b/Sources/ProcedureKit/ProcedureObserver.swift
@@ -58,7 +58,7 @@ public protocol ProcedureObserver {
 
      - parameter procedure: the observed `Procedure`.
      */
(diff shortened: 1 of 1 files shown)


danthorpe:
Hi @pronebird - thanks for this!

You're quite right, the auto-complete is not great here. Personally, I think Xcode should add a placeholder for you to enter your own variable name here (maybe you just want to called it `e` instead of `error` for example).

But yeah, your change is pragmatic 👍, will merge shortly.

pronebird:
@danthorpe build kite failed for no obvious reason. Not sure why. Thanks for merging

PR #909, opened 2018-10-28
Add missing  parameter name in various functions

juliensagot:
There is a missing `error` parameter name in various functions.
It's needed as the code looks weird without it:

Before:
```swift
override func procedureDidFinish(with: Error?) {
      guard with == nil else { return }
       ...
}
```

After:
```swift
override func procedureDidFinish(with error: Error?) {
      guard error == nil else { return }
       ...
}
```


danthorpe:
Thanks so much for this @juliensagot 😁

PR #912, opened 2018-12-10
Open network recovery

ericyanush:
This merge request opens the NetworkRecovery<T> class, allowing customized recovery logic when network procedures fail. 

This makes handling cases when network procedures fail, and are recoverable, but require additional configuration, for example requests that have failed (401) using Oauth, which are possibly recoverable using a refresh token:

```swift
class AuthenticatedNetworkRecovery: NetworkRecovery<SecuredNetworkProcedure> {
    let authManager: AuthManager

    init(authManager: AuthManager) {
        self.authManager = authManager
        super.init(resilience: DefaultNetworkResilience(), connectivity: .any)
    }

    override func recover(withInfo info: RetryFailureInfo<SecuredNetworkProcedure>, payload: RepeatProcedurePayload<SecuredNetworkProcedure>) -> RepeatProcedurePayload<SecuredNetworkProcedure>? {
        let recovery = super.recover(withInfo: info, payload: payload)

        let response = info.operation.makeNetworkResponse()
        if let recoveryOp = recovery?.operation, response.httpStatusCode == .unauthorized {
            let login = TokenRefresh(currentToken: info.operation.accessTokenUsed, 
                                     authManager: authManager)
            recoveryOp.addDependency(login)
            recoveryOp.addCondition(NoFailedDependenciesCondition())
            info.addOperations(login)

            return RepeatProcedurePayload(operation: recoveryOp, delay: nil, configure: nil)
        }

        return recovery
    }

    override func shouldRetry(givenNetworkResponse networkResponse: ProcedureKitNetworkResponse) -> Bool {
        if networkResponse.httpStatusCode == .unauthorized {
            return true
        }

        return super.shouldRetry(givenNetworkResponse: networkResponse)
    }
}
```

Handling cases such as these currently, requires significantly more code/effort, especially if you want to utilize all of the code written to handle reachability, transient errors, et all. From what I can tell, there was currently no way to implement logic where a failed network operation should be retried, but only after an additional dependancy completes successfully first.

As a side effect of opening the NetworkRecovery<T> class, the ProcedureKitNetworkResponse and ProcedureKitNetworkError structs also had to be exposed.

In general, this change won't effect current users of the NetworkProcedure<T>, however, it does allow any cases requiring more complex or custom recovery logic to be handled quite easily.

### DIFF
--- a/Sources/ProcedureKitNetwork/Network.swift
+++ b/Sources/ProcedureKitNetwork/Network.swift
@@ -97,22 +97,22 @@ class NetworkReachabilityWaitProcedure: Procedure {
     }
 }
 
-class NetworkRecovery<T: Procedure> where T: NetworkOperation {
+open class NetworkRecovery<T: Procedure> where T: NetworkOperation {
 
     let resilience: NetworkResilience
     let connectivity: Reachability.Connectivity
-    var reachability: SystemReachability = Reachability.Manager.shared
+    fileprivate var reachability: SystemReachability = Reachability.Manager.shared
 
     var max: Int { return resilience.maximumNumberOfAttempts }
 
     var wait: WaitStrategy { return resilience.backoffStrategy }
 
-    init(resilience: NetworkResilience, connectivity: Reachability.Connectivity) {
+    public init(resilience: NetworkResilience, connectivity: Reachability.Connectivity) {
         self.resilience = resilience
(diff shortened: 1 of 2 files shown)


danthorpe:
Hi @ericyanush - thanks so much for this - sorry it's taken me this long to merge it. 😔

PR #931, opened 2019-04-01
Fix: Add missing presentation style to AlertProcedure initializer

enhorn:
The initializer did not take the presentation style as a parameter as hinted in the deprecation message, but instead always used alert. I've added it and made sure naming matched the deprecation message for the old initializer.


danthorpe:
Hi @enhorn - we did have a discussion about this last year... See: https://github.com/ProcedureKit/ProcedureKit/pull/882#pullrequestreview-149505728

The issue, is that if you use `.action` on an iPad, iOS will attempt to present the alert in a pop-over, and it ultimately causes a crash. I appreciate that it might work for you in some cases.

I need to have a think about how we can resolve this issue. Thanks so much for contributing though - I'm very much aware that `UIProcedure` needs some TLC...

PR #938, opened 2019-06-01
Add call to didSetInputReady in bind(to: ..)

### LINKED ISSUE #936: bind(to: ..) does not call didSetInputReady
Hi,

`func bind<T: InputProcedure>(to target: T) where T.Input == Self.Input` relies on `didSetInput` observer. Yet when it binds the input to the target, it does not notify the target that it set the input. This leads to really hard to track bugs. 

This can easily be reproduced by using two operations that leverage `bind(to: )` to proxy the `Input` to the third child operation:

First Operation bind-to -> Second Operation bind-to -> Third Operation.

Since the `bind(to:)` call from the first operation only sets `input` and does not send the `didSetInputReady` event, the second `bind(to:)` will never be triggered, hence the third operation will never receive the input.

The solution would be to call `didSetInputReady` on the target, i.e:

```swift
public extension InputProcedure {

    func bind<T: InputProcedure>(to target: T) where T.Input == Self.Input {
        addDidSetInputReadyBlockObserver { (procedure) in
            target.input = procedure.input

            // This part is missing?
            if case .ready = procedure.input {
                target.didSetInputReady()
            }
        }
    }
}
```

I don't fully understand such omission. Also, I see that `BatchProcedure` calls `didSetInputReady()` manually. 

The alternative to using the "ready" event is of course to rely on `willExecute` hook, which is exactly what I do in that peculiar case of double binding:

```swift
addWillExecuteBlockObserver { (procedure, _) in
   createRequestProcedure.input = procedure.input
}
```

Thoughts?

pronebird:
As described in the linked issue, `bind(to: InputProcedure)` does not call `didSetInputReady` which causes issues if the procedure, that receives the input, uses `bind(to: ..)` internally to proxy the same input down to children.

It's possible that we would have to remove the manual calls to `didSetInputReady` across the codebase, such as in (IIRC) `BatchProcedure`, perhaps I should address this in this PR as well...

Fixes #936

### DIFF
--- a/Sources/ProcedureKit/ProcedureResult.swift
+++ b/Sources/ProcedureKit/ProcedureResult.swift
@@ -247,7 +247,10 @@ public extension InputProcedure {
 
     func bind<T: InputProcedure>(to target: T) where T.Input == Self.Input {
         addDidSetInputReadyBlockObserver { (procedure) in
-            target.input = procedure.input
+            if case .ready = procedure.input {
+                target.input = procedure.input
+                target.didSetInputReady()
+            }
         }
     }
 }


pronebird:
Closing due to PR being stale and no longer relevant.

PR #960, opened 2021-05-07
Fix Xcode 12.5 compilation

hynekhanke:
As per Xcode 12.5 Release Notes, the TestingProcedureKit target which uses
XCTest needs to specify ENABLE_TESTING_SEARCH_PATHS in order to be able to find
XCTest symbols.

See https://developer.apple.com/documentation/xcode-release-notes/xcode-12_5-release-notes


zocario:
Thanks for the PR, would love to see it merged in a new release, especially for use with Carthage 🙏

stephengurnett (approved):
Tested and works in our project.

